PR #10, opened 2015-08-08
Added comments in configuration files

ilario:
I added some comments in Libre-Mesh configuration files.

### DIFF
--- a/packages/lime-system/files/etc/config/lime
+++ b/packages/lime-system/files/etc/config/lime
@@ -5,7 +5,8 @@
 # calculated from the hash of the ap_ssid value, so that all the nodes that
 # form a mesh cloud (share the same ap_ssid) will produce the same value
 # and %H template that will be replaced by hostname
-
+# For setting a specific WAN port, don't set it globally in section "config lime network", set it in interface specific configuration "config net ..." and install the lime-proto-wan package
+# For setting the WAN port on the default WAN port (according OpenWRT), no configuration here is needed, just install the lime-hwd-openwrt-wan package.
 
 ### System options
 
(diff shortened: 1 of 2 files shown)


altergui:
Thanks ilario!
the changes to the comments look good,
but there are also changes in the (commented) config which don't :P

i don't see any advantage in including %M in main_ipvX_addresses, since that is already handled by leaving trailing zeroes in a more efficient way (it can use netmasks other than /24 , /16)
in fact, they should include %N instead, so that two different networks get two different ip prefixes by default

at some point in time this used to be
option main_ipv4_address '10.%N1.%N2.0/24'             
option main_ipv6_address 'fddf:ca%N1:%N2%N3:0000::/64' 

anyway, i would leave it untouched for now :)

olsr protocols shouldn't have a vlan id >= 16, since that can clash with batadv protocol (randomly chosen between 16 and 272)

vlans id  < 16 were reserved precisely for hardcoded protocols like this :)

try olsr:14 and olsr6:15  , but in a separate commit ("Add olsr and olsr6 to default protocols list")

ilario:
> i don't see any advantage in including %M in main_ipvX_addresses,

I modified it for making its behaviour explicit, without comments it was obscure.

> since that is already handled by leaving trailing zeroes in a more efficient way (it can use netmasks other than /24 , /16)

Ok, if it's more efficient I'll remove this change :)

> in fact, they should include %N instead, so that two different networks get two different ip prefixes by default

Well, dunno, up to you devs.

> olsr protocols shouldn't have a vlan id >= 16, since that can clash with batadv protocol (randomly chosen between 16 and 272)
> vlans id < 16 were reserved precisely for hardcoded protocols like this :)
> try olsr:14 and olsr6:15 , but in a separate commit ("Add olsr and olsr6 to default protocols list")

Thanks for the suggestion! 
I did put 23 and 24 just because they're the default ones I'm going to put this (with 14,15) in a separate pull request :)

G10h4ck:
> I did put 23 and 24 just because they're the default ones
> I'm going to put this (with 14,15) in a separate pull request :)

ilario:
>> I did put 23 and 24 just because they're the default ones
>> I'm going to put this (with 14,15) in a separate pull request :)

> Nope this should be fixed before merge in both pull request, merge should not introduce know bug 
> then we  have to fix with another pull request fix both pull requests if you want to see them merged ;)

I already removed those lines in afa1645 and 5708dc4 ... What do you mean?

G10h4ck:
It seems to me a good idea to squash is a single commit the whole pull request can you do this ?

ilario:
>  squash [...] commit [...] can you do this ?

No!! But I can try...

ilario:
> squash [...] commit [...] can you do this ?

I think I managed, is this ok?

PR #13, opened 2015-08-18
Added comments and description for ground routing package

ilario:
I added the description for lime-hwd-ground-routing package and added comments in /etc/config/lime file in ground routing section.


G10h4ck:
Have you tested it if it compiles and run fine ?

ilario:
The pull request is just about some comments and a description of a package, so it should compile and run just as well as before this commit. Anyway it compiled fine but up to now I have no hardware for testing it.

ilario:
In the Makefile in the URL: field we have to put the homepage http://libre-mesh.org or we could put the module specific page http://libre-mesh.org/projects/libre-mesh/wiki/Ground_routing ?

G10h4ck:
It seems a good idea to point to the specific page for this package

ilario:
Amended adding the complete URL

PR #14, opened 2015-08-31
Removed OLSR from default protocols list

ilario:
Following suggestions in comments on 40d88b6


G10h4ck:
It is not the real default list as soon as it is not in lime-full so i would reject this pull request

PR #15, opened 2015-08-31
New branch: change dependencies, remove BMX6 add OLSR

ilario:
Following suggestions in comments on 40d88b6

### DIFF
--- a/packages/lime-full/Makefile
+++ b/packages/lime-full/Makefile
@@ -22,16 +22,16 @@ define Package/$(PKG_NAME)
   CATEGORY:=LiMe
   MAINTAINER:=Pau Escrich <p4u@dabax.net>
   URL:=http://libre-mesh.org
-  DEPENDS:=+lime-system +lime-proto-bmx6 +lime-proto-batadv +lime-proto-anygw +lime-proto-wan \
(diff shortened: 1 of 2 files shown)


packages/lime-full/Makefile
```diff
@@ -22,16 +22,16 @@ define Package/$(PKG_NAME)
   CATEGORY:=LiMe
   MAINTAINER:=Pau Escrich <p4u@dabax.net>
   URL:=http://libre-mesh.org
-  DEPENDS:=+lime-system +lime-proto-bmx6 +lime-proto-batadv +lime-proto-anygw +lime-proto-wan \
+  DEPENDS:=+lime-system +lime-proto-olsr +lime-proto-olsr6 +lime-proto-batadv +lime-proto-anygw +lime-proto-wan \
```
ilario: Is ok to make *lime-full* depending from just one layer 3 protocol or it should depend on the *full* list of supported protocols, now that this wouldn't hurt?

G10h4ck:
For me it's not a good idea to keep a branch for each routing protocol, at moment the consensus is to install BMX6 by default who need having other protocols by default should do it in it's how fork/sandbox

Issue #18, opened 2015-09-27
Separate settings for netmask and IP range auto assignment


ilario:
I like the flexibility of the [main_ipvX_addresses auto completing feature](https://github.com/libre-mesh/lime-packages/blob/develop/packages/lime-system/files/etc/config/lime#L22) but I would **separate the netmask from the IP auto assignment range**.
Is there a way to **specify a different netmask**?
For example, [I used](https://github.com/VeronaWirelessCommunity/lime-packages/blob/develop/packages/lime-system/files/etc/config/lime-defaults#L11):

     option main_ipv4_address '10.37.188.0/22'

and it works very well for auto assigning the IP, but I did want to have a /16 netmask on this IP...
Thanks

G10h4ck:
What happen if you substitute 22 with 16 in your configuration ? What do you expected instead?

ilario:
> What happen if you substitute /22 with /16 in your configuration ?
> What do you expected instead?

If I use
    option main_ipv4_address '10.37.188.0/22'
I get
* IPv4 auto completed in range 10.37.188.1-10.37.191.254
* netmask /22

If I use
    option main_ipv4_address '10.37.0.0/16'
I get
* IPv4 auto completed in range 10.37.0.1-10.37.255.254
* netmask /16

But I do want
* IPv4 auto completed in range 10.37.188.1-10.37.191.254
* netmask /16

> Moreover I just noticed you are trying to use a /72 for ipv6 this cannot work 
> the minimum required is a /64 subnet, if you need more then one just take it from librenet6.

Same discourse as for IPv4, I want:
* IPv6 auto completed in range 2a00:1508:0001:f015:ef00:0000:0000:0000-2a00:1508:0001:f015:efff:ffff:ffff:ffff and I get this setting 2a00:1508:1:f015:ef00::/72
* netmask /64

ilario:
So, my problem is that in the /16 subnet I have also some static IPs, so the range of IP autocompletion should be smaller than the whole subnet.

I propose to add an **optional** field in *main_ipv4_address* separated by a slash, the first field is the IP or network, the second is for autocompletion and the third is for the netmask. If just two fields are present the second will be used for both concepts, as happens now. So in my case it would result:

    option main_ipv4_address '10.37.188.0/22/16'

obtaining
* IPv4 auto completed in range 10.37.188.1-10.37.191.254 (indicated by /22)
* netmask for subnet /16

If it seems ok to you I think I can implement this.
[...]

PR #22, opened 2015-10-17
Separate IP autocompletion range from subnet

ilario:
Here I implemented a possible approach to #18 (Separate settings for netmask and IP range auto assignment).

Two different ranges for IP auto assignment and for subnet can be set in the same option main_ipv4_address using this format: ip.ad.dr.es/range for IP/subnet. 
For example

    option main_ipv4_address '192.0.128.0/17/16'

will set the IP address in the range 192.0.128.1-192.0.255.254 while the subnet mask will be /16 so leaving the first half of the subnet out of auto assignment range and available for static IPs.

Retrocompatibility is respected, so when the option is specified with just one "/" the patch is not having any effect.

This is just one of the possible way for solving #18, separating the two concepts in two options would also be a good idea.

### DIFF
--- a/packages/lime-system/files/etc/config/lime
+++ b/packages/lime-system/files/etc/config/lime
@@ -1,7 +1,7 @@
 # The options marked with "# Parametrizable with %Mn, %Nn, %H"
 # can include %Mn templates that will be substituted
 # with the n'th byte of the primary_interface MAC
-# and %Nn templates that will be replaced by the n'th network-identifier byte,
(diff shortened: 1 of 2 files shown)


G10h4ck:
I believe for consistence it is better to have

option main_ipv4_address '192.0.128.0/16/17'

instead of

option main_ipv4_address '192.0.128.0/17/16'

ilario:
Ok, I corrected in the way you suggested.

PR #23, opened 2015-10-19
Moving documentation from lime to lime.example file

ilario:
Often happens to me that I write a configuration line in /etc/config/lime file but I forget to uncomment also the relative section name.
A newbie could even not know that uncommenting the section name is necessary.
I think that uncommenting the section names could be a convenient solution.
And it doesn't cause new errors.


ilario:
I expanded the purpose of this pull request.
Having the documentation in the /etc/config/lime file isn't wise as some lime modules could wipe all the comments in it (for example the lime-hwd-openwrt-wan).
So I moved it to lime.example file.

G10h4ck:
When needed rebase instead of merge from develop so the history netowork graph doesn't get complicated

ilario:
Mh, so now what should I do?

PR #24, opened 2015-10-24
Added dependency from lime packages in auto-gw-mode modules

ilario:
batman-adv-auto-gw-mode and bmx6-auto-gw-mode modules are going to check if the internet connection is present in a WAN port, so a WAN port has to be set for these modules to work, likely this will happen using the lime-proto-wan.


ilario:
I added as dependencies also the relative proto modules. 

@devs Maybe the fact that all these dependencies were missing was for making these package LiMe indipendent?

G10h4ck:
I believe it was not introduced as a dependency because of that.

Issue #7, opened 2015-01-29
luci.ip.Hex() does not exist anymore


p4u:
luci.ip.Hex() function has been removed from the luci libraries.

https://github.com/libre-mesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/network.lua#L37

IRC log about this issue:

14:02 < p4u> lua from Jan 15: ip.lua -> function Hex( hex, prefix, family, swap )
14:03 < p4u> now I see all the libraries have been moved and some of then even rewritten
14:03 < p4u> the ip module is missing the Hex() function
14:03 < p4u> as descreibes the documentation
14:03 < p4u> jow_laptop: what you mean for input data?
14:03 -!- iamfrankenstein [~Thunderbi@h90192.upc-h.chello.nl] has quit [Ping timeout: 245 seconds]
14:03 < jow_laptop> p4u: yeah I dropped Hex() because I thought only luci uses it internally for parsing /proc/net/ipv6_route
14:04 < jow_laptop> p4u: therfore I wanted to know what kind of data you feed to hex, to see if you can use another variant or if I should reimplement the Hex() constructor
14:04 < p4u> https://github.com/libre-mesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/network.lua#L37
14:05 < p4u> we were using it ;(
14:05 -!- BankZ- [~bankz@c-73-161-152-148.hsd1.mi.comcast.net] has quit [Ping timeout: 252 seconds]
14:08 < jow_laptop> hm, ok I can readd Hex() but a few other parts of the code will break too because you access private internals of the cidr object
14:08 < p4u> ok
14:09 < p4u> we can fix it
14:09 < p4u> or maybe find another way for doing the same

G10h4ck:
On Thursday, January 29, 2015 05:13:40 AM Pau wrote:
> Assigned #7 to @G10h4ck.
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/libre-mesh/lime-packages/issues/7#event-226805821

Is this bug present in last stable or just in trunk?
If it is just in trunk i would wait that trunk stabilize a little before 
changing stuff in libre-mesh to fix stuff like this

P.S. please keep dev@lists.libre-mesh.org in CC

p4u:
It is not trunk neither stable, it is just "LUCI", the repository
"http://git.openwrt.org/project/luci.git". We are using it in all LiMe
Branches.

[...]

PR #25, opened 2015-10-25
Created package with legacy luci.ip

ilario:
I moved the luci.ip library from LiMe libraries to a standalone package.

I tested this with both OpenWRT Barrier Breaker and OpenWRT Chaos Calmer and it does its work solving #7.

Indeed with Chaos Calmer also the rest of the LiMe system seems working so we should really consider to merge this also in develop branch.

### DIFF
--- a/packages/libluciip-lua/Makefile
+++ b/packages/libluciip-lua/Makefile
@@ -1,0 +1,44 @@
+#
+# Copyright (C) 2006-2015 OpenWrt.org
+#
+# This is free software, licensed under the GNU General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
+GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
+GIT_COMMIT_TSTAMP:=$(shell git log -n 1 --pretty=%at . )
+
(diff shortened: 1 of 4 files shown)


G10h4ck:
I would remove the term legacy around as there is no reason to have this called legacy, they removed it from Luci only because they didn't use it any-more.

ilario:
Removed the "legacy" terms.

G10h4ck:
Seems fine, when you do more testing please consider to do a merge request of this branch to develop

p4u:
Sorry for appearing probably too late, but I would like to comment anyway.

First of all, let me see that you made a single commit named "Created package with legacy luci.ip " and IMO this information is not correct (you made more changes). Secondly, let me make two suggestions:

- You changed the name of the lua package from lime.ipl to libluciip 
I don't see the need of naming it **luci** because the library does not belong to luci anymore, we have forked it to keep using it, so the name IMO should be something like lime.ip or lime.ipcalc

- You added all the old comments
I deleted all the comments because it increases the size of the file and we don't have that much space to waste on not necessary data. If you want to keep the comments you must write some shell commands to remove them before installing (see other lime Makefiles to have some example)

Finally, thank you for contributing!

PR #35, opened 2015-12-26
Added lime-hwd-ground-routing to lime-full

ilario:
Including *lime-hwd-ground-routing* module in lime-full will have no negative impact and will allow to configure [ground routing](http://libre-mesh.org/projects/libre-mesh/wiki/Ground_routing) set-ups.

### DIFF
--- a/packages/lime-full/Makefile
+++ b/packages/lime-full/Makefile
@@ -26,7 +26,7 @@ define Package/$(PKG_NAME)
 		+dnsmasq-lease-share \
 		+lime-webui +luci-app-batman-adv +luci-app-bmx6 \
 		+lime-map-agent \
-		+lime-debug \
+		+lime-debug +lime-hwd-ground-routing \
 		+lime-hwd-openwrt-wan +bmx6-auto-gw-mode
 endef
 


G10h4ck:
Have you tested exhaustively if this has side effects on generated configuration with default settings ?
I believe it was not added to lime-full because ground-routing is not contemplated in the default LiMe networking schema (like olsr or bgp), but if you are sure there is no unexpected side effects this can be merged as it doesn't add space consuming extra dependencies like lime-proto-* usually do.
Cheers!

G10h4ck:
Any news ?

ilario:
No, I haven't tested it yet.

ilario:
I tested comparing /overlay and /etc on Libre-Mesh on top of LEDE (compiled using branch develop-lede of lime-build) including or not the lime-hwd-ground-routing package. Without configuring the ground-routing I can see no relevant difference in the observed directories (except some change in latitude and longitude in libremap conf wtf). You can find the directories in study [here](https://uz.sns.it/~ilario/libre-mesh-ground-routing-in-lime-full/).

G10h4ck:
As discussed at LimeCat 2016q3 i have added it as dependency of lime-basic
see 59997b597cef49cd92f51a28a06ffa6a0f61a759

PR #36, opened 2015-12-30
lime-debug: add tcpdump-mini

ilario:
In my opinion tcpdump mini should be included in lime-debug.


G10h4ck:
I was thinking to include tcpdump some time ago, what is the difference between tcpdump-mini and tcpdump ? If you are using batman-adv you can use batctl tcpdump eth0 but it doesn't support filters like tcpdump

ilario:
tcpdump-mini_4.5.1-4_ar71xx.ipk   140815 byte
tcpdump_4.5.1-4_ar71xx.ipk   269267 byte

https://dev.openwrt.org/browser/packages/net/tcpdump/Makefile?rev=19711#L58

Can't find more information.

PR #38, opened 2016-01-10
Update README.md, Grammatical structure corrected

Humzini:
Added some grammar and syntax, "by" and a comma to make line 11 read better and become grammatically correct.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ The firmware (the main piece) will allow simple deployment of auto-configurable,
 
 ## Building a firmware image
 
-The Libre-Mesh firmware can be compiled either manually adding the feed to a [OpenWrt buildroot][1] environment or using the easy to use [lime-build][2] tool.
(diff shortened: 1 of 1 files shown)


ilario:
The "by" sounds correct to me (as is sounded without it anyway...).
But the comma, in that place, is detrimental because "adding the feed to a OpenWrt buildroot environment" is meant to be an explication of what "manually" means. Anyway I admit that is absolutely not clear, can you amend your commit correcting the phrase in another way?
Thanks

G10h4ck:
I would use the incidental form with two commas like this

---------------------------------------------
The Libre-Mesh firmware can be compiled either manually, adding the feed to a [OpenWrt buildroot][1] environment, or by using the easy to use [lime-build][2] tool.
---------------------------------------------

In Italian putting a phrase between two commas like that is mostly like putting it between parentheses but with less ink :p is the same for English ?

ilario:
Inserting a `;` between `manually` and `adding the feed` separates the two concepts while the second is an explanation of the first.

Issue #39, opened 2016-03-09
lime & luci packet dependencies


trizonelabs:
we do have a hard time to find & control the dependencies which seems to be messed up.
trying to build for 4MB devices with CC we always run in overhead of bildin packets.

exmpl : * bxm , ipv6..a.s.o*

 please can someone to a grep / sort & dependency tree to rebuild the menuconfig points
& check if they are really necessary.

The openwrt ipv6 global build setting / 'Enable IPv6 support in packages' could come quite handy if implemented in the Makefile

we do have to do this manually with every new feeds - update

seems to us there is a sever lack of consolidation in project base.
sorry but we dont have the knowledge & manpower for pull requests because we have to concentrate our effords at the application layer.

The additional work makes it very difficult to keep lime into our standard.setup

regards 3zl /  i4free-gr

ilario:
> dependencies which seems to be messed up

We had a huge merge in February 28th 761e9c8, changing also some Makefile. @lime-devs could this be related?

> overhead of bildin packets. exmpl : * bxm

I suppose this, regarding BMX6, is caused by #28 and analogous problems in other Makefiles.

ilario:
> The openwrt ipv6 global build setting / 'Enable IPv6 support in packages' could come quite handy if implemented in the Makefile

For me `CONFIG_IPV6` which is "Global build settings > Enable IPv6 support in packages (passes --enable-ipv6 to configure scripts)." is selected by default...

G10h4ck:
I can't reproduce this problem, 3zl can you do a double check ?

Issue #28, opened 2015-11-01
bmx6 NOT selected builds/installes  bmx6 libs


trizonelabs:
make menuconfig bmx6 deselcted builds bmx6 related blobs into the image.
As its quite big would be better to inhibit this via deselected bmx6.

3zl

ilario:
Could you rephrase this? 
I'm not used to bmx6 so I didn't get the problem.
Thanks,
Ilario

trizonelabs:
Ilario tnx for response .
configs from last build 
 
build ESXi 20151102.1707

feeds:
src-git lime https://github.com/libre-mesh/lime-packages.git;sandbox/release/15.09

diffconfig:
Routing and Redirection
CONFIG_PACKAGE_bmx6=y

CONFIG_PACKAGE_luci-app-bmx6=y
CONFIG_PACKAGE_luci-lib-libremap-bmx6=y


.config
LiMe
# CONFIG_PACKAGE_bmx6-auto-gw-mode is not set
LUCI 
	3.Applications 
CONFIG_PACKAGE_luci-app-bmx6=y
	8.Libraries
CONFIG_PACKAGE_luci-lib-libremap-bmx6=y

menuconfig search :
Symbol: PACKAGE_luci-app-bmx6 [=y]                                                                      

  │  
  │ Type  : tristate                                                                               

           │  
  │ Prompt: luci-app-bmx6........ bmx6 configuration, status and visualization module     

                    │  
  │   Location:                                                                  

                             │  
  │     -> LuCI                                                         

                                      │  
  │ (1)   -> 3. Applications                                   

                                               │  
  │   Defined at tmp/.config-package.in:29668         

                                                        │  
  │   Selects: PACKAGE_libssp [=n] && 

PACKAGE_bmx6 [=y] && PACKAGE_luci-lib-json [=y] && PACKAGE_luci-mod-adm │  
  │   Selected by: 

PACKAGE_lime-full [=n] && IPV6 [=y] && !PACKAGE_dnsmasq [=y] || PACKAGE_lime-webui [=y] 

so its seems bmx6 is selected though lime_webui ?
I could not find another place where it is activated.

(sorry for the format , but editing text quite combersome)

cheers 
3zl (Reinhard)

ilario:
Hey you closed the issue, I suppose by error.

[...]

PR #40, opened 2016-03-09
Change to conditional dependencies for lime-proto-x

ilario:
Using [conditional dependencies](https://wiki.openwrt.org/doc/devel/dependencies#dependency_types) for avoiding the selection of some routing packages when not explicitly requested.
This should solve #28 and partially #39.


G10h4ck:
Thanks you

PR #41, opened 2016-03-09
Added conditional dependencies to non-lime packages and update dependency from BMX6 to BMX7

ilario:
Sorry for putting the two commits in the same pull requests but both commits were changing the same line.
The first is for making the optional proto related modules to select also the relative proto module when lime-system is selected. This solves also #24.
The second is for coherence with 52bfb5a.

### DIFF
--- a/packages/batman-adv-auto-gw-mode/Makefile
+++ b/packages/batman-adv-auto-gw-mode/Makefile
@@ -19,7 +19,8 @@ define Package/$(PKG_NAME)
   TITLE:=Set batman-adv gw_mode by internet availability
   MAINTAINER:=Gui Iribarren <gui@altermundi.net>
   URL:=http://libre-mesh.org
-  DEPENDS:= +kmod-batman-adv +watchping +dnsmasq-dhcpv6 +lime-eb-ip-tables +ip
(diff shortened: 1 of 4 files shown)


ilario:
When I do a `make menuconfig` I get errors (warnings?):

    error: recursive dependency detected!
      symbol PACKAGE_lime-system is selected by PACKAGE_lime-proto-batadv
      symbol PACKAGE_lime-proto-batadv is selected by PACKAGE_lime-system
    error: recursive dependency detected!
      symbol PACKAGE_lime-proto-bmx6 is selected by PACKAGE_bmx6-auto-gw-mode
      symbol PACKAGE_bmx6-auto-gw-mode depends on PACKAGE_bmx7 
      symbol PACKAGE_bmx7 is selected by PACKAGE_lime-proto-bmx6

These relates to the fact that, in some packages, the selection of `lime-system` is a condition for the selection of `lime-proto-x` dependencies but `lime-system` is also a dependency of these `lime-proto-x`. In my opinion this is a legitimate set of dependencies... Isn't?

G10h4ck:
It seems a legitimate set of dependencies to me too, would you take care of talking with OpenWrt developers about this strange warning so i would merge with less concerns

ilario:
The behavior in menuconfig seems correct.

ilario:
I reported this in OpenWrt Trac [here](https://dev.openwrt.org/ticket/22226).

PR #42, opened 2016-03-09
Added dependency from lime-system in lime-webui

ilario:
Seems logic for lime-webui to depend from lime-system, in my opinion.

### DIFF
--- a/packages/lime-webui/Makefile
+++ b/packages/lime-webui/Makefile
@@ -25,7 +25,8 @@ define Package/$(PKG_NAME)
 	+luci-app-batman-adv \
 	+luci-mod-status \
 	+uhttpd +libiwinfo-lua \
-	+luci-theme-bootstrap +luci-i18n-english
+	+luci-theme-bootstrap +luci-i18n-english \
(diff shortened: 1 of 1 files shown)


G10h4ck:
Have you tested, if this cause some warning/build break ?

p4u:
I won't do it. Lime-webui requires LuCi and IMO there is no need for lime-system to require a whole web interface.

I think lime-webui is fine as a dependency for lime-full, which means lime-system plus some other stuff.

ilario:
> Lime-webui requires LuCi and IMO there is no need for lime-system to require a whole web interface.

Absolutely agree on this, but I think you got the meaning of the pull request the other way around.
As the title says, the idea is to add a *dependency from lime-system in lime-webui* and not the dependency from lime-webui for lime-system.

@G10h4ck no, I still have to check.

ilario:
@G10h4ck I just tried to compile using lime-build and the additional dependency, no new dependency problems arose.

G10h4ck:
Thanks you!

PR #48, opened 2016-05-27
Lime freifunk

dangowrt:


p4u:
IMO all changes make sense and they are correct. But let's wait for some other opinion before merging it.

Thanks.

FreifunkUFO:
nice!
do we have to set mcast rate (lower than 24!)?
is it the correct config file to set things?
is bmx7 included in lime-system?

FreifunkUFO:
why do we need that default-file? other communities doent have that. it seems to be configurable at an other config file.

if its possible (flash-size) please add horst (and netperf?) as an standard-package for our preset!?
also have a look at 
https://github.com/dangowrt/lime-packages/blob/3479c3ca89b337205a7e763f64de876d773c7c9f/packages/lime-altermesh/Makefile
https://github.com/dangowrt/lime-packages/blob/3479c3ca89b337205a7e763f64de876d773c7c9f/packages/lime-debug/Makefile

dangowrt:
some other communities (altermundo, qmp, ...) do have (meta-)packages
as well. others apparently use lime-build to include stuff as plain
files in the root filesystem. I prefer a package, as that would work
for all ways (buildroot, imagebuilder, lime-build or even just
installing the package at runtime).

I didn't bother to add packages yet, as I also want to use
lime-freifunk to only apply the default settings needed and not pull
any extra packages on devices with only 4 megs...

maybe we should have a '-basic' and a '-full' package with the full
package being a pure metapackage pulling in all the dependencies
(with lime-proto-olsr and all the legacy-foo as well).

p4u:
@FreifunkUFO The idea behind libre-mesh is to became a very modular firmware which might be used by any community in the world. So this is why we have packages for everything. What Daniel is doing makes sense under the project point of view.

@dangowrt Bmx7 is included in lime-proto-bmx6 (in some moment we must change the name to bmx7) which is also included by lime-full (this is the default set of packages). But lime-systems does not include it.

nicoechaniz:
On 05/28/2016 09:35 AM, Pau wrote:
> IMO all changes make sense and they are correct. But let's wait for some
> other opinion before merging it.

I also reviewed this yesterday. It seems correct to me too.

PR #49, opened 2016-07-09
customize distance setting

dangowrt:
10km is a bit to extreme and leads to duplicate packets...


G10h4ck:
Did you already tested it ?

dangowrt:
I did test it with distance_5ghz and distance_2ghz defined in my local /etc/config/lime. I'll test it while omitting either or both of them tomorrow, but expect it to work, it's quite trivial.
To make it much more robust, we could add boundary checks for all user-supplied parameters, in this case that would be some like like
`if tonumber(distance) < 20 or tonumber(distance) > 100000 then
  distance = 10000
end`

G10h4ck:
Thanks you!

PR #52, opened 2016-07-15
bmx6/bmx7 mixup cleaning

egon0:
i just cleaned the bmx6/bmx7 mixup and made bmx7 packages where it was imho appropriate


egon0:
damn, there is also a fix for the luci-mod-status package inside

PR #53, opened 2016-07-15
updated index view of luci-mod-status

egon0:
its just a copy of the original luci status index view


G10h4ck:
It is just an updated copy from upstream?
Did you tested this?

Cheers!

egon0:
yes, is tested and an updated copy from upstream.

G10h4ck:
Many thanks

PR #54, opened 2016-07-15
Bmx6 & bmx7 cleaning

egon0:
cleaned the bmx6/bmx7 mixup. bmx7 part untestet because i have not enough 8mb devices.


G10h4ck:
I was thinking on doing something like this since i introduces this mix/mess.
Many thanks!

PR #55, opened 2016-07-15
add batctl as dependency for luci-app-batman-adv

egon0:
batctl seems to be hardly needed for luci-app-batman-adv :)


G10h4ck:
#I haven't touched this part of the code, so I do prefer to wait for @p4u and @altergui to say something

Cheers!

egon0:
Example: https://github.com/libre-mesh/lime-packages/blob/develop/packages/luci-app-batman-adv/files/usr/lib/lua/luci/controller/batman.lua#L45

G10h4ck:
Thanks you!

PR #60, opened 2016-07-25
lime-system: fix firewall work-around

dangowrt:
Execute scripts in /etc/firewall.user.d/ if firewall is not installed.


p4u:
Since this is an issue which affects also the development branch I would prefer to apply it there and then merge or cherry-pick it to testing/16.07 instead of the other way around.

p4u:
Already applied.

Issue #61, opened 2016-07-25
unregister_netdevice problems on openwrt-cc


FreifunkUFO:
when using openwrt-cc instead of openwrt/lede-trunk (so its an older kernel) we should take care about batman-adv kernel problems. (seen on tplink 710v2.1)  a daily reboot would be needed to avoid loosing meshnodes, but it shell be at least "reboot -f" instead of "reboot".

with p4us build (4 days ago) i got:

dhoc_262 to become free. Usage count = 1
[ 1221.890000unregister_ne] tdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1232.030000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1242.170000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1252.310000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1262.450000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1272.590000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1282.730000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1292.870000] unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1
[ 1297.100000] IPv6: ADDRCONF(NETDEV_UP): wlan0-ap: link is not ready
[ 1297.120000] device wlan0-ap entered promiscuous mode
[ 1297.160000] br-lan: port 3(wlan0-ap) entered forwarding state
[ 1297.160000] br-lan: port 3(wlan0-ap) entered forwarding state
[ 1297.170000] IPv6: ADDRCONF(NETDEV_CHANGE): wlan0-ap: link becomes ready
[ 1297.380000] IPv6: ADDRCONF(NETDEV_UP): wlan0-adhoc: link is not ready
[ 1297.400000] wlan0-adhoc: Created IBSS using preconfigured BSSID ca:fe:00:c0:ff:ee
[ 1297.410000] wlan0-adhoc: Creating new IBSS network, BSSID ca:fe:00:c0:ff:ee
[ 1297.420000] IPv6: ADDRCONF(NETDEV_CHANGE): wlan0-adhoc: link becomes ready
[...]

Issue #50, opened 2016-07-10
Files included by chef not part of lime-build process


marvinmarnold:
> As pointed out in the LiMeCat notes, Chef adds to Libre-Mesh a few scripts.
Some of these are needed for Chef to customize the build, some others
should be included in lime-packages if we want lime-build to produce
images as good as the ones from chef.

> For example:
    /etc/uci-defaults/95_add-sshkeys
    /etc/config/lime-defaults
    /etc/chef_version
are needed for some of the customization features of chef (but the
lime-defaults file should be up to date with the one in
lime-packages)

[These files](https://github.com/libre-mesh/lime-packages/files/356154/lime-packages-missing-files.zip) should be moved into lime-packages.

FreifunkUFO:
a there it is.... i just was looking for 95_reboot-daily and couldnt found it.
the reboot command has to be discussed. i think you need at least "reboot -f", to avoid the problems with "unregister_netdevice: waiting for wlan0-adhoc_262 to become free. Usage count = 1" (problems only on openwrt-cc)

dangowrt:
@FreifunkUFO @p4u i think most of those work-arounds are obsolete and no longer needed. at least in the environments i've seen and setup for now, we are fine without them.

ilario:
Looking at the files and using common sense I see the following interesting points:

- lime-defaults file has a few differences with the one from lime-packages
- 93_ugly-fixes file adds an ap interface with the hostname as essid, this is useful for maintenance (you know exactly which node you're connecting to) and should be included in lime-system IMO
- 95_snmpd file should be included in lime-system
- 95_reboot-daily should be included in lime-system but I would rather switch to reboot weekly or monthly, daily is too often
- reset_deaf_phys.sh checks if an interface is doing traffic and if not does a scan, does it really work? If yes we could include it in lime-system

ilario:
Reading #61 seems that a **daily** `reboot -f` is needed for the coming release, maybe in the next ones monthly will be enough.

p4u:
Once in a month is not helping if the issue happens once in a day or even more often.

It looks like the problem is related to some atheros chips. So I would prefer to not include a reboot by default in all nodes. It will only hidde this problem and probably some more that we don't know yet.

First of all we should try to reproduce and identify the problem. @FreifunkUFO can you post which kind of WiFi chip are you using in your deployment?

If this problem is fixed on LEDE (kernel 4.4) I think we should try to base the release on it instead of CC.

[...]

PR #63, opened 2016-07-25
Including reboot cron command taken from Chef

ilario:
For addressing #61 a solution is a daily reboot.
The file in this pull request comes from Chef (see #50) but Chef includes it with all the lines commented out.

### DIFF
--- a/packages/lime-system/files/etc/uci-defaults/95_periodic-reboot
+++ b/packages/lime-system/files/etc/uci-defaults/95_periodic-reboot
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+### To schedule a daily reboot at 6:00am
+mkdir -p /etc/crontabs ; touch /etc/crontabs/root
+sed "\| \* \* \* .*reboot$|d" -i /etc/crontabs/root
+echo "0 6 * * * sleep 70; touch /etc/banner; reboot" >> /etc/crontabs/root
+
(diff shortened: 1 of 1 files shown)


dangowrt:
I'd rather see a focus on hunting down bugs like that rather than augmenting it with a reboot for another decade. Soon we'll have to reboot every couple of hours, then maybe every 15 minutes? At least try to find specific work-arounds for bugs which cannot be tracked down easily or only occur rarely (the above smells like a locking issue, things as simple as switching on kernel lock debugging via `CONFIG_KERNEL_PROVE_LOCKING` might provide a hint).
Please at least *try* to debug it if your goal is something stable.

ilario:
You're talking to the wrong guy, I'm not the one including this file in Libre-Mesh firmware, it is already there included by Chef. For me having this workaround is ugly, but having it outside lime-packages is much worse. I neither experienced ever the issue #61 and I have no idea how to trigger that. Just trying to get shit in the right place. Or to get a clear decision saying "this is not needed/wanted, please Chef people DON'T add it again, especially not that way".

dangowrt:
Maybe there should then be a group of lime-tweak-XXX packages for each of those files in lime-packages? They could then be individually selected for each build (both, chef and lime-build). I'm just against unconditional inclusion in lime-system.

p4u:
I'm also against including a reboot to the lime-system code. The idea of creating a packet (or several) for this kind of tweaks might be better. In any case I'm pretty sure all of us would prefer to avoid any need of reboot, so let's see if we can debug the problem.

ilario:
> The file in the chef, as you noted, is all commented out ... for a reason! 
> for people that want to enable such behaviour, for whatever reason.

You're right

> i'd vote to leave it in chef as it is

Totally against this option.
Both Chef and lime-build are official ways to obtain the firmware, they can't make images with different contents.

> i think creating a package is overkill

In my opinion the discussion should be on:

- create a package `lime-tweak-periodic-reboot` (as suggested by @dangowrt) with the uncommented file but not contained in lime-basic

or

- include the file in lime-system but in its *commented* version

G10h4ck:
let discuss this stuff here https://github.com/libre-mesh/lime-packages/issues/50 before doing pull requests

PR #64, opened 2016-07-26
lime-proto-bmx6: add tunnel announcement for node own IPv6/IPv4 

p4u:
Current bmx6 tunnel configuration announces IPv4 and IPv6 network ranges but not the own node addresses. 

Having by default an announcement for /32 and /128 addresses might be usefull for users and network administration. In a network with multiple clouds where BAT-ADV is split, BMX6 would have the information to know which is the actual best path to reach the node, if not the packet will be drop in a random entry point for the destination cloud. 

In addition it help the users to know the IPv4/IPv6 of the nodes, so in a big network where the layer2 is not propagated (thus the Hostname sharing won't be useful) it facilitates the finding of node's IPs for administration or visualization. Luci-app-bmx6 will show the IP information on the "Nodes" and "Tunnels" tabs.

I don't see any negative implications for this small change. I've tested it and works fine.

### DIFF
--- a/packages/lime-proto-bmx6/src/bmx6.lua
+++ b/packages/lime-proto-bmx6/src/bmx6.lua
@@ -6,6 +6,7 @@ local fs = require("nixio.fs")
 local libuci = require("uci")
 local wireless = require("lime.wireless")
 local opkg = require("luci.model.ipkg")
+local string = require("string")
 
 bmx6 = {}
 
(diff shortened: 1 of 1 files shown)


G10h4ck:
I remember this was done like that in the past and at some point we decided to drop those announcements for some reason @altergui @axel do you remember why ?

Cheers!

G10h4ck:
We talked with Gui why those announcements were removed, there is various reasons:
- Having each node announce it's IP make the routing table bigger, in fact we add one route per node, loosing part of the scalability gained having just one route per L2 clouds
- When an L2 cloud get broken we don't concern about reaching nodes, we just assume it is broken and we have to fix it ;)

There were more motivation probably @altergui will complement soon

PR #71, opened 2017-01-12
support IPv6 gateway detection

dangowrt:
For the beginning, make watchping aware of IPv6 hosts (probably this won't even be needed on newer busybox versions, on ChaosCalmer it still is).
Also create tested scripts which announce the IPv6 gateway to the mesh.

### DIFF
--- a/packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-fail.d/bmx6-gw
+++ b/packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-fail.d/bmx6-gw
@@ -1,0 +1,4 @@
+#!/bin/sh
+logger -t bmx6-auto-gw "Searching IPv6 Internet access into the mesh"
+bmx6 -c tunIn -inet6
+bmx6 -c tunOut inet6 /network 2000::/3 /maxPrefixLen 64
--- a/packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw
+++ b/packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw
@@ -1,0 +1,13 @@
+#!/bin/sh
+logger -t bmx6-auto-gw "We got IPv6 Internet access"
+netstatus="$(ubus call network.interface dump)"
+prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
+mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
+[ "$prefix" -a "$mask" ] || exit 0
+logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
+
+bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
+bmx6 -c tunOut -inet6
+bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=64
+bmx6 -c tunIn inet6 /n=2000::/3
+
--- a/packages/watchping/files/etc/init.d/watchping
+++ b/packages/watchping/files/etc/init.d/watchping
@@ -107,6 +107,7 @@ service_triggers() {
 		for iface in $trigger_interfaces; do
 			logger -p user.info -t "watchping" "adding ifup trigger for $iface"
 			procd_add_interface_trigger "interface.*.up" "$iface" /etc/init.d/watchping restart
+			procd_add_interface_trigger "interface.*.update" "$iface" /etc/init.d/watchping restart
 		done
 	fi
 }
--- a/packages/watchping/files/etc/uci-defaults/90_watchping
+++ b/packages/watchping/files/etc/uci-defaults/90_watchping
@@ -2,11 +2,16 @@
 
 uci -q get system.@watchping[0] || {
         uci batch <<EOF
-        add system watchping
-        set system.@watchping[0].interface=wan
-        set system.@watchping[0].timeout=2m
-        set system.@watchping[0].pinghosts=8.8.8.8
-        set system.@watchping[0].pinginterval=20s
+        set system.watchping=watchping
+        set system.watchping.interface=wan
+        set system.watchping.timeout=2m
+        set system.watchping.pinghosts=8.8.8.8
+        set system.watchping.pinginterval=20s
+        set system.watchping6=watchping
+        set system.watchping6.interface=wan6
+        set system.watchping6.timeout=2m
+        set system.watchping6.pinghosts=2001:470:20::2
+        set system.watchping6.pinginterval=20s
         commit system
 EOF
 }
--- a/packages/watchping/files/etc/watchping/wan6-fail.d/README
+++ b/packages/watchping/files/etc/watchping/wan6-fail.d/README
@@ -1,0 +1,4 @@
+# All executable files in this directory are processed
+# when watchping gets no reply through wan6 ('interface')
+# from all of the hosts listed in 'pinghosts'
+# after the time set in 'timeout'
--- a/packages/watchping/files/etc/watchping/wan6-ok.d/README
+++ b/packages/watchping/files/etc/watchping/wan6-ok.d/README
@@ -1,0 +1,3 @@
+# All executable files in this directory are processed
+# when watchping can get a reply through wan6 ('interface')
+# from any of the hosts listed in 'pinghosts'
--- a/packages/watchping/files/usr/bin/watchping
+++ b/packages/watchping/files/usr/bin/watchping
@@ -18,7 +18,8 @@ run_hooks () {
 watchping_ping() {
 	local ifname="$1"; local timeout="$2"; local pinghosts="$3"; local pinginterval="$4"; local hookname="$5"
 	local last_hooks_run="fail"
-	
+	local pingexec="ping"
+
 	if [ ! -r "/sys/class/net/$ifname/ifindex" ] ; then
 		echo "Interface $ifname unsupported or not found!"
 		exit 1
@@ -45,8 +46,13 @@ watchping_ping() {
 		time_now="${time_now%%.*}"
 		time_lastcheck="$time_now"
 
+		case "$host" in
+			*":"*)
+				[ "$(which ping6 2>/dev/null)" ] && pingexec="ping6"
+				;;
+		esac
 		for host in "$pinghosts" ; do
-			if ping -I "$ifname" -c 1 "$host" &> /dev/null ; then 
+			if $pingexec -I "$ifname" -c 1 "$host" &> /dev/null ; then
 				time_lastcheck_withinternet="$time_now"
 			else
 				time_diff="$((time_now-time_lastcheck_withinternet))"


dangowrt:
What's still missing is:
- an option /autoAdressing for bmx6 tunOut to pick tun6Address on a gateway client within the prefix of the gateway
- hooks to tell netifd on a gateway client about the delegated prefix
- tell dnsmasq to offer adresses within a delegated prefix to clients

G10h4ck:
This is already tested and working or it is a a request for comment?

p4u:
The implementation looks good, thank you.

However I would not enable this IPv6 auto gw announcements by default until this feature is implemented:

> an option /autoAdressing for bmx6 tunOut to pick tun6Address on a gateway client within the prefix of the gateway

If the mesh network uses a non-global IPv6 scheme or there is not a gateway entry point for the global IPv6 used on all nodes, the Internet announcement won't work (or will partially work only on those nodes with a valid IPv6 range).

So I propose to have an option on /etc/config/lime (thus /etc/config/lime-defaults) to enable/disable this feature. By default on generic firmware it would be disabled, but it can be enabled for specific customized firmwares such as freifunk.

dangowrt:
This **is** tested on our lime-testbed to announce the IPv6 arpanet gateway. gateway clients not having an IPv6 address within the announced prefix simply won't pick that gateway and not offer IPv6 connectivity, so nothings breaks and imho on the server side, no further changes are needed. On the gateway-client side, things should be properly integrated with netifd as there is no other mechanism to deal with acquired IPv6 prefix(es), such as to delegate sub-prefixes to local interfaces (```ip6assign``` option in ```/etc/config/network```).

dangowrt:
I particularly like the idea of having this enabled by default as IPv6 prefix delegation is a standard thing and most ISPs started handing out at least a dynamic /64 for every DOCSIS or xDSL customer. So imho we should detect and redistribute it, so if someone connects a lime node to his cable router, he'd automatically offer an IPv6 gateway to the cloud. I mean, we do NAT and a lot of unintended (but de-facto standard) things on IPv4 to allow that kind of behaviour, why not some much more transparent things to allow the same for IPv6?

p4u:
I see what you say. I did not know about "ingress6Prefix" option for bmx6/7. Is it working on both branches of bmx? If so, I agree, it can be enabled by default.

p4u:
@G10h4ck any comment on this? From my side, it can be merged.

dangowrt:
```ingress6Prefix``` works for bmx6 and bmx7. However, I only tested this feature on bmx6 and thus also didn't modify ```bmx7-auto-gw-mode``` which needs to be done to have this for bmx7 as well.

**Comment by G10h4ck** [comment]
Yet i haven't understand how the prefix delegation is handled by PR, is that the cloud configuration must be coherent with the prefix assigned by the ISP?

Cheers!

File: packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw [hunk 1]
```diff
@@ -0,0 +1,13 @@
...
+logger -t bmx6-auto-gw "We got IPv6 Internet access"
+netstatus="$(ubus call network.interface dump)"
+prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
+mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
+[ "$prefix" -a "$mask" ] || exit 0
+logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
+
+bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
+bmx6 -c tunOut -inet6
+bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=80
```
altergui: you probably meant tunOut=inet6

altergui: and why minPrefixLen=80? i'd leave that out

packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw
```diff
@@ -0,0 +1,13 @@
...
+bmx6 -c tunIn inet6 /n=2000::/3
```
altergui: use /n=::/0 here as well

packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-fail.d/bmx6-gw
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+logger -t bmx6-auto-gw "Searching IPv6 Internet access into the mesh"
+bmx6 -c tunIn -inet6
+bmx6 -c tunOut inet6 /network 2000::/3 /maxPrefixLen 96
```
altergui: /network should be ::/0 and maxPrefixLen 0  (to be consistent with current LibreMesh)

p4u: On the current generic LiMe configuration, there are two default tunnel announcements named "inet6" and "publicv6".
The first (inet6) refers to an internet Uplink (maxPrefixLen=0), the second to any global network announced on the mesh with a maximum prefix of 64bits (this is according to RIPE rules).

> bmx6.inet6=tunOut
> bmx6.inet6.tunOut='inet6'
> bmx6.inet6.network='::/0'
> bmx6.inet6.maxPrefixLen='0'
> bmx6.publicv6=tunOut
> bmx6.publicv6.tunOut='publicv6'
> bmx6.publicv6.network='2000::/3'
> bmx6.publicv6.maxPrefixLen='64'

So bmx6-auto-gw should use _/network ::/0_ and _/maxPrefixLen 0_

packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw
p4u: There is a typo. It should be _tunout=inet6_

Why minPrefixLen=80 ?

packages/bmx6-auto-gw-mode/files/etc/watchping/wan6-ok.d/bmx6-gw
p4u: That should be _/n=::/0_

p4u:
@dangowrt please see my comments on the code

dangowrt:
On Wed, Feb 15, 2017 at 03:47:55PM -0800, Pau wrote:
> p4u requested changes on this pull request.
> 
> 
> 
> > @@ -0,0 +1,4 @@
> +#!/bin/sh
> +logger -t bmx6-auto-gw "Searching IPv6 Internet access into the mesh"
> +bmx6 -c tunIn -inet6
> +bmx6 -c tunOut inet6 /network 2000::/3 /maxPrefixLen 96
> 
> On the current generic LiMe configuration, there are two default tunnel announcements named "inet6" and "publicv6".
> The first (inet6) refers to an internet Uplink (maxPrefixLen=0), the second to any global network announced on the mesh with a maximum prefix of 64bits (this is according to RIPE rules).

If we apply RIPE rules (ie. minPrefixLen=64) this is what most ISPs do,
unfortunately. Ie. most people got a /64 falling out of their cable or
ADSL. They can't redistribute a /64, obviously.
I reckon the requirement comes from the idea of just announcing RAs and
having clients to stateless self-addressing with EUI-64, so no DHCPv6
is needed. Actually, that's all some clients (Android!) can do.
However: EUI-64 implies a privacy hell. You'll be leaking the MAC
address of your device to the web unless you got IPv6 privacy switched
on and all your applications grab the right socket to initiate
connections. And not get tainted by WebRTC and other crap tech which
will once again expose the EUI-64 generated address which contains the
MAC address.
To be at least on the same level of privacy as IPv4, we can use DHCPv6
on a short prefix (on IPv4 we got 16-bits of freedom in a 10.x.0.0/16
subnet -- I thought having twice of that for IPv6, ie. minPrefixLen=96
with thus still at least 32 bits of freedom is more than enough)


> 
> > bmx6.inet6=tunOut
> > bmx6.inet6.tunOut='inet6'
> > bmx6.inet6.network='::/0'
> > bmx6.inet6.maxPrefixLen='0'
> > bmx6.publicv6=tunOut
> > bmx6.publicv6.tunOut='publicv6'
> > bmx6.publicv6.network='2000::/3'
> > bmx6.publicv6.maxPrefixLen='64'
> 
> So bmx6-auto-gw should use _/network ::/0_ and _/maxPrefixLen 0_

Hm, I get the idea. But what watchping tests for is actually the
2000::/3 network supposedly being reachable. And I like that much more,
because it's not a 'default gateway' but rather an 'ARPAnet gateway'
right? IPv6 allows for that destinction, and that's nice.

> 
> > @@ -0,0 +1,13 @@
> +#!/bin/sh
> +logger -t bmx6-auto-gw "We got IPv6 Internet access"
> +netstatus="$(ubus call network.interface dump)"
> +prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
> +mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
> +[ "$prefix" -a "$mask" ] || exit 0
> +logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
> +
> +bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
> +bmx6 -c tunOut -inet6
> +bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=80
> 
> There is a typo. It should be _tunout=inet6_

I think both work. I basically used what Axel sent me in an earlier
email (and it works).

> 
> Why minPrefixLen=80 ?

So there'll be a bunch of /80 prefixes to hand-out to clouds even if
the ISP only gave me a /64 (that's thus about 2^16 prefixes for
clouds, should be more than enough).


Cheers


Daniel

> 
> > @@ -0,0 +1,13 @@
> +#!/bin/sh
> +logger -t bmx6-auto-gw "We got IPv6 Internet access"
> +netstatus="$(ubus call network.interface dump)"
> +prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
> +mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
> +[ "$prefix" -a "$mask" ] || exit 0
> +logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
> +
> +bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
> +bmx6 -c tunOut -inet6
> +bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=80
> +bmx6 -c tunIn inet6 /n=2000::/3
> 
> That should be _/n=::/0_

See above. If you like the idea of 'default gateway' then yes...

p4u:
On 19/02/17 10:51, Daniel Golle wrote:
> On Wed, Feb 15, 2017 at 03:47:55PM -0800, Pau wrote:
>> p4u requested changes on this pull request.
>>
>>
>>
>>> @@ -0,0 +1,4 @@
>> +#!/bin/sh
>> +logger -t bmx6-auto-gw "Searching IPv6 Internet access into the mesh"
>> +bmx6 -c tunIn -inet6
>> +bmx6 -c tunOut inet6 /network 2000::/3 /maxPrefixLen 96
>>
>> On the current generic LiMe configuration, there are two default tunnel announcements named "inet6" and "publicv6".
>> The first (inet6) refers to an internet Uplink (maxPrefixLen=0), the second to any global network announced on the mesh with a maximum prefix of 64bits (this is according to RIPE rules).
> 
> If we apply RIPE rules (ie. minPrefixLen=64) this is what most ISPs do,
> unfortunately. Ie. most people got a /64 falling out of their cable or
> ADSL. They can't redistribute a /64, obviously.
> I reckon the requirement comes from the idea of just announcing RAs and
> having clients to stateless self-addressing with EUI-64, so no DHCPv6
> is needed. Actually, that's all some clients (Android!) can do.
> However: EUI-64 implies a privacy hell. You'll be leaking the MAC
> address of your device to the web unless you got IPv6 privacy switched
> on and all your applications grab the right socket to initiate
> connections. And not get tainted by WebRTC and other crap tech which
> will once again expose the EUI-64 generated address which contains the
> MAC address.
> To be at least on the same level of privacy as IPv4, we can use DHCPv6
> on a short prefix (on IPv4 we got 16-bits of freedom in a 10.x.0.0/16
> subnet -- I thought having twice of that for IPv6, ie. minPrefixLen=96
> with thus still at least 32 bits of freedom is more than enough)
> 

I don't like the idea of not being RIPE compatible on this sense but I
get your point. I don't have any real experience on DSL-IPv6. So I would
change the maxPrefixLen to 96 but it must be consistent with the rest of
LiMe. So if we change it here we have to change it also on the bmx6/bmx7
configuration scripts.

https://github.com/libremesh/lime-packages/blob/develop/packages/lime-proto-bmx6/src/bmx6.lua

>>
>>> bmx6.inet6=tunOut
>>> bmx6.inet6.tunOut='inet6'
>>> bmx6.inet6.network='::/0'
>>> bmx6.inet6.maxPrefixLen='0'
>>> bmx6.publicv6=tunOut
>>> bmx6.publicv6.tunOut='publicv6'
>>> bmx6.publicv6.network='2000::/3'
>>> bmx6.publicv6.maxPrefixLen='64'
>>
>> So bmx6-auto-gw should use _/network ::/0_ and _/maxPrefixLen 0_
> 
> Hm, I get the idea. But what watchping tests for is actually the
> 2000::/3 network supposedly being reachable. And I like that much more,
> because it's not a 'default gateway' but rather an 'ARPAnet gateway'
> right? IPv6 allows for that destinction, and that's nice.

Same of before. I also like the idea of using 2000::/3 instead of ::/0
but it must be compatible with the rest of LiMe

>>
>>> @@ -0,0 +1,13 @@
>> +#!/bin/sh
>> +logger -t bmx6-auto-gw "We got IPv6 Internet access"
>> +netstatus="$(ubus call network.interface dump)"
>> +prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
>> +mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
>> +[ "$prefix" -a "$mask" ] || exit 0
>> +logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
>> +
>> +bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
>> +bmx6 -c tunOut -inet6
>> +bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=80
>>
>> There is a typo. It should be _tunout=inet6_
> 
> I think both work. I basically used what Axel sent me in an earlier
> email (and it works).

Mmm, you are first deleting a tunOut named "inet6" and afterwards adding
a new one named just "6". So maybe it end up working but it is not correct.

>>
>> Why minPrefixLen=80 ?
> 
> So there'll be a bunch of /80 prefixes to hand-out to clouds even if
> the ISP only gave me a /64 (that's thus about 2^16 prefixes for
> clouds, should be more than enough).
> 

Ok, here you are the experienced with DSL-IPv6, so I would accept this
/80 if there is not any other opinion against it.

> Cheers
> 

Thank you!

> Daniel
> 
>>
>>> @@ -0,0 +1,13 @@
>> +#!/bin/sh
>> +logger -t bmx6-auto-gw "We got IPv6 Internet access"
>> +netstatus="$(ubus call network.interface dump)"
>> +prefix="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['address']" )"
>> +mask="$( jsonfilter -s "$netstatus" -e "@.interface[*]['ipv6-prefix'][*]['mask']" )"
>> +[ "$prefix" -a "$mask" ] || exit 0
>> +logger -t bmx6-auto-gw "Announcing prefix $prefix/$mask to the mesh"
>> +
>> +bmx6 -c tunDev main /tun6Address ${prefix}1/$mask /ingress6Prefix ${prefix}/$mask
>> +bmx6 -c tunOut -inet6
>> +bmx6 -c tunOut=6 /n=${prefix}/${mask} /minPrefixLen=80
>> +bmx6 -c tunIn inet6 /n=2000::/3
>>
>> That should be _/n=::/0_
> 
> See above. If you like the idea of 'default gateway' then yes...
> 
> 
> 
>>
>> -- 
>> You are receiving this because you were mentioned.
>> Reply to this email directly or view it on GitHub:
>> https://github.com/libremesh/lime-packages/pull/71#pullrequestreview-22141408
> 
> 

-- 
./p4u

G10h4ck:
Are you sure ipv6 autoconfiguation works with prefix smaller then /64 ?
When i tried this some years ago most devices just ignored the announced prefix as it was smaller then /64... other where taking the configuration but like it was announced as a /64 the result was strange behaving non working network...

please double check this too

cheers!

p4u:
Let's try to close this PR @dangowrt @G10h4ck

dangowrt:
@G10h4ck no, IPv6 stateless autoconfiguration requires at least a /64 to work. DHCPv6 works fine even with a /120, ie. only 8 bits for the host address within a prefix, so that shouldn't limit people from redistributing an IPv6 prefix. **All** ISPs I know give exactly one /64 to non-business customers and those may still want to offer a gateway without having to setup tunnels or VPNs. If you got a larger (ie. /48 or shorter) prefix, that will allow you to delegate several /64 further down the line, however, imho this should not be made a requirement (as that would make it ridiculously hard for people to share access to the IPv6 ARPAnet). Mesh networks should be a grass-roots thing, entrance barrier to start a new mesh somewhere should be as low as possible and not depend on big organizations which are RIPE members like guifi or funkfeuer...

altergui:
i agree that giving a /64 to non-business customers is bad practice, and share the feeling to try to workaround that. But AFAIU we can't do much since client support is bad (i.e. android, as you said)

i mean, even if we merge this in libremesh, and then if you get a /64, you can distribute smaller prefixes, android devices won't get IPv6 right (or even break, as gio mentioned)

anyway, my suggestion is to split this into two parts: modify this pull request so it's "non-polemic", it only adds support respecting upstream decisions (basically follow pau's comments)
and then, you/we can make another PR discussing the possibility to modify those upstream decisions (not only in this script, but everywhere) like maxPrefixLen=96 instead of current 64,
and whether tunOut inet6 should point to 2000::/3 or current ::/0, etc
(personally, i agree with maxPrefixLen=96 but not with 2000::/3, but let's discuss that elsewhere)

dangowrt:
Fair enough. I believe that odhcpd was fixed to not send Router Advertisements on prefixes longer than /64 because Stateless Autoconfiguration wouldn't work anyway. I'm not sure about dnsmasq, however, relying in DHCPv6 should be fine. Clients not supporting DHCPv6 will then just come up without IPv6 if the prefix is too long and that's not worse than refraining from distributing that prefix at all from the point of view of those (horrible) Stateless-Autoconfiguration-only IPv6 client devices.
For Android, this has been a lengthy and still ongoing [debate](https://code.google.com/p/android/issues/detail?id=32621) as it renders IPv6 to be unusable in quite a lot of environments. Imho broken clients should not be taken as a reference, as then we'll end up in an even more broken world by the end of the day.
So what I will do:
- Fix things according to what @p4u said (typos, naming of the tunOut)
- As watchping detects a route to the `publicv6`, should we just name the tunIn instance `publicv6`? Or you are insisting in an actual **default gateway** with a route to `::/0`
- minPrefixlen and maxPrefixlen are two different things, read in the BMX6/7 docs about it. so minPrefixlen=56 and maxPrefixlen=64 is probably close to the RIPE suggested standards but I wouldn't even be able to test it myself, because all we got is also just consumer-grade IPv6 with a /64 falling out of DOCSIS and DSL lines from mainstream ISPs.

p4u:
`As watchping detects a route to the publicv6, should we just name the tunIn instance publicv6? Or you are insisting in an actual default gateway with a route to ::/0`

I agree with @dangowrt here. It makes more sense to use 2000::/3  instead of ::/0. But we have to change it also here https://github.com/libremesh/lime-packages/blob/develop/packages/lime-proto-bmx6/src/bmx6.lua#L68
@altergui what is your point for not accepting 2000::/3 as route to the Internet?


About the IPv6 prefix delegation. If we are going to accept IPv6 prefixes smaller than /64 we have to implement some mechanism to configure Dnsmasq properly and disable RA announcements on the nodes (remember that in LibreMesh all nodes are serving IP configuration). But in any case, IMO it is fine to accept this PR which only affects bmx6. All the stuff working right now will be also working with this PR commited. So I see no point for keeping this open.

p4u:
That is actually what @dangowrt was saying at the beginning of this infinite discussion

> What's still missing is:
>     an option /autoAdressing for bmx6 tunOut to pick tun6Address on a gateway client within the prefix of the gateway
>     hooks to tell netifd on a gateway client about the delegated prefix
>     tell dnsmasq to offer adresses within a delegated prefix to clients

First point has to be implemented on bmx6, so it depends on @axn
Second depends on first.
Third can be implemented right now after or before merging this PR

G10h4ck:
> what is your point for not accepting 2000::/3 as route to the Internet?

For example any sort of IPv4 mapped in IPv6 address is not included in that subnet, and I believe is a good idea to keep compatibility with that sort of mechanisms, just to do an example at FOSDEM 2017 the network was IPv6 only and they provided access to some sort of IPv4 internet via those mapping mechanisms, so imagine we were doing a LibreMesh workshop there...

dangowrt:
> > what is your point for not accepting 2000::/3 as route to the Internet?
> For example any sort of IPv4 mapped in IPv6 address is not included in that subnet, and I believe is a good idea to keep compatibility with that sort of mechanisms, just to do an example at FOSDEM 2017 the network was IPv6 only and they provided access to some sort of IPv4 internet via those mapping mechanisms, so imagine we were doing a LibreMesh workshop there...

Those mechanisms (I reckon you refer to RFC6877 aka 464xlat) are only available on some gateways v6-only, so we should probably have several watchping sections testing for them. Ie. pinging a well-known IPv6-mapped IPv4 address and setting up explicit tunIn for them.
Imho relying that if an arbitary host within the 2000::/3 prefix works that those mappings are also offered on that gateway (which maybe a true dual-stack thing without any support for accessing v4 services via v6) is not a good idea.
Also note that there are several methods, 464xlat being just one of them.
Imho dual-stack with NAT'ed IPv4 (just like we do now) is the most convenient and backward-compatible choice, as 464xlat also has several disadvantages (we may discuss them on the mailing list to get more opinions, if there is interest).

G10h4ck:
> Those mechanisms (I reckon you refer to RFC6877 aka 464xlat) are only available on some gateways
> v6-only, so we should probably have several watchping sections testing for them. Ie. pinging a
> well-known IPv6-mapped IPv4 address and setting up explicit tunIn for them.

This solution seems ok to me, would you open an issue to keep track of this ?

p4u:
By the way, from dnsmasq MAN page http://www.thekelleys.org.uk/dnsmasq/docs/dnsmasq-man.html

> For IPv6, the parameters are slightly different: instead of netmask and broadcast address, there is an optional prefix length which must be equal to or larger then the prefix length on the local interface. If not given, this defaults to 64. Unlike the IPv4 case, the prefix length is not automatically derived from the interface configuration. The mimimum size of the prefix length is 64. 

So dnsmasq does NOT support prefix length smaller than 64. Either we have to change to another DHCPv6 software (quite hard, too many depending things on dnsmasq...) or we patch dnsmasq to allow smaller prefixes.

dangowrt:
maybe we should consider odhcpd instead of dnsmasq? we don't seem to
use the dns part of it anyway...
the ipv6 support in recent versions of dnsmasq being broken is clearly
an additional argument for that move.

p4u:
Let's keep this discussion in the mailing list thread with subject: 
"Open discussion about IPv6 delegation prefix in LibreMesh"

dangowrt:
ping @p4u

aparcar:
ping @p4u @dangowrt

ilario:
Check out #299

gmarcos87:
This pull request was open for a long time without any novelty. Can we close it? Is it still relevant?

aparcar:
might be resolved within the next few days :)

dangowrt:
It works equally for bmx7, we are just waiting for the client side to be implemented (https://github.com/bmx-routing/bmx7/commit/79b2332dba412c22db061414fdcf5bc196a70d40)

dangowrt:
I hope that @axn will add capabilities for IPv6 prefix delegation at some point...

spiccinini:
Closing as reported most of the work is already merged. Please reopen if the code on the PR is still needed and has to be merged.

PR #78, opened 2017-03-13
Add line for psk in example of private ap mode configuration

ilario:
I didn't test this but the encryption type configuration should be needed in the example of a private ap configuration.


p4u:
Thanks. I also noticed of this mistake but I never fixed it.

PR #113, opened 2017-04-06
Fix batman-adv Originators Status WebUI.

### LINKED ISSUE #110: Fix batman-adv Status WebUI
Expected behaviour: see nodes

Actual behaviour: nodes are not being displayed

nicopace:
It solves part of the issue #110.
I have no environment to test the Gateways section, just the Originators one.

### DIFF
--- a/packages/luci-app-batman-adv/files/usr/lib/lua/luci/controller/batman.lua
+++ b/packages/luci-app-batman-adv/files/usr/lib/lua/luci/controller/batman.lua
@@ -14,6 +14,26 @@ $Id$
 
 module("luci.controller.batman", package.seeall)
 
+local function split(str, pat)
+   local t = {}  -- NOTE: use {n = 0} in Lua-5.0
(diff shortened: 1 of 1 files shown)


p4u:
@altergui should review this pull request

ilario:
Can't this info be taken from [alfred](https://git.open-mesh.org/alfred.git/blob_plain/refs/heads/master:/README) and its utility [batman-vis](https://downloads.open-mesh.org/batman/manpages/batadv-vis.html)?
Parsing the output of `batctl -o` sounds like a bad idea...

nicopace:
I don't know... I just fixed what was broken... if there is a better way to do it, we could do it after we get fixed what was broken ;)

p4u:
So @altergui is not responding and I don't see any point to not accept this fix.

Thanks.

nicopace:
@p4u It solves part of the issue #110.
I have no environment to test the Gateways section, just the Originators one.

PR #122, opened 2017-05-03
Changes the way bmx6 hosts are used, adding them to the /etc/hosts file.

nicopace:
This pull request changes the way the bmx6hosts is used so it is more reliable.
Before this change, the bmx6hosts were unreliable, it woked most of the time, but sometimes it did not (random).
Using this strategy we guarantee that we always resolve locally, without depending on dnsmasq.

### DIFF
--- a/packages/lime-proto-bmx6/src/bmx6hosts_dnsmasq_update
+++ b/packages/lime-proto-bmx6/src/bmx6hosts_dnsmasq_update
@@ -1,11 +1,5 @@
 #!/bin/sh
 bmx6hosts > /var/hosts/bmx6
-new=/tmp/bmx6hosts.new.md5
-last=/tmp/bmx6hosts.last.md5
-touch $last
-md5sum /var/hosts/bmx6| awk '{print $1}' > $new 
-[ "$(cat $new)" != "$(cat $last)" ] && {
-	logger -t bmx6hosts "New bmx6 hosts found, updating dnsmasq"
-	killall -HUP dnsmasq
-}
-cp -f $new $last
+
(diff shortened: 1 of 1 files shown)


p4u:
I don't like the idea of writing to the flash memory every 5 minutes. No process from LEDE (AFAIK) does it. All of them are configured to write always on tmpFS (which is in RAM). The only stuff written on the flash memory are the configurations, but of course not periodically every N minutes.

So before breaking this rule, I would prefer to find another solution to this problem. Maybe now with "smonit" we can find when dnsmasq is not resolving properly and execute the required actions.

https://github.com/libremesh/lime-packages/tree/feature/smonit

nicopace:
@nicoechaniz can you comment on this?

G10h4ck:
I agree with @p4u that writing to the flash is a no go

nicoechaniz:
Ok, after reading a bit more about this matter:
https://wiki.openwrt.org/doc/techref/flash
https://forum.openwrt.org/viewtopic.php?id=55982

I also agree this is not a good solution.

I'm aware that if dnsmasq were working stable, we wouldn't need to take any action regarding this issue. But the fact is that it is not stable and I cannot remember a release where it was really stable.

So... I think it's important that we consider alternatives to this proposed pull request.

There are some ideas in this stackexchange topic:
https://unix.stackexchange.com/questions/60549/etc-hosts-file-refer-to-another-configuration-file

All of which involve some trick... the most simple would be to have /etc/hosts link to a /tmp/hosts file where this or other scripts would write. We could even write to /tmp/hosts.d/* and consolidate this data in /tmp/hosts every X minutes.

The link should always be recreated at boot time.

G10h4ck:
This should be solved in a better way by https://github.com/libremesh/lime-packages/pull/142

PR #126, opened 2017-05-18
Feature/dumped config

G10h4ck:
/etc/config/lime is populated with all options used, this shoud facilitate web interface creation and options tweaking

### DIFF
--- a/packages/lime-system/files/etc/config/lime-defaults
+++ b/packages/lime-system/files/etc/config/lime-defaults
@@ -24,6 +24,8 @@ config lime network
 	list resolvers 4.2.2.2   # b.resolvers.Level3.net
 	list resolvers 141.1.1.1 # cns1.cw.net
 	list resolvers 2001:470:20::2 # ordns.he.net
+	option bmx6_over_batman false
+	option bmx7_over_batman false
 
 config lime wifi
 	option channel_2ghz '11'
--- a/packages/lime-system/files/usr/lib/lua/lime/config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/config.lua
@@ -18,7 +18,13 @@ function config.get(sectionname, option, default)
 	if limeconf then return limeconf end
 
 	local defcnf = config.uci:get("lime-defaults", sectionname, option, default)
-	config.set(sectionname, option, defcnf)
+	if ( defcnf ~= nil ) then
+		config.set(sectionname, option, defcnf)
+	else
+		local cfn = sectionname.."."..option
+		print("WARNING: Attempt to access undeclared default for: "..cfn)
+		print(debug.traceback())
+	end
 	return defcnf
 end
 
--- a/packages/lime-system/files/usr/lib/lua/lime/config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/config.lua
@@ -14,7 +14,12 @@ config = {}
 config.uci = libuci:cursor()
 
 function config.get(sectionname, option, default)
-	return config.uci:get("lime", sectionname, option) or config.uci:get("lime-defaults", sectionname, option, default)
+	local limeconf = config.uci:get("lime", sectionname, option)
+	if limeconf then return limeconf end
+
+	local defcnf = config.uci:get("lime-defaults", sectionname, option, default)
+	config.set(sectionname, option, defcnf)
+	return defcnf
 end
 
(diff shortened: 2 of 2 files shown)


packages/lime-hwd-openwrt-wan/src/openwrt_wan.lua
```diff
@@ -9,7 +9,9 @@ local openwrt_wan = {}
 openwrt_wan.sectionName = hardware_detection.sectionNamePrefix.."openwrt_wan"
 
 function openwrt_wan.clean()
-	if config.autogenerable(openwrt_wan.sectionName) then config.delete(openwrt_wan.sectionName) end
+	if config.autogenerable(openwrt_wan.sectionName) then
+		config.delete(openwrt_wan.sectionName)
```
nicopace: This change has nothing to do with the commit. I would remove it.

packages/lime-system/files/etc/config/lime-defaults
```diff
@@ -24,6 +24,8 @@ config lime network
 	list resolvers 4.2.2.2   # b.resolvers.Level3.net
 	list resolvers 141.1.1.1 # cns1.cw.net
 	list resolvers 2001:470:20::2 # ordns.he.net
+	option bmx6_over_batman false
+	option bmx7_over_batman false
```
nicopace: Is there any way of checking this are all the defaults that need to be? Just to ensure all are defined

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -1,11 +1,9 @@
 #!/usr/bin/lua
 
---! LibreMesh is modular but this doesn't mean parallel,
---! modules are executed sequencially, so we don't need
---! to worry about transaction and all other stuff that
---! affects parrallels database, at moment we don't need
---! parallelism as this is just some configuration stuff
---! and is not performance critical.
+--! LibreMesh is modular but this doesn't mean parallel, modules are executed
```
nicopace: This is a code style change that has nothing to do with the pull request. I suggest to remove it.

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -14,24 +12,39 @@ config = {}
...
-	if section_exists then
+	if lime_section or lime_def_section then
 		local ret = lime_section or {}
 
-		if lime_default_section then
-			for key,value in pairs(lime_default_section) do
+		if lime_def_section then
+			for key,value in pairs(lime_def_section) do
 				if (ret[key] == nil) then
+					config.set(sectionname, key, value)
```
nicopace: I understand that as we are migrating from one state where lime-defaults have all the options and lime does not, but if we are migrating to a context where lime has all the options, this write is unnecesary.
We should think of an strategy of migration from one config to the other, or add a deprecation tag to this code so we don't leave it for life here.

packages/lime-system/files/etc/config/lime-defaults
G10h4ck: When you run lime-config it will print warnings on the shell if some default is missing

packages/lime-system/files/usr/lib/lua/lime/config.lua
G10h4ck: This is necessary at least at first run, ATM /etc/config/lime do not ship default configuration, but someway I agree with you that at this point we should talk if /etc/config/lime-still make sense

G10h4ck:
Thanks for reviewing, I have done the requested changes

packages/lime-system/files/usr/lib/lua/lime/config.lua
nicopace: we could add something to lime-config a check for this, and tell the user that this is happening in his connfig, and if he wants us to update the file

G10h4ck: I believe this goes outside the scope of this PR, the aim of this was to make it easier to develop the web interface as the configuration are all listed in /etc/config/lime

packages/lime-system/files/etc/config/lime-defaults
nicopace: Agree... but we do have to consider this... will add it as an issue

nicopace:
Thanks Gio, good to go!

PR #139, opened 2017-05-27
Fix lime-example disappeared from lime-system #125

G10h4ck:
This reverts commit 3e3c0f82feeabc82244cc210653d711222af66e9.

Issue #125, opened 2017-05-18
lime-example disappeared from lime-system


G10h4ck:
After this commit 3e3c0f82feeabc82244cc210653d711222af66e9 the /etc/config/lime-example file is not present anymore in lime-system.

This file is the only documentation available to the user in case of internet loss, moreover it is very similar to lime-default so i believe that in the squashfs compression it ends up requiring almost no space even in 4MB images.

The decision to have this file included was taken a bunch of assemblies ago under @ilario suggestion, now unless there are proof that the image size is significantly reduced by 3e3c0f82feeabc82244cc210653d711222af66e9 i vouch to revert it.

@p4u do you have clues on how much bytes the final squashfs image is reduced by this commit?

p4u:
On 18/05/17 11:14, Gioacchino wrote:
> After this commit 3e3c0f82feeabc82244cc210653d711222af66e9 the /etc/config/lime-example file is not present anymore in lime-system.
> 
> This file is the only documentation available to the user in case of internet loss, moreover it is very similar to lime-default so i believe that in the squashfs compression it ends up requiring almost no space even in 4MB images.

To complete your summary: the file is still available on 8MB targets,
but now it is part of the package lime-debug.

> The decision to have this file included was taken a bunch of assemblies ago under @ilario suggestion, now unless there are proof that the image size is significantly reduced by 3e3c0f82feeabc82244cc210653d711222af66e9 i vouch to revert it.
> 
> @p4u do you have clues on how much bytes the final squashfs image is reduced by this commit?
> 

I don't know how compressed is it, however in the squashFS of a
libremesh node, 'ls -s' says 7kB (on a non-compressed FS it is 12kB). In
the lime-mini images (4MB) the free space is around 70-80 kB, so we are
talking about almost 10% of the free space. I think it's worth, at least
until we find the way to get rid of some kB more.

In addition (this is another discussion) I don't think /etc/config is a
good place for adding this example file, since UCI processes it. So I
would move it somewhere else also if it is part of lime-debug.

-- 
./p4u

ilario:
I think that some documentation is needed in the firmware. 
lime-example is a bare minimum and 7 kB are worth for sure.
If the problem is the location we can move it and point out the new location in the banner.
For example we could expand the `/etc/config/lime-example` file into a `/README-LibreMesh` documentation file or move it into a `/lime-docs/` folder.

p4u:
[...]

PR #140, opened 2017-05-30
Copy documentation from lime-web to the firmware

ilario:
`lime-docs` downloads all the text files originating pages in http://libremesh.org/docs and puts it in `/docs` directory in firmware root;

`lime-docs-minimal` is thought to replace `/etc/config/lime-example`, following 3e3c0f8, #125 and #139. So it copies just the config.txt file from the aforementioned direction and puts it in the `/docs` directory in the firmware root.

I just checked that the image compiles ok, I can do test on a real router next week.


G10h4ck:
The point made by Pau seems good to me, if the documentation is accessible also via HTTP is very good!

ilario:
I applied the suggestions, please review.
Still have to test on real device.
I have a doubt: is it acceptable to introduce the subversion dependency for the host compiling the image? Or is it already a dependency of other LEDE stuff?

G10h4ck:
subversion it's already a build dependency of LEDE/OpenWrt

p4u:
Looks good to me.

Thanks!

Issue #137, opened 2017-05-25
Tiny monitor daemon


p4u:
In the last versions of libremesh, dnsmasq is not started properly or it simply dies in run time. Also bmx6 experienced some unexpected crashes in the past.

A small daemon might be used for running tests every N minutes on different processes, and execute a set of orders to fix (such as restart the process) the wrong state of it. Also a log might be written somewhere in the system so errors can by analysed afterwards.

p4u:
Current development at branch https://github.com/libremesh/lime-packages/tree/feature/smonit

ilario:
This would solve #111

p4u:
Yes.
But it will solve also more scenarios. @nicoechaniz from quintanalibre reports that some times dnsmasq just stop working but the process is still alive. This might be checked using smonit.

G10h4ck:
This cases should be handled by OpenWrt/LEDE init system, we should not add another custom init daemon that may eventually crash itself, we should have this solved in upstream init system

p4u:
I agree, as many things may be handled by procd/init LEDE system, the better. But:

1.  It depends on another team, so at the end accept or deny a pull request is up to them
2.  If we launch a release, we cannot wait for their approval, so until the problem is fixed upstream, we have to deal with our own tools
3. The reality is that this daemon monitoring is currently done by many network communities, but performed by a uci script which writes at crontab (see https://github.com/libremesh/network-profiles/tree/master/quintanalibre.org.ar/comun/etc/uci-defaults). So, this is happening and hiding it (by not adding them to the official source) does not help at all. Smonit will be a better way to handle these situations and will bring the possibility to reuse the same code for the different communities.

In addition, smonit is not a daemon (my previous definition was not accurate) but a small script executed by cron to check states and perform actions according these states. It's a small layer between crontab and the system based on hooks which might be enabled or disabled depending on system and user needs.

Not all cases that can be covered by smonit can by covered by procd/init system. While the first might only check if the daemon PID is alive, the second can execute also some actions to detect a wrong behaviour, i.e  dnsmasq is alive but it is not resolving hosts. 

Also the logs handling (when crash or wrong behaviour) is something that smonit can take care of. This will help the developers to debug the current problems.

p4u:
Another good example is this one: https://github.com/libremesh/network-profiles/blob/master/quintanalibre.org.ar/comun/usr/sbin/reset_deaf_phys.sh

[...]

PR #142, opened 2017-06-05
Feature/smonit

p4u:
Information at issue: _Tiny monitor daemon_ #137

### DIFF
--- a/packages/smonit/files/etc/init.d/smonit
+++ b/packages/smonit/files/etc/init.d/smonit
@@ -5,13 +5,13 @@ SMONIT="/usr/sbin/smonit"
 CRONTAB="/etc/crontabs/root"
 CRON_LINE='*/5 * * * *'
 
-start() {
+enable() {
   echo "Enabling smonit to crontab"
   grep -q "$SMONIT" $CRONTAB || echo "$CRON_LINE $SMONIT" >> $CRONTAB
   /etc/init.d/cron reload
 }
 
-stop() {
+disable() {
   echo "Disabling smonit from crontab"
   cat $CRONTAB | grep -v "$SMONIT" > /tmp/cron.tmp
   cp -f /tmp/cron.tmp $CRONTAB
--- a/packages/smonit/files/etc/init.d/smonit
+++ b/packages/smonit/files/etc/init.d/smonit
@@ -1,0 +1,22 @@
+#!/bin/sh /etc/rc.common
+
+START=99
+SMONIT="/usr/sbin/smonit"
+CRONTAB="/etc/crontabs/root"
+CRON_LINE='*/5 * * * *'
+
+start() {
+  echo "Enabling smonit to crontab"
+  grep -q "$SMONIT" $CRONTAB || echo "$CRON_LINE $SMONIT" >> $CRONTAB
+  /etc/init.d/cron reload
+}
+
+stop() {
+  echo "Disabling smonit from crontab"
+  cat $CRONTAB | grep -v "$SMONIT" > /tmp/cron.tmp
+  cp -f /tmp/cron.tmp $CRONTAB
+}
+
+reload() {
+  $SMONIT
+}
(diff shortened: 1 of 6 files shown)


nicopace:
I couldn't test it, but I proofread it and it seems ok :)

G10h4ck:
I have reviewed the code and made comments inline, beside to that i believe that checkers like tha one of bmx6 should be split in their own packages, like we do with lime-proto-*

packages/smonit/files/etc/smonit/dnsmasq
```diff
@@ -0,0 +1,33 @@
+#!/bin/sh
+
+hook_name="dnsmasq"
+
+function hook_is_installed() {
+  installed=no
+  [ -x /usr/sbin/dnsmasq ] && [ -x /etc/init.d/dnsmasq ] && installed=yes                                         
+}   
+
+function hook_check() {
```
p4u: It is not enough. It is just an "initial" hook (almost an exampe), to help understand the concept.

packages/smonit/files/etc/smonit/dnsmasq
```diff
@@ -0,0 +1,33 @@
...
+  }
+}
+
+function hook_quickfix() {
+  log="Restarting dnsmasq process"
+  PATH="/usr/sbin:/usr/bin:/sbin:/bin"
+  /etc/init.d/dnsmasq restart
+}
+
+function hook_longfix() {
```
p4u: True. It was an example to help understand the concept. But I agree on changing the log message.

packages/smonit/files/etc/init.d/smonit
```diff
@@ -0,0 +1,22 @@
...
+CRONTAB="/etc/crontabs/root"
+CRON_LINE='*/5 * * * *'
+
+start() {
+  echo "Enabling smonit to crontab"
+  grep -q "$SMONIT" $CRONTAB || echo "$CRON_LINE $SMONIT" >> $CRONTAB
+  /etc/init.d/cron reload
+}
+
+stop() {
```
p4u: This is right, I'll fix it.

packages/smonit/files/etc/smonit/bmx6
```diff
@@ -0,0 +1,26 @@
...
+  [ -n $bmx6pid ] && [ -d /proc/$bmx6pid ] && status=ok
+}
+
+function hook_quickfix() {
+  log="Restarting bmx6 process"
+  PATH="/usr/sbin:/usr/bin:/sbin:/bin"
+  /etc/init.d/bmx6 restart
+}
+
+function hook_longfix() {
```
p4u: True. It was an example to help understand the concept. But I agree on changing the log message.

dangowrt:
Which functionality is provided by smonit which isn't already provided by procd? I'm wondering because to me it looks like everything smonit does can already be done using the procd 'respawn' parameter. Can you enlighten me?

p4u:
See #137

I agree, everything that might be controlled by procd must be moved there. The two snippets added are kind of examples. The idea is to extend them to for instance:

- bmx6: check if there is tunnel connectivity with the nodes (using ping), sometimes we deteced an extrange bug which has not yet been identified. Smonit can help to identify it

- dnsmasq: some times it is still alive but it does not resolve. Smonit might try to resolve a hostname and execute actions if it does not work. 

Can these two things be handled by procd?

dangowrt:
No, procd can handle restartes based on ubus events and config changes and automaticly respawn a process when it died. Typically ubus events from netifd are used for the restarts, such as interface link state changes. Another thing already possible is to tie services in such way that the depending services are restarted once the depended-on service is (such as restarting dnsmasq when ever bmx6 is or the similar stuff you do in [hotfix/bmx6-restart-watchping](https://github.com/libremesh/lime-packages/tree/hotfix/bmx6-restart-watchping) can all be done using procd triggers).
The (in my opinion) ideal pattern to solve the examples you described above would be to have a daemon which statefully monitors things (like smonit does now) but rather emmits ubus events than directly restarting services -- then you can either attach procd triggers to it or simply write scripts which fire on occurance of an object or event using `ubus wait_for` or `ubus listen`.

Issue #108, opened 2017-03-31
Remove any-MAC functionality on network interfaces


nicopace:

G10h4ck:
Asi explicado no estoy del todo convencido, no veo claro como implementarlo y las implicaciones...

nicopace:
@nicoechaniz could you give @G10h4ck context for this issue?

p4u:
@Gui knows better about this issue, but the problem is mainly created by
the physical switch of the network devices such as Tplink 3500/3600/4300.

When using anygw in Ethernet and two nodes are connected through cable,
the same MAC address (anygw) is shared among both nodes. The CPU of the
switch stores a table with the relation MAC<->PORT. Then the state of
this table is changing constantly depending if the packet comes from one
node or from the other (depending on the Ethernet port). If this change
was done immediately, this behavior would not be a problem. But it seems
that most of these switch CPUs have a Time window (something close to
20s) where a new change for the same MAC cannot be done. Then it is
actually a problem and the consequences are very unpredictable.

In any case, disabling anygw from the ethernet port does not seem have
important implications for the whole network architecture. Usually when
a client is connected through cable it does not need to roam. The only
corner case I can imagine is when using stupid APs (not libremesh
bridged AP-Ethernet device) connected to a Ethernet port. In this case
the roaming would not perfectly work among these stupid APs if connected
to different libremesh nodes. But IMO we might accept this issue.

I hope it is more clear now.

On 04/04/17 17:23, Nicolás Pace wrote:
> @nicoechaniz could you give @G10h4ck context for this issue?
> 

-- 
./p4u

p4u:
I've recently experienced the problem described above in a small setup of LibreMesh. An option would be to add a new config boolean option "anygw_only_wifi" to enable or disable AnyGW on the Ethernet ports. So at the end user (or community administrator) can choose. @G10h4ck, do you see it feasible?

G10h4ck:
Still this bug miss the context as the explanation you provided is about another bug, adding an option is not a problem, thought I still miss the context where this happens (if it is a setup we support) and It is not clear to me how to disable anygw on specific interfaces as it is built on top of br-lan

p4u:
[...]

PR #145, opened 2017-06-09
Add anygw_mac_enable option on lime config

p4u:
Adds the possibility to disable anyMAC from anyGW, so the nodes are not
sharing the same MAC address anymore. This should fix unexpected
behaviors on some routers (those with internal switch). By default
anyMAC will be disabled, it might be enabled for a more agressive
roaming of the network, which might be useful on specific deployments.

Signed-off-by: Pau Escrich <p4u@dabax.net>

### DIFF
--- a/packages/lime-proto-anygw/src/anygw.lua
+++ b/packages/lime-proto-anygw/src/anygw.lua
@@ -9,7 +9,11 @@ anygw = {}
 anygw.configured = false
 
 function anygw.anygw_mac()
-	local anygw_mac = config.get("network", "anygw_mac") or "aa:aa:aa:%N1:%N2:aa"
+	local anygw_mac = utils.applyMacTemplate16("aa:%N1:%N2:%M4:%M5:%M6",
+	                                          network.primary_mac())
+	if config.get_bool("network","anygw_mac_enable",false) then
+		anygw_mac = config.get("network", "anygw_mac","aa:%N1:%N2:aa:aa:aa")
+	end
 	return utils.applyNetTemplate16(anygw_mac)
 end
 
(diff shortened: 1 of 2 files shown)


nicoechaniz:
I believe this calls for a deeper debate, specially if we would be changing the default behavior.

nicopace:
Related with #108

p4u:
If I'm not wrong, the only thing that might be broken is the roaming. In the theory it will still work, but in some scenarios not with the optimal behavior (until the ARP table updates).

In the other side it will fix some already experienced problems which are very annoying and we have they exist. 

So IMO is worth it to enable by default this option for the current develop branch. Once tested, if it works fine, it might come in the upcoming release.

nicopace:
What about the thisnode.info access that we talked about in Athens?

p4u:
> What about the thisnode.info access that we talked about in Athens?

If two or more routers are connected through Ethernet cable (using the Internal switch), and a client is connected to the same Ethernet switch. You can actually never be 100% sure that _thisnode.info_ or _anygw_ domains will resolve to the actual node where you are connected. Even if anyMAC is enabled it will depend on the behavior of the switch. The theory says that the closest will reply first to the MAC request, but it is up to the switch implementation and external input we do not control. 

In the other side, in the same scenario but the client connected to the WiFi AP (which is bridged to the Ethernet), I'm not sure which will be the behavior. But the theory says that as the packets are going first to the AP, then to the kernel and to the bridge, the node where the client is directly connected will reply first. However we must test and see to be sure.

p4u:
After having some discussions about this, I don't think we should make this change but instead we should find why the TCP connextions are broken (TCP reset) when the internal switch changes from one port to another.

PR #153, opened 2017-06-14
Hotfix/bmx6 mtu anticlimax

G10h4ck:
BMX6 has a bug in it's tunnel setup that if the MTU is not constant between all the links then an mtu anticlimax starts so the MTU goes down till no traffic goes through the tunnels.
This configures the routers so this problem doesn't happen any more.


packages/lime-proto-bmx6/src/bmx6.lua
```diff
@@ -102,9 +102,27 @@ function bmx6.configure(args)
...
+		then hasBatadv = true; end
+	end
+
+	if(hasLan) then
+		uci:set("bmx6", "lm_net_br_lan", "dev")
+		uci:set("bmx6", "lm_net_br_lan", "dev", "br-lan")
+
+		if(hasBatadv and not bxmOverBatdv) then
+			fs.mkdir("/etc/firewall.lime.d")
+			fs.writefile("/etc/firewall.lime.d/20-bmx-not-over-bat0-ebtables",
```
dangowrt: shouldn't there be an equivalent rule for the OpenWrt stock fw3 firewall as well?

p4u: All firewall.lime.d rules are executed by the OpenWRT stock Firewall if it is installed. So this rule would be executed too if fw3 installed.

https://github.com/libremesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/firewall.lua#L32

pedro-nonfree:
Is this issue related with the issue of mailing list **[lime-dev] one-way ip6ip6 tunnels done right** ? Axel said that could not find a solution to encapsulate ipv4 on ipv6 tunnel following this "new approach".

This hotfix: *This configures the routers so this problem doesn't happen any more.* If is easy to explain, can you provide more information please?

I read the commit messages and the source code. For me it looks like a "tidy up" of bmx6.lua code (improving integration of bmx6 with libremesh) but I don't see the *MTU solution*

if ( reaction == "wtf" ) then print "sorry for the noise" end

nicopace:
ajajaj @pedro-nonfree thanks for asking.
It takes a little bit to explain but yes, at least on libremesh, this solves the MTU issue.
now you have to migrate to LibreMesh ;)

packages/lime-proto-bmx6/src/bmx6.lua
p4u: Opened a new issue to discuss this ebtables issue https://github.com/libremesh/lime-packages/issues/169

PR #160, opened 2017-06-15
firewall: add lm_* interfaces to 'lan' zone

dangowrt:
To get the stock firewall working nicely, add a all lm_* interfaces
to the 'lan' zone. And we really shouldn't change the defaults for
the wan zone (!), why ever this was done, it really shouldn't.
If the wan interface isn't used as a wan interface, it shouldn't be
called wan -- calling it wan and opening the firewall for all
incoming packages while keeping SSH with no-password-auth open is
real madness and will get people pwned.

### DIFF
--- a/packages/lime-system/files/etc/firewall.lime
+++ b/packages/lime-system/files/etc/firewall.lime
@@ -2,6 +2,9 @@
 # they will be executed during each firewall (re-)start
 # or, if firewall package is not installed, just during boot.
 # They are interpreted as shell script.
+
+[ $(ls -1 /etc/firewall.lime.d/* 2>/dev/null | wc -l) = 0 ] && return 0
+
 for hook in /etc/firewall.lime.d/* ; do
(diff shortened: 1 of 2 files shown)


dangowrt:
I made two more changes since the original submission:
Now only `lm_.*_if` interfaces are added to the `lan` zone, as otherwise `eth0` would end up in both, `lan` and `wan`, which doesn't make sense. I also made sure that the firewall.lime no longer returns 1 if there are now hooks to be executed (probably there is a more elegant way to do this though, I'm open for suggestions).

dangowrt:
Tested on several devices and works fine. Removed 'RFC' prefix from the PR 😁

dangowrt:
ping @p4u

ilario:
Would this solve the passwordless SSH problem discussed in lime-dev [long time ago](https://lists.libremesh.org/pipermail/lime-dev/2016-December/000821.html)?

PR #163, opened 2017-06-20
Removing overlap between lime-docs and lime-docs-minimal

ilario:
Merging of pull request #140 introduced the following error when compiling using LEDE buildroot (using lime-sdk such errors are ignored maybe?):

```
 * check_data_file_clashes: Package lime-docs wants to install file /home/ilario/software/lede3/build_dir/target-powerpc_8540_musl/root-mpc85xx/www/docs/config.txt
	But that file is already provided by package  * lime-docs-minimal
 * opkg_install_cmd: Cannot install package lime-docs.
```

So this fixes removing `config.txt` from lime-docs and introducing it via lime-docs-minimal as a dependency of lime-docs.

### DIFF
--- a/packages/lime-docs/Makefile
+++ b/packages/lime-docs/Makefile
@@ -23,6 +23,7 @@ include $(INCLUDE_DIR)/package.mk
 define Package/$(PKG_NAME)
   CATEGORY:=LiMe
   TITLE:=LibreMesh documentation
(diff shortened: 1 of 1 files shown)


nicopace:
@p4u can you look at this?

p4u:
It looks fine. IMO can be merged.

PR #168, opened 2017-06-30
Do not depend on iptables if fw3 installed

p4u:
lime-proto-wan and lime-proto-bmx[6/7] only depend on iptables if
lede/openwrt firewall is not selected.

Signed-off-by: Pau Escrich <p4u@dabax.net>

### DIFF
--- a/packages/lime-proto-bmx6/Makefile
+++ b/packages/lime-proto-bmx6/Makefile
@@ -22,7 +22,7 @@ define Package/$(PKG_NAME)
   CATEGORY:=LiMe
   MAINTAINER:=Gioacchino Mazzurco <gio@eigenlab.org>
   URL:=http://libremesh.org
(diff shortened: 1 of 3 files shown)


p4u:
Still missing set FW3 to allow Forward on WAN

p4u:
I think this PR can be merged, even if the FW3+mesh-on-wan is not yet fixed. If iptables is selected on menuconfig it will be the same as until now. Please review and accept or comment.

nicopace:
I would add that could be merged if the default includes iptables till we deal with the FW3+mesh-on-wan.

p4u:
There is not default on the libremesh packages. It depends on the cooking process, so if the user uses lime-sdk (aka cooker) by default it installs iptables so FW3 is not needed.

The only thing that introduces this PR is the possibility of use FW3 without iptables (which until now was not possible), but ONLY if the user decides to do so.

Issue #47, opened 2016-05-09
WiFi client mode


p4u:
In addition to ap, adhoc and 802.11s, it would be useful to have also the client mode.

Example of configuration for /etc/config/lime

```
config wifi radio0
    list modes 'client'
    list protocols 'wan'
    option channel_2ghz 'auto'
    option client_ssid 'WLAN1'
    option client_key '9781saf84q09asf1'
    option encryption 'psk'
```

ilario:
Sounds very useful! 
But you mention that we have an "ap" mode... Do you mean the ap interface for non-lime clients connection? On that interface there is no routing protocol running, right? 
If it's so, we would need to add also an ap mode with routing protocols.
It's good to test alternatives to ad-hoc, its implementation in current hardware and drivers could be much worse than AP/STA implementation (for example I could find absolutely **no** mention to ad-hoc mode in 802.11n certification related documentation, for example [here](http://unina.stidue.net/Protocolli%20per%20Reti%20Mobili/Materiale/resource_2/802.11n_White_Paper.pdf)).

p4u:
In one hand we have the "mode" which might be: ap, adhoc or 802.11s

In the other we have the "protocol" which might be: wan, lan, bmx6...

So we are talking about different layers

ilario:
Ok.
So, on current LiMe, configuring something like

```
config wifi radio0
	list modes 'ap'
	list protocols 'bmx6:13'
	option channel_2ghz '1'
	option ap_ssid 'libre-mesh'
```
does work? Good to know!!

Regarding 802.11s, is it already supported in LiMe?

G10h4ck:
It is easy to implement but you should do two config sections because wifi and net are two separated layers

config wifi radioX
    list modes 'sta'
    ....

config net wlan_client
    option linux_name wlanX_sta
    list protocols wan

Cheers!

p4u:
Partially added on commit 7510ba643d766986607fc349825d311df8f16e7a

ilario:
What is missing to this feature to be considered completed?

nicopace:
@p4u could you check if this is already implemented, or state what is missing?
Thanks! :)

ilario:
I and @G10h4ck think that the hardcoded WAN network for client mode should be removed, even if this requires the user to write a second configuration section.
Client mode could be used also for LAN (building backbone) and not only for WAN (reaching internet).
[...]

PR #173, opened 2017-07-04
Update lime-example and move it into lime-docs

ilario:


ilario:
Do you think I should move adhoc options in the main wifi configuration of lime-example or leave them in the specific one? Related discussion on #174

ilario:
There are conflicts caused by merge of #176, I will solve them in a while.

packages/lime-debug/Makefile
```diff
@@ -23,8 +23,9 @@ define Package/$(PKG_NAME)
 	MAINTAINER:=Gioacchino Mazzurco <gio@eigenlab.org>
 	URL:=http://libremesh.org
 	DEPENDS:=+PACKAGE_lime-proto-batadv:batctl \
-		+busybox +ethtool +iwinfo +iw +mtr +ip +iputils-ping6 +iputils-ping \
-		+sprunge +safe-reboot +netperf +pv +tcpdump-mini +bwm-ng
+       		+busybox +ethtool +iwinfo +iw +mtr +ip +iputils-ping6 +iputils-ping \
+		+sprunge +safe-reboot +netperf +pv +tcpdump-mini \
+		+bwm-ng
```
nicopace: This change shouldn't be here, as it is just aesthetic

packages/lime-proto-bmx6/Makefile
```diff
@@ -7,7 +7,7 @@
 include $(TOPDIR)/rules.mk
 
 LIME_BUILDDATE:=$(shell date +%Y%m%d_%H%M)
-LIME_CODENAME:=bigbang
+LIME_CODENAME:=DaybootRelay
```
nicopace: this change has nothing to do with the topic of documentation, neither changes related with this package

ilario:
> There are conflicts caused by merge of #176, I will solve them in a while.

This. @nicopace ignore this PR until when I fix the things please.

packages/lime-proto-bmx6/Makefile
ilario: @nicopace ignore this PR until when I fix the things please.

packages/lime-debug/Makefile
ilario: This doesn't come from me, it's from the revert of a Pau's commit 3e3c0f8 "Move lime-example", if you think it's better, I can avoid reverting it and build my PR on top of it.

nicopace:
@ilario this is making builds break. can you check and correct ?

ilario:
Ok, now it's clean.
lime-example still misses the specification of network for client mode, see discussion on #47 but this can be part of a successive PR.

PR #174, opened 2017-07-04
Remove adhoc specific options from lime-defaults

ilario:
We moved from adhoc to IEEE 802.11s (mesh), so having options for adhoc in lime-defaults is just confusing.


nicopace:
How this will work with the new changes @G10h4ck added to the lime apply?

ilario:
Link to these changes pls?

G10h4ck:
Those options should not be removed

nicopace:
@G10h4ck can you share why? (so collective knowledge works?)
Cause @ilario thought taht it should be removed

G10h4ck:
As you can see in the whole file we saves defaults for all modules even for those not instelled, nor used by default, that is because this way if one just enable a module it works like a charm without having to edit lot of configurations

p4u:
Agree with @G10h4ck

nicopace:
Thanks @G10h4ck !

ilario:
Ok, thanks :)

PR #176, opened 2017-07-08
Add lime-docs-it

ilario:
This change requires the documentation in lime-web to start with a prefix indicating the language e.g `en_config.txt`, see discussion on libremesh/lime-web#28


nicopace:
Related with https://github.com/libremesh/lime-web/pull/28

PR #179, opened 2017-07-11
Add libremesh opkg repositories

p4u:
On first boot create opkg feeds repostory file and libremesh official
signing key. So now opkg can be used for installing libremesh packages.

Example: opkg update && opkg install lime-full


p4u:
I've perform the next test:

1. Install lime-zero on a wdr4300
2. Run: opkg install lime-full
3. reboot

It worked perfectlly.

G10h4ck:
seems nice the only thing i believe we should someway clarify, is from line 19 to 22

PR #180, opened 2017-07-11
fix #170 

### LINKED ISSUE #170: firmware upgrade via webui simple mode: no keep-no-settings
trying firmware upgrade via "simple mode" gives me no chance to say "keep no settings".
its mentioned on the webui, but there is no checkbox!

unchecking the "keep settings" checkbox is very important, because libremesh is an autoconfig mesh firmware.. (so keeping settings is not needed for most of the devices. To uncheck that option is mandatory for testing)

no-keep-no-settings

aparcar:
remove misleading "keep settings" in sysupdgrade view

### DIFF
--- a/packages/lime-webui/luasrc/view/flashops.htm
+++ b/packages/lime-webui/luasrc/view/flashops.htm
@@ -15,7 +15,7 @@
 		<% if upgrade_avail then %>
 			<form method="post" action="<%=url('lime/flashops/sysupgrade')%>" enctype="multipart/form-data">
 				<input type="hidden" name="token" value="<%=token%>" />
-				<div class="cbi-section-descr"><%:Upload a LibreMesh image here to replace the running firmware.%></div>
+				<div class="cbi-section-descr"><%:Upload a LibreMesh sysupgrade image here to replace the running firmware.%></div>
 				<div class="cbi-section-node">
 					<div class="cbi-value cbi-value-last<% if image_invalid then %> cbi-value-error<% end %>">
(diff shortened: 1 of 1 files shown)


ilario:
+1
It would be useful to specify that the sysupgrade (and not the factory) image should be used here.

aparcar:
@ilario does that make a difference for the lime-sysupgrade?

ilario:
Hm... I suppose so... I always thought that there was a real reason for the existence of "-sysupgrade.bin" images but I never verified this!

nicopace:
I flashed non-sysupgrade images with openwrt and they worked just fine... we discussed this with Gui in the past... you should try to clarify this once and for all.

ilario:
So the existence of "-sysupgrade.bin" images when "-factory.bin" is also present is completely useless waste of disk space, compilation time and confusing for users...... WTF
Ok, @aparcar sorry for asking that change, it was not needed (but helps to reduce confusion for end user).

nicopace:
@ilario please test!

aparcar:
@ilario factory is still needed for initial flash from stock. No waste of space unless you want to first flash pure LEDE (factory image) and LiMe sysupgrade afterwards. 

I'd stick to the current formulation and keep the factory images on the server as well.

nicopace:
The question @aparcar is if sysupgrade is needed :)

ilario:
Regarding the PR: for me is ok for merge.

ilario:
This is just a change of text in a webpage, no reason for delaying, I merge.

PR #183, opened 2017-07-12
Move lime-example to lime-docs and use it for lime-docs-minimal fix #182

ilario:
Remove the fragile part of lime-docs using $(wildcard ...) function which apparently got broken in d23122a and caused #182
From now on lime-docs-minimal will not install anything from lime-web, avoiding overlap with lime-docs-*, instead will install lime-example to /www/docs

Issue #182, opened 2017-07-12
Error compiling lime-docs in fresh install lime-sdk 17.01.2


gmarcos87:
```
lime-sdk git:(master) ✗ J=4 V=s ./cooker -b x86/generic --package=lime-docs
-> Building 17.01.2/x86/generic/sdk
Updating feed 'base' from '/home/okupa/libremesh/lime-sdk/feeds/base' ...
Create index file './feeds/base.index' 
Updating feed 'packages' from '/home/okupa/libremesh/lime-sdk/feeds/packages' ...
Create index file './feeds/packages.index' 
Updating feed 'luci' from '/home/okupa/libremesh/lime-sdk/feeds/luci' ...
Create index file './feeds/luci.index' 
Updating feed 'routing' from '/home/okupa/libremesh/lime-sdk/feeds/routing' ...
Create index file './feeds/routing.index' 
Updating feed 'telephony' from '/home/okupa/libremesh/lime-sdk/feeds/telephony' ...
Create index file './feeds/telephony.index' 
Updating feed 'libremesh' from '/home/okupa/libremesh/lime-sdk/feeds/libremesh' ...
Create index file './feeds/libremesh.index' 
Updating feed 'libremap' from '/home/okupa/libremesh/lime-sdk/feeds/libremap' ...
Create index file './feeds/libremap.index' 
Updating feed 'limeui' from '/home/okupa/libremesh/lime-sdk/feeds/limeui' ...
Create index file './feeds/limeui.index' 
Installing all packages from feed libremesh.
Installing all packages from feed libremap.
Installing all packages from feed limeui.
-> Building package lime-docs on sdk 17.01.2/x86/generic/sdk
make: se entra en el directorio '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
make[1]: Entering directory '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
make[2]: Entering directory '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
make[2]: Leaving directory '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
make[2]: Entering directory '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
make[2]: Leaving directory '/home/okupa/libremesh/lime-sdk/17.01.2/x86/generic/sdk'
[...]

Issue #178, opened 2017-07-10
Remove lime-full and lime-basic packages, just use lime-sdk flavors


ilario:
Widening the discussion started on #175:
now that we have collections in the new lime-sdk called [flavors](/libremesh/lime-sdk/blob/master/flavors.conf) why do we still need lime-basic and lime-full collections?
Wouldn't be more direct to just use one of the two ways for selecting and deselecting a list of packages?

p4u:
Having collections of packages is useful, not only to have everything
more organized (if lime-basic does not exist, the flavor must contain a
long list of packages) but also when using opkg to from a plain LEDE
router to install libremesh (opkg install lime-full).

ilario:
Ok, like in #179? Nice!
Anyway I would disentangle lime-sdk from lime-packages collections.
I mean: let's keep the collections but let's avoid their use in lime-sdk.
This prevents situations like #175

ilario:
Are there more opinions before closing this?
@altergui @G10h4ck @nicopace @nicoechaniz

nicopace:
i think it is clear to close it

p4u:
I don't see a strong point for removing the collections from lime-sdk. So IMO it is better if we use the collections. The users of lime-sdk can define their own flavors without using any collection if they have a point for doing so.

ilario:
lime-basic has been removed in #674

PR #184, opened 2017-07-12
Move lime-example to lime-docs, separate lime-docs and lime-docs-minimal

ilario:
This is an alternative to #183 solution to #182 problem.
Here lime-docs and lime-docs-minimal become incompatible, they can't be installed at the same time.
This should break thing as the first is selected by lime-full and the second i selected by lime-basic which is selected by lime-full.
Unexpectedly lime-sdk doesn't complain.
Still this solution is not optimal, so I would suggest either to go for #183 or to avoid using lime-full and lime-basic (related to #178).


ilario:
Another possibiltiy is that someone fix the 
```
$(INSTALL_DATA) $(filter-out $(PKG_BUILD_DIR)/en_config.txt,$(wildcard $(PKG_BUILD_DIR)/en_*.txt)) $(1)/www/docs/
```
line in lime-docs/Makefile which seems to be broken (refer to #182).

packages/lime-docs/Makefile
```diff
@@ -63,13 +64,15 @@ endef
 
 define Package/$(PKG_NAME)/install
 	$(INSTALL_DIR) $(1)/www/docs/
-	$(INSTALL_DATA) $(filter-out $(PKG_BUILD_DIR)/en_config.txt,$(wildcard $(PKG_BUILD_DIR)/en_*.txt)) $(1)/www/docs/
+	$(INSTALL_DATA) $(PKG_BUILD_DIR)/en_*.txt $(1)/www/docs/
+	$(INSTALL_DATA) ./files/lime-example $(1)/www/docs/
```
gmarcos87: ./files/lime-example not exist on lime-docs folder

ilario: Wow I failed totally, I moved the lime-example file but I moved it to the wrong place!!!!

ilario: Fix in #185

ilario:
I moved the lime-example file to the wrong dir, fixed in #185.

PR #185, opened 2017-07-13
Fixup to error in #184

ilario:
In #184 I moved lime example from `lime-debug/files/etc/config/lime-example` to `packages/lime-debug/files/lime-example` which is clearly wrong. Here I move the same file to `packages/lime-docs/files/lime-example`.


nicopace:
At least from my undestanding this is ok.
We should take note (or leave comments) on external dependencies like lime-docs from now on!

ilario:
Mmh, I don't know where to write this. 
But I agree that is easy to break things in situations like this where there are dependencies from lime-packages to lime-web repositories...

PR #186, opened 2017-07-14
Implementing Contiuous Integration

nicopace:
Based on https://github.com/probonopd/announce/blob/master/.travis.yml

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,32 @@
+sudo: false
+dist: trusty
+addons:
+  apt:
+    sources:
+    - sourceline: 'ppa:ubuntu-toolchain-r/test'
+    packages:
+    - build-essential
+    - subversion
+    - libncurses5-dev
+    - zlib1g-dev
+    - gawk
+    - gcc-multilib
+    - flex
+    - git-core
+    - gettext
+    - libssl-dev
+    - libstdc++6-4.7-dev
+notifications:
+  email: false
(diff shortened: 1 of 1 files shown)


nicopace:
Blocked because of this:
* https://github.com/libremesh/lime-sdk/issues/13
* https://github.com/libremesh/lime-sdk/issues/12

ilario:
libremesh/lime-sdk#13
has been solved by libremesh/lime-sdk#15

libremesh/lime-sdk#12
has a workaround forcing it to exit code 0 (via `|| true`) in .travis.yml file but was also forced to exit code 0 due to libremesh/lime-sdk#15

So what fails now is the cooking time that using Travis for free is limited to 50 minutes **for each thread** so maybe it's enough to split the compilation process in more than one command.
For example we can try adding `./cooker -b ar71xx/generic` before the actual cooking command?
Otherwise how could we split the cooking command in smaller tasks?

ilario:
Travis fails when cooker fails, Travis works when cooker works. The running time is below 30 min (limited to 50 min from travis-ci.org). @nicopace what's missing? Should we merge all the commits in a few or one commit?

ilario:
Example of a failing job: https://travis-ci.org/libremesh/lime-packages/jobs/259410507
Example of a working job: https://travis-ci.org/libremesh/lime-packages/jobs/259368260

PR #187, opened 2017-07-17
Avoid lime-basic to select lime-docs-minimal when lime-docs is selected by lime-full

ilario:
As explained in #184, its merge introduced a dependency problem which seemed innocuous (?) but appears to break continuous integration (#186).


ilario:
On my PC, using lime-sdk, compiles ok the lime_default profile, I merge

PR #188, opened 2017-07-17
Remove @ from dependency in lime-docs

ilario:
@p4u suggested to remove the @ from the dependency in lime-docs, which got introduced in #184 
Reference documentation [here](https://wiki.openwrt.org/doc/devel/dependencies).


p4u:
Is is somehow tested?

ilario:
Not yet. Can yo list exactly what I should verify?

ilario:
I tested a bit, the changes result in an invalid syntax.

ilario:
But I don't know why. I realized that "+!PACKAGE_..." is a valid value for DEPENDS, as is used also elsewhere. Does anyone know why this commit is producing errors in menuconfig when compiling with LEDE buildroot?

PR #198, opened 2017-08-02
travis-ci badge

nicopace:


ilario:
Are you sure that the link to `travis-ci.org/libremesh/lime-packages` won't result in a link to `https://github.com/travis-ci.org/libremesh/lime-packages` ? In doubt I would add `https://` in front of it.

nicopace:
That is why I loooooove peer reviews :)
Thanks for catching that @ilario, alredy corrected and tested...  it works now !

PR #200, opened 2017-08-02
Move lime-hwd-openwrt-wan from lime-basic* to lime-full

### LINKED ISSUE #175: remove lime-hwd-openwrt-wan as dependency of lime-basic
creating a new flavor with `-lime-hwd-openwrt-wan`does not work anymore. 
I want to create an image for routers with low power not needing any gateway capabilities.

ilario:
This fixes #175.
Needs to be merged at the same time of libremesh/lime-sdk#20

### DIFF
--- a/packages/lime-basic-uing/Makefile
+++ b/packages/lime-basic-uing/Makefile
@@ -27,7 +27,7 @@ define Package/$(PKG_NAME)
 	DEPENDS:= +lime-system +lime-proto-bmx6 +lime-proto-batadv \
 	         +lime-proto-anygw +lime-proto-wan +dnsmasq-lease-share \
 	         +dnsmasq-distributed-hosts +lime-webui-ng \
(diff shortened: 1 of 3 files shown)


p4u:
I don't see the point for doing this. This package has almost no impact on the flash size and it is useful for autodetecting the WAN ports of the routers (if not enabled, everything is configured as LAN).

ilario:
@p4u I would keep the discussion centralized on #175.
Anyway personally I don't have an opinion on this... But maybe the same approach can be applied on many other packages (e.g. lime-hwd-ground-routing) allowing the users to make even more reduced images.

nicopace:
So... @ilario do you consider this still valid? in that sense @p4u can we move about this?

ilario:
@nicopace I'm in favour of moving most of the not strictly needed packages from lime-basic-* packages to the lime-sdk flavors, but from the discussion on #175 seems that @p4u prefers to keep things unchanged. 
Let's keep the discussion on #175.

ilario:
Closing as issue #175 is wontfix.

Issue #201, opened 2017-08-03
Set a better default distance parameter on the WiFi conifguration


p4u:
Currently distance is set to 10000 (10Km). Reducing it to 1000 (1Km) would increase the network performance for most of the libremesh deployments. For those which are using long distance links, the distance should be set manually to ensure the best performance.

In addition to reduce the default distance, there should be a way to easy configure this parameter from the libremesh webui.

ilario:
What about a different distance setting for 2.4 GHz (usually clients) and for 5 GHz (usually mesh nodes)?
For example 200 m for 2.4 GHz and 5 km for 5 GHz. 

This is how it would be set in `/etc/config/lime-defaults`:
```
config lime wifi
	option distance_2ghz '200'
	option distance_5ghz '5000'
```

p4u:
I don't think this is possible on the WiFI configuration of LEDE. Actually, if I don't remember wrong, this parameter affects mainly to the 5GHz band.

ilario:
It should be possible, it has been implemented in LibreMesh [here](/libremesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/wireless.lua#L106), and in LEDE the configuration of different radios (2.4 GHz and 5 GHz) is separated, so it's the distance setting in LEDE for the two radios.
Even if nowadays maybe drivers for 2.4 GHz are not affected by this parameter, this could change.

ilario:
Closing as #202 has been merged.

p4u:
You are right, it can be set independently for the radios. It might be implemented as you proposed.

nicoechaniz:
When you are using dual band devices such as tl-wdr3500 the distance on each radio is the same. This is the most common scenario in the networks in our region. So a different value for 5Ghz and 2.4Ghz would not be our default.

ilario:
@nicoechaniz so in your scenario both 2.4 and 5 GHz are used for node to node connections? Isn't 2.4 GHz radio involved more heavily in node to client connections?

digitigrafo:
On 18/05/23 19:00, Ilario Gelmetti wrote:
>
> Let's reopen this discussion.
> @digitigrafo <https://github.com/digitigrafo> reported that plenty of 
> communities had problem with 1000 m as maximum default distance.
> I propose to set it to 5 km.
>
I reported it in 2019.. when I was in brasil and I find 2 communities 
that abandoned to use Lime because can't update from lime17 to lime19..

[...]

PR #202, opened 2017-08-03
Set default WiFi distane to 1000. Add webui option to set distance

p4u:
Related issue #201


p4u:
Tested on a local router.

PR #204, opened 2017-08-04
Adds rsync copying artifacts to ci.libremesh.org

nicopace:
After the CI finishes, it copies the outcome of the build to a branch directory in the ci.libremsh.org host.


ilario:
Could you create a temporary known_hosts file from the Travis script including correct repo.libremesh.org SSH host key? 
Then you can use it in "-o UserKnownHostsFile" :)

PR #205, opened 2017-08-05
Changes PR name for ID in artifacts copy

nicopace:
If two branches have the same name, they will collide.
In order to prevent that, we will use the PR ID as the base for directories.


nicopace:
This branch has been included in #206

Issue #97, opened 2017-03-31
Bundle Lime Workarounds


nicopace:
During the development of LibreMesh we found several way to deal with upstream inconcistencies or hardware failures. These ones we called them worarounds.
Get all of them together into a package to easen its maintenaince.

G10h4ck:
Hardware specific quircs are supposed to be handled with hardware detection plugins

ilario:
Are the workarounds these ones?
https://github.com/libremesh/network-profiles/tree/master/quintanalibre.org.ar/comun/etc/uci-defaults

nicopace:
@nicoechaniz ?

nicopace:
I guess so... but let's hope others can comment

ilario:
Related discussion is going on in #137

ilario:
smonit was merged in #142, some of the workarounds got implemented in it.

nicopace:
@ilario are you sure all the workoarounds already in smonit? @G10h4ck @p4u @altergui can you confirm?

ilario:
@nicopace no I'm not

nicopace:
Then let's the guys close the issue :)

PR #211, opened 2017-08-19
Hotfix/bundle lime workaround 97

G10h4ck:
Adapt quintana workarounds to smonit, once merged this would fix bug #97


p4u:
@G10h4ck instead of adding your comments about SMONIT on the code it would be better if you add an issue on github or just send the PR for fixing it.

However, the PATH variable specification is not needed, it was added for 'just in case', to be sure that the PATH is the correct PATH. Anyway it can be removed, I don't think it makes a real difference.

PR #212, opened 2017-08-20
Sandbox/ubus luci

gmarcos87:
This pull request aims to share the work that I have been doing in the migration of luci services that uses libremesh to ubus/luci2. Still luci2 frontend architecture is not completely defined (you can see this pull request for more information https://github.com/jow-/luci-ng/pull/42). What is clear is that the backend will be via ubus.
So I'm working on keeping the functionalities currently implemented (in principle in the view without authentication) and be able to expose the bmx6, batman-adv, spectrum analysis and geographic location data by ubus.
Most of the code is reused from previous modules with modifications to remove any dependency related to luci.
It is a work in progress, there are three libraries to remove (luci-lib-json, nixio and luci.model.network).
The packages are:
* [ubus-luci-openairview](https://github.com/gmarcos87/lime-packages/tree/sandbox/ubus-luci/packages/ubus-luci-openairview)
* [ubus-luci-bmx6](https://github.com/gmarcos87/lime-packages/tree/sandbox/ubus-luci/packages/ubus-luci-bmx6)
* [ubus-luci-batman-adv](https://github.com/gmarcos87/lime-packages/tree/sandbox/ubus-luci/packages/ubus-luci-batman-adv)
* [ubus-luci-location](https://github.com/gmarcos87/lime-packages/tree/sandbox/ubus-luci/packages/ubus-luci-location)

Each one has a readme explaining the methods and with sample answers.


p4u:
Congratulations for such a great job!

IMO this packages might be included in the future to some official LEDE feed. So I propose to create a new (empty) git repository named something like "lime-ubus" which will be imported by libremesh  but it might be also imported by LEDE or any other project.

Also there are packages like ubus-bmx6 where the code has been imported from luci-app-bmx6, we should merge both packages. So luci-app-bmx6 might use ubus-bmx6 and it might go to the official repository openwrt-routing (where luci-app-bmx6 is currently hosted).

So, this is a very good starting point but IMO some changes should be made to have better compatibility. To this end, having these packages (which are not yet final packages) in a separeted repository will make things easier and better organized.

gmarcos87:
Well, I understand what you're proposing. At this point I am trying to cover the functions (set and get) that are used in libremesh. Note that for the standard configuration there is "ubus uci", the ubus methods I am adding are oriented mainly to the GUI. To be useful in the LEDE level, I must consider all the needs of each module.
If you want to create the repository and give me the permissions, I will migrate this pull request with pleasure.

I forget to clarify that removing all dependencies with "luci" will mean a temporary increase in the final build, since there will be duplicate code if they are installed together.

p4u:
https://github.com/libremesh/lime-packages-ui

ilario:
Moved to libremesh/lime-packages-ui#1

PR #214, opened 2017-08-21
Fix travis fails on forks pull request

gmarcos87:
Pull requests sent from forked repositories do not have access to encrypted variables or data.
As you can not access the variables you can not decrypt the key and perform the tasks in after_success.sh
This change verifies that the action comes from the libremesh/lime-packages repository and not from a fork.


nicopace:
@gmarcos87 it seems it is not working in the case of being a branch belonging to the libremesh repository.
If you want to check it, great... if not, i'll check it asap.

gmarcos87:
Another alternative is to generate an environment variable in travis and check if it exists in the script. That is a simpler solution but you have to configure it from the Travis panel. However if you work from an existing branch in lime-packages travis build by commit, you do not need to run it again in the pull request.

aparcar:
currently merging PRs is blocked as long as the password entry isn't fixed 😿

gmarcos87:
I moved the verification to the "after_success" script. If the branch is from a fork and you do not have access to the environment variables it does not fail. I also added a timeout extension, since it used to fail for that cause too.

I think that after doing a merge of this branch it will be necessary to update pending pull requests to take the new configuration of travis. From github it can be done very easily.

PR #215, opened 2017-08-21
luci-webview: add attendedsysupgrade tab if exists

aparcar:
the attended sysupgrade shoudl simplify sysupgrades. If the app is
installed an entry should appear in the simple admin config.

@p4u

Signed-off-by: Paul Spooren <paul@spooren.de>


aparcar:
thanks

PR #216, opened 2017-08-22
Document that specific options can't go in /etc/config/lime-defaults

ilario:
Interface specific options does not work when included in /etc/config/lime-defaults, refer to @p4u's answer in lime-dev mailing list: https://lists.libremesh.org/pipermail/lime-dev/2017-August/000976.html


packages/lime-docs/files/lime-example
```diff
@@ -1,5 +1,6 @@
 # The /etc/config/lime-defaults file contains the default configuration.
 # To configure LibreMesh insert options in /etc/config/lime file, these will override the default ones.
+# Interface specific options have to be included in /etc/config/lime, if in /etc/config/lime-defaults they'll be ignored.
```
G10h4ck: It's not that they are ignored is that they cause unpredictable behaviour

ilario:
@G10h4ck I corrected the think you pointed out, can you unlock the review?

Issue #210, opened 2017-08-12
travis after_success.sh asks for password


aparcar:
My current [PR ci build](https://travis-ci.org/libremesh/lime-packages/builds/263925135?utm_source=github_status&utm_medium=notification) seems never to finish as `after_success.sh` asks for a password:
```
iv undefined
iv undefined
Warning: Permanently added the ECDSA host key for IP address '138.255.88.52' to the list of known hosts.
Enter passphrase for key '/home/travis/build/libremesh/lime-packages/.ci/travisci_rsa': 
```

gmarcos87:
This error occurs when the encrypted key is created with a password. @nicopace in this commit you created the key (5729b781fa3a0f9019cedcd733ff87ca390842de), did you use password?

nicopace:
No, and this worked for the other commits... I will try later to see what is happening

El 13 de agosto de 2017 1:16:35 AM GMT+02:00, Marcos Gutierrez <notifications@github.com> escribió:
>This error occurs when the encrypted key is created with a password.
>@nicopace in this commit you created the key
>(5729b781fa3a0f9019cedcd733ff87ca390842de), did you use password?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/libremesh/lime-packages/issues/210#issuecomment-322011704

-- 
Enviado desde mi dispositivo Android con K-9 Mail. Por favor, disculpa mi brevedad.

nicopace:
For sure it is related wirh the fact that the branch is not from the libremesh repository...
I will modify the script to be able to finish the test without doing the privileged tasks like uploading the firmware or commenting the image.
Regards

El 13 de agosto de 2017 12:45:20 AM GMT+02:00, Paul Spooren <notifications@github.com> escribió:
>My current [PR ci
>build](https://travis-ci.org/libremesh/lime-packages/builds/263925135?utm_source=github_status&utm_medium=notification)
>seems never to finish as `after_success.sh` asks for a password:
>```
>iv undefined
>iv undefined
>Warning: Permanently added the ECDSA host key for IP address
>'138.255.88.52' to the list of known hosts.
>Enter passphrase for key
>'/home/travis/build/libremesh/lime-packages/.ci/travisci_rsa': 
>```
>
>-- 
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/libremesh/lime-packages/issues/210

-- 
[...]

Issue #107, opened 2017-03-31
Alfred automatic master node definition


nicopace:
In a BATMAN network that uses Alfred, you need at least one Alfred's master node, many is better (but not too many).
Define an heuristic and implement a way for this to happen.

ilario:
What happens currently? Are them all master or all slaves?

nicopace:
@nicoechaniz ?

ilario:
Currently they're all master. The decision for being master or slave could be based on the fact of being a gateway or not, which is a completely arbitrary factor but works IMHO.
I'm not sure about where to place the `alfred --modeswitch` commands for going slave or master, the easy way would be to include them in `bmx6-auto-gw-mode`'s `wan-ok.d` and `wan-fail.d` directories.

nicoechaniz:
Illario, that is exactly what we are doing manually in many networks. Setting to master the nodes that are gw. It is arbitrary but it makes sense because you usually build the network taking special consideration in having good path to a gw from every node, so this decision will optimize alfred's performance too.

nicopace:
one question regarding this.Ir a network doesn't have an uplink (a disconnected network), how will it behave?Can we add that if there is no uplink everyone will get to be master, and if there is uplink, all turn into slaves but the one that has the link? (@ilario moved my comment from #220 cause it was generating noise there... sorry)

ilario:
@nicopace ok, so the proposed approach does not work...
Or has to be completed by an additional check regarding the existence of at least an alfred master or about the total number of nodes and number of alfred masters. No idea how to do this...

nicopace:
So... no matter which routing protocol we use, we can now if we have access to internet or not, and whether the node that gives internet has an alfred-master node or not.
Using those two checks, we can decide if the node needs to be master or not.

ilario:
@gmarcos87 I suppose this has been closed as alfred is going to be replaced by shared-state? Can you confirm? Anyway [alfred is still dependency of lime-basic via dnsmasq-lease-share and dnsmasq-distributed-hosts](https://github.com/libremesh/lime-packages/blob/master/packages/lime-basic/Makefile#L28-L29).

gmarcos87:
I'm sorry I thought the changes in #476 modified the dependencies. We can leave it open until that happens, which I think is a matter of a couple of days.

nicopace:
we can close now, as that has been fixed.

PR #220, opened 2017-09-06
lime-config: configure ALFRED batman interface

p4u:
If batman-adv is not enabled, ALFRED does not start cause it waits for bat0 interface (which does not exist). However if ALFRED is not running, dnsmasq distributed leases does not work. On a network with LAN enabled more than one libremesh nodes might be wired connected and lease share must work between them.

This commit solves this case.

1. If batadv enabled on configuration, set ALFRED batman interface to bat0.
2. If batadv not enabled but LAN is, set it to "none", so ALFRED still works on LAN.

### DIFF
--- a/packages/lime-proto-batadv/src/batadv.lua
+++ b/packages/lime-proto-batadv/src/batadv.lua
@@ -47,6 +47,12 @@ function batadv.configure(args)
 	uci:set("network", owrtInterfaceName, "proto", "batadv")
 	uci:set("network", owrtInterfaceName, "mesh", "bat0")
 	uci:save("network")
+
+	-- enable alfred on bat0 if installed
+	if utils.is_installed("alfred") then
+		uci:set("alfred", "alfred", "batmanif", "bat0")
+		uci:save("alfred")
+	end
 end
 
 function batadv.setup_interface(ifname, args)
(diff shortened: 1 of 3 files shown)


G10h4ck:
Me parece bien, no entiendo porque travis peta

ilario:
@G10h4ck ...no has visto el ticket #210 y el pull request #214?

p4u:
@G10h4ck can you accept de review then?
Thanks!

nicopace:
one question regarding this.
Ir a network doesn't have an uplink (a disconnected network), how will it behave?
Can we add that if there is no uplink everyone will get to be master, and if there is uplink, all turn into slaves but the one that has the link?

ilario:
@nicopace I think alfred will be master on all nodes anyway, as discussed on #107

p4u:
I don't understand what this PR has to do with ALFRED in Maser or Slave mode. Please, don't block the merge with non related issues...

p4u:
I don't understand why github now says

> p4u dismissed G10h4ck’s stale review via b9b52a7 9 minutes ago

Can someone review and accept please?

nicopace:
> I don't understand what this PR has to do with ALFRED in Maser or Slave mode. Please, don't block the merge with non related issues...

Sorry... my bad... I did a cross posting from another discussion! deleting my cumments now!

PR #230, opened 2017-09-12
Use libiwinfo-lua do detect radio mode instead of uci

G10h4ck:
This is expected to correctly detect available radio modes (abgnc) on
meshpotato too, before being merged need testing on meshpotato too

I reopen the PR because github someway got stuck believing the other PR has been merged while it is not


p4u:
can be merged

PR #235, opened 2017-09-18
Hotfix/consistent release name

altergui:
Cleanup garbage CODENAMEs


altergui:
On 19/09/17 13:40, Pau wrote:
> *@p4u* commented on this pull request.
> 
> Can you explain why the removal of LIME_CODENAME var?

it's only actually used in lime-system Makefile
in the other makefiles, is evidently a copypaste leftover, that is not
kept in sync

p4u:
I think @aparcar uses this variable for the remote sysupgrade system.

aparcar:
I think only `LIME_RELEASE` is used, merging is fine for me

PR #241, opened 2017-09-23
Pkg description reflecting that Alfred doesn't need batman-adv

ilario:
Thanks to bcfae39, alfred works also when batman-adv is missing.
Adapt *dnsmasq-distributed-hosts* and *dnsmasq-lease-share* packages descriptions to this change.


p4u:
The information provided is not true. Alfred do not share the hosts among all the mesh network but only among the local LAN connected nodes.

ilario:
@p4u can you suggest an alternative phrasing?

p4u:
Share DHCP lease file and hostnames accross nodes of the same mesh cloud

ilario:
@p4u thanks, is it ok now?

p4u:
yes, thanks.

PR #246, opened 2017-10-11
Add bmx6 hosts file to dnsmasq

p4u:
Adds a new entry to /etc/config/dhcp to tell dnsmasq look into
/var/hosts/bmx6 (for .mesh domains provided by lime-proto-bmx6 package).

It ensures that .mesh domains are resolved without depending in any other package.


G10h4ck:
seems ok to me

Issue #248, opened 2017-10-29
Replace Chef in readme with https://imagebuilder.libremesh.org/imagebuilder


ilario:
In the readme there's still the indication for the old Chef.

PR #250, opened 2017-10-29
Update README.md

vcuculo:
:raised_hands: :raised_hands: :raised_hands:
#248

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,9 +15,9 @@ This is the easiest way to first test and install LibreMesh in your router.
 
 You can download a firmware image with **generic configuration** of the last **release** at [downloads][9] subdomain.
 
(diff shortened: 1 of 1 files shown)


ilario:
I suggest a couple of corrections: 
replace LEDE Imagebuilder (2 occurrences in your commit) with online ImageBuilder, 'cause it has been developed for LibreMesh mainly, LEDE secondly :D
specify that we encourage the communities to create their profile on [network-profiles](/libremesh/network-profiles/) repo and build the images with the imagebuilder (the point is that the network profiles are in another place, your phrasing was misleading)

vcuculo:
I'm sorry @ilario , the issue was not well specified. It should be ok now. :ok_hand:

PR #251, opened 2017-10-30
Checks for $file_etc_lastgood existance before removal. Fixes #203

### LINKED ISSUE #203: Make safe-reboot aware if there is a backup or not
If there is no backup, [safe-reboot](/libremesh/lime-packages/blob/develop/packages/safe-reboot/Makefile#L23) shouldn't remove the /etc overlay

svalo:
The check is done right before deleting $dir_etc, if no file is found
then a message is printed and the script exits with status code 2

I tried to be consistent with the behaviour of this script regarding error code and message.

How can I improve this PR?


ilario:
Looks good to me, @nicopace?

PR #254, opened 2017-11-12
create qqq language code

patogit:
The qqq language code is for string documentation for translators. I haven't filled in the documentation messages yet.

I'm not sure if I've done the correct steps to create this pull request...?

### DIFF
--- a/packages/lime-webui/po/qqq/lime-webui.po
+++ b/packages/lime-webui/po/qqq/lime-webui.po
@@ -2,116 +2,116 @@ msgid ""
 msgstr "Content-Type: text/plain; charset=UTF-8"
 
 msgid "About"
-msgstr "documentation-to-do"
+msgstr ""
 
 msgid "Access Point ESSID (usually name of your network)"
-msgstr "documentation-to-do"
+msgstr ""
 
 msgid "Advanced"
-msgstr "documentation-to-do"
+msgstr ""
 
 msgid "Channel 2.4 GHz"
-msgstr "documentation-to-do"
+msgstr ""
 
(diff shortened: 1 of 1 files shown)


patogit:
Also, how do you want to connect with TranslateWiki.net? Where does it make sense for TW.net to have direct push commit access: a new "translate" branch, or the "develop" branch, or somewhere else?

ilario:
I think the approach we discussed and was most supported was to have a separate repository for the TW direct access and merge frequently to the main one, is it correct @nicopace ?

patogit:
To see examples of how other projects on GitHub connect with Translate Wiki, look at the contribution activity of the Translate Wiki GitHub user at https://github.com/translatewiki

patogit:
@Nikerabbit from TranslateWiki.net might have ideas.

Nikerabbit:
I am not in favor of separate repositories (or separate branches) because I don't want delivery of translations to depend on human actions.
I am willing to provide some flexibility if good reasons are given (protected branches, or complex builds which may get broken by translation updates).
I also don't understand why are you creating qqq file full of TODOs. Why don't you just leave those empty so that it is clear which ones are missing.

nicopace:
Hi @Nikerabbit .
This is exactly the case... we use protected branches in all repositories, so there is a pull request that needs to happen there.

nicopace:
@patogit can you change the documentation-to-do strings with empty strings as @Nikerabbit was mentioning?
Apart from that, it looks ok.

patogit:
I have removed the TODOs.

ilario:
The discussion on commid access moved to #257

PR #269, opened 2017-12-17
Enables lime-app as default package on lime-full

p4u:

### DIFF
--- a/packages/lime-full/Makefile
+++ b/packages/lime-full/Makefile
@@ -23,7 +23,7 @@ define Package/$(PKG_NAME)
   SUBMENU:=Collections
   MAINTAINER:=Pau Escrich <p4u@dabax.net>
   URL:=http://libremesh.org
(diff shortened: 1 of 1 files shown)


packages/lime-full/Makefile
```diff
@@ -23,7 +23,7 @@ define Package/$(PKG_NAME)
   SUBMENU:=Collections
   MAINTAINER:=Pau Escrich <p4u@dabax.net>
   URL:=http://libremesh.org
-  DEPENDS:=+lime-basic +lime-debug +luci-mod-admin-full +lime-map-agent +lime-docs
+  DEPENDS:=+lime-basic +lime-debug +luci-mod-admin-full +lime-map-agent +lime-docs +lime-app
```
aparcar: Shouldn't we remove luci-mod-admin-full as a dependence and add it to the full flavor?

p4u:
No, the APP does not substitute LuCi (yet?), it only allows few things
for the moment.

PR #271, opened 2017-12-19
Enable batadv protocol on ethernet by default

G10h4ck:
Due to lack of reproducibility and eventual alternative solution (tc+netem)
proposed for bla+ogm batadv deadloop on bridged ethernet interfaces.
Batman advanced is now re-enabled by default on plain ethernet interfaces.

### DIFF
--- a/packages/lime-proto-batadv/src/batadv.lua
+++ b/packages/lime-proto-batadv/src/batadv.lua
@@ -58,7 +58,6 @@ end
 function batadv.setup_interface(ifname, args)
 	if not args["specific"] then
 		if ifname:match("^wlan%d+.ap") then return end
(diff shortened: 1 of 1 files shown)


p4u:
So, as far as I understand, this bug cannot be reproduced anymore.

You talk about an eventual alternative solution, is this implemented somewhere? Does this solution implies to install more software? In that case, does it fit on 4MB devices?

altergui:
i managed to see very strange behaviour in a quick test of Gio's approach adding latency with netem. so adding latency doesnt work around this :( please don't merge this until someone else can reproduce (or deny) my observations

Am 28. Dezember 2017 16:23:39 MEZ schrieb Pau <notifications@github.com>:

PR #274, opened 2018-01-06
Introduce new option bmx6_publish_ownip

p4u:
If enabled bmx6 will publish the own node IPv4/IPv6 addresses via tunIn. By default it is disabled.

This option on some scenarios such as:

1. The users want to have an easy way to identify node IPs (bmx6 tunnels will show it)
2. A cloud (sharing the same network domain) is broken. If this option is enabled the nodes of one side will be able to reach the nodes of the other side throgh the standard ipv4 and ipv6
3. The users want to let bmx6 decide which path to take from one node to another (instead of just dropping the packet to br-lan and let bat-adv decide).

### DIFF
--- a/packages/lime-proto-bmx6/src/bmx6.lua
+++ b/packages/lime-proto-bmx6/src/bmx6.lua
@@ -32,8 +32,8 @@ function bmx6.configure(args)
 	uci:set(bmx6.f, "main", "tun6Address", ipv6:string())
 
 	-- If publish own IP enabled, configure tunIn
-	local pub_own_ip = config.get("network", "bmx6_publish_ownip")
-	if (pub_own_ip == '1' or pub_own_ip == 'true') then
+	local pub_own_ip = config.get_bool("network", "bmx6_publish_ownip", false)
+	if (pub_own_ip) then
 		uci:set(bmx6.f, "myIP4", "tunIn")
 		uci:set(bmx6.f, "myIP4", "tunIn", "myIP4")
 		uci:set(bmx6.f, "myIP4", "network", utils.split(ipv4:string(),'/')[1]..'/32')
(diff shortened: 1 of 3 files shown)


packages/lime-system/files/etc/config/lime-defaults
```diff
@@ -17,6 +17,7 @@ config lime network
 	option anygw_dhcp_limit '0'
 	option main_ipv6_address '2a00:1508:0a%N1:%N200::/64'
 	option bmx6_mtu '1500'
+	option bmx6_publish_ownip '0'
```
G10h4ck: I believe it more "self documenting" use the boolean value false instead of 0 (may mislead the user to think it's an integer value)

packages/lime-proto-bmx6/src/bmx6.lua
```diff
@@ -31,6 +31,17 @@ function bmx6.configure(args)
 	uci:set(bmx6.f, "main", "tun4Address", ipv4:string())
 	uci:set(bmx6.f, "main", "tun6Address", ipv6:string())
 
+	-- If publish own IP enabled, configure tunIn
+	local pub_own_ip = config.get("network", "bmx6_publish_ownip")
+	if (pub_own_ip == '1' or pub_own_ip == 'true') then
```
G10h4ck: config.get_bool should be used in this cases ;)

p4u:
Check new commits please.
Thanks!

G10h4ck:
Most of the code in new commits is unrelated and basically reverts a bunch of decisions and fixes that has been made in the past, to speed up a bit I'll cherry pick the commits that seems ok to me and do a new PR

p4u:
> Most of the code in new commits is unrelated and basically reverts a bunch of decisions and fixes that has been made in the past, to speed up a bit I'll cherry pick the commits that seems ok to me and do a new PR

Hey, I made this modification cause you asked for it. But for using config.get_bool() (as you suggested) a default option is needed. 

What do you propose to not break the libremesh installations after an upgrade if they are keeping an old/customized lime-defaults file?

Anyway, I don't get why you closed this PR without merging anything or giving the change to modify it. If you want to cherry-pick and send PR yourself, at least give me the chance to reply.

p4u:
Let's continue to the other PR.

PR #278, opened 2018-01-10
Feature/bmx6 publish own ip v2

G10h4ck:
Obsoletes https://github.com/libremesh/lime-packages/pull/274

### DIFF
--- a/packages/lime-proto-bmx6/src/bmx6.lua
+++ b/packages/lime-proto-bmx6/src/bmx6.lua
@@ -31,6 +31,17 @@ function bmx6.configure(args)
 	uci:set(bmx6.f, "main", "tun4Address", ipv4:string())
 	uci:set(bmx6.f, "main", "tun6Address", ipv6:string())
 
+	-- If publish own IP enabled, configure tunIn
+	local pub_own_ip = config.get("network", "bmx6_publish_ownip")
+	if (pub_own_ip == '1' or pub_own_ip == 'true') then
+		uci:set(bmx6.f, "myIP4", "tunIn")
+		uci:set(bmx6.f, "myIP4", "tunIn", "myIP4")
+		uci:set(bmx6.f, "myIP4", "network", utils.split(ipv4:string(),'/')[1]..'/32')
+		uci:set(bmx6.f, "myIP6", "tunIn")
+		uci:set(bmx6.f, "myIP6", "tunIn", "myIP6")
+		uci:set(bmx6.f, "myIP6", "network", utils.split(ipv6:string(),'/')[1]..'/128')
+	end
+
 	-- Enable bmx6 uci config plugin
 	uci:set(bmx6.f, "config", "plugin")
 	uci:set(bmx6.f, "config", "plugin", "bmx6_config.so")
--- a/packages/lime-proto-bmx6/src/bmx6.lua
+++ b/packages/lime-proto-bmx6/src/bmx6.lua
@@ -32,14 +32,13 @@ function bmx6.configure(args)
 	uci:set(bmx6.f, "main", "tun6Address", ipv6:string())
 
 	-- If publish own IP enabled, configure tunIn
-	local pub_own_ip = config.get("network", "bmx6_publish_ownip")
-	if (pub_own_ip == '1' or pub_own_ip == 'true') then
-		uci:set(bmx6.f, "myIP4", "tunIn")
-		uci:set(bmx6.f, "myIP4", "tunIn", "myIP4")
-		uci:set(bmx6.f, "myIP4", "network", utils.split(ipv4:string(),'/')[1]..'/32')
-		uci:set(bmx6.f, "myIP6", "tunIn")
-		uci:set(bmx6.f, "myIP6", "tunIn", "myIP6")
-		uci:set(bmx6.f, "myIP6", "network", utils.split(ipv6:string(),'/')[1]..'/128')
+	if config.get_bool("network", "bmx6_publish_ownip") then
+		uci:set(bmx6.f, "ownIPv4", "tunIn")
+		uci:set(bmx6.f, "ownIPv4", "tunIn", "ownIPv4")
+		uci:set(bmx6.f, "ownIPv4", "network", ipv4:host():string()..'/32')
+		uci:set(bmx6.f, "ownIPv6", "tunIn")
+		uci:set(bmx6.f, "ownIPv6", "tunIn", "ownIPv6")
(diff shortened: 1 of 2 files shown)


p4u:
If the option does not exist in lime-defaults (which will be the most common case because network-profiles usually include a customized lime-defaults file), lime-config prints a horrible WARNING message which is quite scary for standard users.

IMO it is not a good strategy that every time a new option is added to lime-defaults, the user must handle with WARNING messages (which looks like errors cause debug.traceback is used). The system SHOULD have a way to solve this by itself.

p4u:
Funy thing here is that now none of us can approve nor merge this PR. That will _speed-up_ a lot the merging for sure XD. 

Please propose someone to review the PR.

G10h4ck:
p4u:
> Hey, I made this modification cause you asked for it. But for using config.get_bool()
> (as you suggested)

I just suggested to use +config.get_bool(...)+ because is the API we consensus to use in such cases 

p4u:
> a default option is needed.

for that we have +/etc/config/lime-default+ so it is easy for the user to see all options that are available and avoid to fill all the code with +if+s that may create even more complicated corner cases

p4u:
> What do you propose to not break the libremesh installations after an upgrade if they are
> keeping an old/customized lime-defaults file?

ATM I have no other proposal then taking care of the network profiles, and have a branch for each release for that repository too like we do with the other repositories as the changes between releases usually have more far reach and complex interactions than one that could be easily fixed by adding an +if(default)+ in the configuration API.

So IMHO it is not a good idea to provide partial ~solutions~ to this problem in the name of not displaying an ~horrible WARNING message~ that is completely legitimate IMHO. Either we consensus one solution that work in all cases (for other repository we consensus to keep a branch for each release) without breaking other things or we leave things as they are. 

p4u:
> Anyway, I don't get why you closed this PR without merging anything or giving the change
> to modify it.

Because there were much changes unrelated with the PR https://github.com/libremesh/lime-packages/pull/274 topic and I think it is easier to continue in this PR https://github.com/libremesh/lime-packages/pull/278 that has just the modifications related to the topic that to revert and patch on top of unrelated material line by line commit by commit like important configuration API changes that has deep implication in the whole logic of the system and that need an assembly discussion by themselves like the deletion of +config.sanitize()+ that we consensus to add in last hackaton and adding default third parameter to both +config.get(...)+ and +config.get_bool(...)+ that we consensus to remove because it caused major impact on system logic (we figured this out because same happened in UCI Lua interface) a bunch of assemblies ago and one of the reasons was that it was breaking the automatic +/etc/config/lime+ option population in case of default parameter being used.

p4u:
> If you want to cherry-pick and send PR yourself, at least give me the chance to reply.

Just done it in this PR https://github.com/libremesh/lime-packages/pull/278 also I managed to preserve your original commit intact with authorship and hash using +git merge hash+ instead of +git cherry pick hash+

p4u:
> If the option does not exist in lime-defaults (which will be the most common case because
> network-profiles usually include a customized lime-defaults file), lime-config prints a horrible
> WARNING message which is quite scary for standard users.

The most common case is that code must be coherent to what we consensus in assembly, not that we add an if in our code for each thing that may scare a ~normal user~ that IMHO is not the one who know how to use SSH to access the nodes, edit +/etc/config/lime+ with +vi+ and then run +lime-config+, one that has all those skills isn't most probably not scared by an ~horrible WARNING message~ as she must have seen lot of them in her life to get to that point at lest the ~horrible WARNING message~ that SSH print when you connect to the node the first time because the public key is not known...

p4u:
> IMO it is not a good strategy that every time a new option is added to lime-defaults, the user must
> handle with WARNING messages (which looks like errors cause debug.traceback is used).

That can happen only between major stable releases that are not that commons and it is common sense that one (that is not any user but just who take care of the community network profiles that are people that know what they do) look for changes to be made in the configurations files between major releases of software, historically we have even broken retro-compatibility (that really impact ANY user) a bunch of times between stable releases (because we had consensus on that like we had consensus on this warning message) and we showed less concern about that than you are showing now about a simple ~horrible WARNING message~ that in the end may be seen just by someone who already know how to use terminal, ssh, vim, git, ping all software that prints warning messages very often


p4u:
> The system SHOULD have a way to solve this by itself.

Of course the system should get added all capabilities we consensus without breaking all the other we consensus before

p4u:
IMHO what makes the real difference between our point of view is that you consider this "consensus assambly" (new concept for me in LibreMesh) as the official way to make decisions. I suposo you are making reference to the face-to-face meetings that we have had in the past. For me, the way to make consensus is via some official digital channel (such as the mailing list). So the whole set of developers (code commiters) can participate in the decisions, not only those who are able to make a face-to-face meeting (usualy near Barcelona). Also I don't remember having this specific discussion in any meeting. So or my memory is not as good as I would like to be or you discuss it with someone else in a face-to-face meeting.

In addition IMHO any decision from the past can be modified if needed via some kind of public channel (Github, MailingList, IRC). But argue that a decision was taken in the past is not an argument to not discuss about changing it.

So this in one hand, I think we must deeply consider discussing about what should be the official and common way to make consensus decisions. If we already had this though an open/common/inclusive channel, sorry but I do not know.

In the other hand regarding the technical details of this issue. Also without the optional parameter I was suggesting, there is a default behavior (but it is just more hidden in the code). So if you write a code like this:

`if config.get_bool("network", "bmx6_publish_ownip") then`

If the option does not exist (so return value is Null) the behavior will be to not enter this condition (and jump to the else statement). So it is the same behavior of having false or 0. Then a decision is anyway taken, the program will keep running considering a behavior.

I was proposing this:

`if config.get_bool("network", "bmx6_publish_ownip", False) then`

Which has exactly the same result but it is more explicit in the code and more flexible cause the programmer can easily define which is the default value (note that this is done anyway by the programmer but in a implicit way). In addition a Warning message might be printed in case of using the default value, something like 

_Warning: option network.bmx6_publish_ownip is not defined in lime-defaults, using default value False_

So IMO there exist a solution/way both points of view (yours and mine) can be represented in the code operation. We just needed some more friendly discussion about it to find it. But you decided to suddenly close my Pull Request not giving any chance to keep working over the previous discussion and code. You probably did not meant it but I felt this way of doing quite violent, I felt like my work and opinions were belittled. Ths is why my reaction was not as good and rational as I would like to be, and this is why minutes after I decided to remove the post.

This text do not meant to be a flame, I just want you to know how I felt about your doings and put over the table a topic what for me requires some discussion.

p4u:
https://github.com/libremesh/lime-packages/pull/274/commits/8ea068ca153dfe16f4f192cd348fbdc5f5e0bd4a

PR #284, opened 2018-01-22
Update lime-example

aparcar:
add comment how to use the gateways DNS instead of a hard coded in lime-config


aparcar:
@ilario

aparcar:
done, should be squashed before merging

Issue #104, opened 2017-03-31
If dnsmasq is choosen in menuconfig, lime-full disappears


nicopace:
@p4u and @altergui, can you expand on this?

G10h4ck:
This not a bug, is the expected behaviour, in openwrt a packet cannot say something like -dnsmasq, you can only express incompatibilities, some of the lime packages are broken if the dnsmasq package is installed and as limefull depends on them nor them nor lime-full appears, lime-build was the one in cvharge of removing problematic packets like dnsmasq

p4u:
It is not a bug but IMO we must find a way to do it differently. I see
two use cases where the current implementation originates problems:

1. It is very weird for a end-user to enter the menuconfig and do not
see the LiMe collection packages because dnsmasq is enabled by default.
Only those who knows that dnsmasq must be disabled would be able to
select LibreMesh.

2. If a user has a node with a standard LEDE installation and wants to
install LibreMesh trough OPKG, there will be a dependency conflict and
the packages might not be installed.

So, I would remove this -dnsmasq dependency and find another way to fix
the possible conflicts among dnsmasq-dhcp6 and standard dnsmasq.

On 04/04/17 11:41, Gioacchino wrote:
> This not a bug, is the expected behaviour, in openwrt a packet cannot say something like -dnsmasq, you can only express incompatibilities, some of the lime packages are broken if the dnsmasq package is installed and as limefull depends on them nor them nor lime-full appears, lime-build was the one in cvharge of removing problematic packets like dnsmasq
> 

-- 
./p4u

dangowrt:
Maybe the best would be to implement lease-sharing via alfred to work with odhcpd? dnsmasq is PITA imho and the sooner we get rid of it, the better...

p4u:
On 04/04/17 18:35, Daniel Golle wrote:
> Maybe the best would be to implement lease-sharing via alfred to work with odhcpd? dnsmasq is PITA imho and the sooner we get rid of it, the better...
> 

I agree, we must get rid of dnsmasq.

Actually we are already using ALFRED for lease sharing. But IMHO the
best would be to use bmx6/7 SMS plugin instead. ALFRED does not work as
well as we would like to (according to our experience).

@Gui is our dnsmasq guy, would be interesting to know why are we
actually using it. So we can see if odhcp + some-tiny-dns-server would
be enough.


-- 
./p4u

dangowrt:
[...]

PR #285, opened 2018-01-23
lime-proto-bmx7: enable SMS plugin

aparcar:
The [SMS plugin](https://github.com/bmx-routing/bmx7#sms-plugin) can be very useful for various projects

See https://github.com/libremesh/lime-packages/issues/104


aparcar:
@p4u ping

G10h4ck:
Isn't the dependency on the plugin needed too? Or it just ignore it if it is not installed?

aparcar:
@G10h4ck no it's just not enabled by default

p4u:
bmx7-sms plugin is already a dependency for lime-proto-bmx7

PR #286, opened 2018-01-23
replace luci{-mod-admin-full,} to fix #260

### LINKED ISSUE #260: Missing PPPoE Luci App in WDR3500
I think is a package that is usefull to be included in the images. At least in the **lede-17.01.2-lime-default-ar71xx-generic-tl-wdr3500-v1-squashfs-factory.bin** is not included by default (although is easily installed).
Regards :-)

aparcar:
Install full luci instead of *only* the luci-mod-admin-full. The package luci-proto-ppp is additionally installed which result's in a squashed file size increase of < 1KB.

### DIFF
--- a/packages/lime-full/Makefile
+++ b/packages/lime-full/Makefile
@@ -23,7 +23,7 @@ define Package/$(PKG_NAME)
   SUBMENU:=Collections
   MAINTAINER:=Pau Escrich <p4u@dabax.net>
   URL:=http://libremesh.org
-  DEPENDS:=+lime-basic +lime-debug +luci-mod-admin-full +lime-map-agent +lime-docs +lime-app
+  DEPENDS:=+lime-basic +lime-debug +luci +lime-map-agent +lime-docs +lime-app
 endef
 
 define Package/$(PKG_NAME)/description


aparcar:
Additional packages:
```
luci-app-firewall
luci-proto-ipv6
luci-proto-ppp 
uhttpd-mod-ubus
```

Filesize
```
 7478738 4612 -rw-r--r-- 1 a a 4718765 Jan 23 20:38 lede-17.01.4-lime-default-ramips-mt7620-miwifi-mini-squashfs-sysupgrade.bin
 7478705 4612 -rw-r--r-- 1 a a 4718765 Jan 23 20:36 lede-17.01.4-lime-luci-ramips-mt7620-miwifi-mini-squashfs-sysupgrade.bin
```

aparcar:
@p4u

p4u:
luci-proto-ppp depends also on ppp and some extra kmods so I don't think it is just less than 1KB.

In any case as it only affects lime-full I would accept this change.

aparcar:
> luci-proto-ppp depends also on ppp and some extra kmods so I don't think it is just less than 1KB.

I'm sure you're right but the images I got are the same size...

dangowrt:
...because kmod-ppp, kmod-pppox, kmod-pppoe, ppp, ppp-mod-pppoe are all
already installed (because they are not being explicitely removed in
lime-full)

aparcar:
As always, thanks for that clarification

PR #288, opened 2018-01-23
Feature/sanitize debug output when default used

p4u:
details on #278


p4u:
looks wrong, I'll review and send it again. sorry

PR #290, opened 2018-01-23
Introduce an optional default value in config.get()

p4u:
Method config.get() function can now be called with a third (optional) value
if it's not Null and the option is not found, will be the returned value.

This allows the easy integration of new options which might be not (yet)
introduced in lime-defaults of a node after upgrading.

A Warning message is printed if default value used.

details on #278

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/config.lua
@@ -26,13 +26,17 @@ function config.sanitize()
 	end
 end
 
-function config.get(sectionname, option)
+function config.get(sectionname, option, default)
 	local limeconf = config.uci:get("lime", sectionname, option)
 	if limeconf then return limeconf end
 
 	local defcnf = config.uci:get("lime-defaults", sectionname, option)
 	if ( defcnf ~= nil ) then
 		config.set(sectionname, option, defcnf)
+	elseif ( default ~= nil ) then
+		defcnf = default
+		local cfn = sectionname.."."..option
+		print("WARNING: Option "..cfn.." is not defined. Using value "..default)
 	else
 		local cfn = sectionname.."."..option
 		print("WARNING: Attempt to access undeclared default for: "..cfn)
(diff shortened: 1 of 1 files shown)


G10h4ck:
BTW why is the CI failing?

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -26,13 +26,17 @@ function config.sanitize()
...
 	local limeconf = config.uci:get("lime", sectionname, option)
 	if limeconf then return limeconf end
 
 	local defcnf = config.uci:get("lime-defaults", sectionname, option)
 	if ( defcnf ~= nil ) then
 		config.set(sectionname, option, defcnf)
+	elseif ( default ~= nil ) then
+		defcnf = default
+		local cfn = sectionname.."."..option
+		print("WARNING: Option "..cfn.." is not defined. Using value "..(default or 'False'))
```
G10h4ck: I think `(default or 'False')` is not needed just appending default should work for any type that may get there (it would work also with nil but we are sure it will not get there in any case because of the elseif)

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -71,8 +75,8 @@ function config.get_all(sectionname)
 	return nil
 end
 
-function config.get_bool(sectionname, option)
-	local val = config.get(sectionname, option)
+function config.get_bool(sectionname, option, default)
+	local val = config.get(sectionname, option, default)
```
G10h4ck: Please double check if config.autogenerable(section_name) keep working with this new implementation

packages/lime-system/files/usr/lib/lua/lime/config.lua
p4u: You'r right

packages/lime-system/files/usr/lib/lua/lime/config.lua
p4u: Why should it not work? The default parameter is the last parameter so it can be just "nil".

p4u:
> BTW why is the CI failing?

In case CI fails nothing will work, not even the standard OpenWRT configuration. I think it will be quite obvious that something is not working in UCI side.

Also this modification only affects such parameters that are defined with a default value which are very few right now. It is not expected that all options use this method. Only those not so relevant for the LibreMesh operation (I would not use this default option for main_ipv4_address for instance). At the end it is at the hands of the developer to decide if the option should implement a default value or not.

IMO CI errors are not common but adding new, not so relevant, values which are not in lime-default (cause the user is using an outdated configuration) is quite common.

PR #291, opened 2018-01-24
lime-proto-bmx7: update script and options

p4u:
Follows bmx6 changes.

Still not tested in a real, production node. Waiting for comments...

### DIFF
--- a/packages/lime-proto-bmx7/src/bmx7.lua
+++ b/packages/lime-proto-bmx7/src/bmx7.lua
@@ -24,6 +24,8 @@ function bmx7.configure(args)
 	uci:set(bmx7.f, "general", "bmx7")
 	uci:set(bmx7.f, "general", "dbgMuteTimeout", "1000000")
 	uci:set(bmx7.f, "general", "tunOutTimeout", "100000")
+	uci:set(bmx7.f, "general", "configSync", "0")
+	uci:set(bmx7.f, "general", "syslog", "0")
 
 	uci:set(bmx7.f, "main", "tunDev")
 	uci:set(bmx7.f, "main", "tunDev", "main")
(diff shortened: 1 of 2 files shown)


packages/lime-proto-bmx7/src/bmx7.lua
```diff
@@ -37,7 +39,7 @@ function bmx7.configure(args)
 	-- Enable JSON plugin to get bmx7 information in json format
 	uci:set(bmx7.f, "json", "plugin")
 	uci:set(bmx7.f, "json", "plugin", "bmx7_json.so")
-	
+
 	-- Enable SMS plugin to enable sharing of small files
```
aparcar: @axn does the SMS plugin add considerable overhead when unused? Maybe this should be configurable but default to true.

packages/lime-system/files/etc/config/lime-defaults
```diff
@@ -17,6 +17,7 @@ config lime network
 	option anygw_dhcp_limit '0'
 	option main_ipv6_address '2a00:1508:0a%N1:%N200::/64'
 	option bmx6_mtu '1500'
+	option bmx7_mtu '1500'
```
aparcar: If 1500 is fallback there is no need to put it in lime-default but in lime-examples right?

aparcar:
Tested `bmx7_mtu` and `bmx7_pref_gw` and looks good to me.

packages/lime-system/files/etc/config/lime-defaults
p4u: No. By design all available options must exist in lime-defaults. In addition the MTU is not fallback, it is a needed parameter.

packages/lime-proto-bmx7/src/bmx7.lua
p4u: AfAIK SMS plugin does nothing if it's not used.

PR #292, opened 2018-01-24
lime-defaults: remove unused olsr

aparcar:
olsr is not installed by default and so shouldn't be configured per default. 
This should slightly clean up lime-defaults to be easier readable.


ilario:
Similar (IMHO) to #174

aparcar:
>Similar (IMHO) to #174

Okay if so please close

p4u:
Yep, those options should NOT be removed. Closing.

PR #295, opened 2018-01-25
lime-hwd-openwrt-wan: use board.json instead of copying /etc/config

### LINKED ISSUE #166: Avoid copy of /etc/config on /etc/lime-config-firstboot
On first boot the directory /etc/config is fully copied to /etc/lime-config-firstboot

Is it really needed? I'm not able to find a use case.

Related commit: https://github.com/libremesh/lime-packages/commit/de31181ed4e4a0c9c2249848cf3a5418295dcc70

aparcar:
my first ever contact with lua, feel free to close this PR straight away


packages/lime-hwd-openwrt-wan/Makefile
```diff
@@ -34,6 +34,9 @@ define Package/$(PKG_NAME)/install
 	@mkdir -p $(1)/usr/lib/lua/lime/hwd || true
 	$(CP) ./build/openwrt_wan.lua $(1)/usr/lib/lua/lime/hwd/openwrt_wan.lua
 	@chmod +x $(1)/usr/lib/lua/lime/hwd/openwrt_wan.lua
+
+	$(CP) ./files/openwrt_wan.sh $(1)/usr/lib/lua/lime/hwd/openwrt_wan.sh
```
aparcar: should /src

p4u:
@G10h4ck is better for review this, since he was the author of lime-hwd-openwrt-wan

G10h4ck:
Please use descriptive titles for the pull requests, it's in the title contain a reference a issue but if the title is just the issue it is not good for who do the review.

G10h4ck:
About the code, we agreed that libremesh would be written mainly in Lua keeping shellism infiltrations at a minimum,

This PR instead rewrite from scratch the wole module in shell, AFAIK our agreement on writing most of the code in Lua hasn't changed, so I would sk that you write in non-Lua only the minimum required part

Cheers!
Gio

aparcar:
@G10h4ck I was not aware of this decision, could you please provide a link to the discussion? I'm just curios.

I reworked the setup and and now shell is used as minimal as I could think of. 

Tested on a x86 vm

p4u:
IMO the shell code included now in this PR is quite small and required to use the standard OpenWRT functions which are made in shell script. 

So if it works I see it as a better and more standard way to do the same job.

G10h4ck:
It is ok for me now, thanks you!

PR #301, opened 2018-01-28
network.lua: set again the prefix in case of invalid IPv4 address, fix #287

### LINKED ISSUE #287: IPv4 parametrization
> option main_ipv4_address '10.%N1.0.0/16'	

>  Here you have 4 possibilities: set a static IP and the subnet, like '10.0.2.1/16'; parametrize with %Mn and %Nn, and set the subnet, like '10.%N1.%M5.%M6/16'; set a whole network address (not a specific IP) for getting the IP autocompleted in that network with bits from MAC address, this works also with netmasks other than /24 or /16, like '10.0.128.0/17' (but not valid network addresses, for example '192.0.128.0/16' or '192.0.129.0/17' won't get parametrized); set two different parameters, the first for subnet and the second for IP parametrization, like '10.0.128.0/16/17', this results in /16 subnet but IP parametrized in a /17 range (from 10.0.128.1 to 10.0.255.254).
> 	option main_ipv6_address '2a00:1508:0a%N1:%N200::/64'	# Parametrizable in the same way as main_ipv4_address. If used, the IP autocompletion will fill maximum the last 24 bits, so specifying an IP autocompletion range bigger than /104 is not useful.
>

So according to lime documentation, these are the possible configuration options:

1. `10.0.2.1/16` So the node will self configure 10.0.2.1/16
2. `10.%N1.%M5.%M6/16` The node will self configure an IP based on cloud ID (essid) and last MAC bytes, the result would be something like: 10.2.3.4/16
3. `10.0.128.0/16/24` So first is the Netmask (which will be 255.255.0.0) and second is the IP space to autoconfigure the IP. Then a random IP from 10.0.128.0/24 will be used but the LAN range will still be /16

Why something like this `10.2.%M5.1/27` does not work (netmask=255.255.255.255 after running lime-config)? Is it not the second case?

Also I'm missing a way to specify the same of the third case but the other side around, so:

10.2.0.0/27/16 -> meaning that the node should self configure a /27 of the biggest range /16, something like 10.2.5.33/27

Until now I though the third scenario was refering to this (I even used in some nodes and it worked fine!) but after reading the lime-example documentation and experiencing some strange behavior in some nodes, I think it is for some "random" reason that it worked for me.

Can we discuss about it? Am I right with this configuration description? @G10h4ck and @ilario where the two modifying these lines of code.

ilario:
As described in #287, network.lua was correctly identifying invalid IPv4 (in the specific case IPs which were overlapping with anygw IP, so the first of the domain) but was not setting correctly the netmask in this case.
Let me know if there is a more elegant way to do this, I expected it was possible to change just the IPv4 address with just one command without changing the prefix (something like `ipv4_template:host(mc:maxhost())`) but I couldn't find it.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -68,7 +68,10 @@ function network.primary_address(offset)
 	invalid = invalid or ipv4_template:equal(mc:minhost())
 	--! Generated address is the broadcast address like 192.0.2.255/24 ?
 	invalid = invalid or ipv4_template:equal(mc:broadcast())
-	if invalid then ipv4_template = mc:maxhost() end
+	if invalid then
+		ipv4_template = mc:maxhost()
+		ipv4_template:prefix(tonumber(ipv4_maskbits))
+	end
 
 	ipv6_template:prefix(tonumber(ipv6_maskbits))
 


p4u:
If anygw is not cofigured, the IP addresses should be valid, is this taken into account?

ilario:
Nope, not at all. I'm not sure how we could do that check... @G10h4ck any thought?

G10h4ck:
It can be checked by looking at the list of enabled protocols

ilario:
> It can be checked by looking at the list of enabled protocols

anygw can be enabled in /etc/config/lime* but `lime-proto-anygw` be absent in the installation. This is the same concept of OLSR being enabled in the configuration files but `lime-proto-olsr*` not being installed, see discussion on #292.
How can be one actually check if anygw is active?

ilario:
All this discussion on anygw is NOT related to the PR, can we please create a new issue and move this discussion on anygw there?

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -68,7 +68,10 @@ function network.primary_address(offset)
 	invalid = invalid or ipv4_template:equal(mc:minhost())
 	--! Generated address is the broadcast address like 192.0.2.255/24 ?
 	invalid = invalid or ipv4_template:equal(mc:broadcast())
-	if invalid then ipv4_template = mc:maxhost() end
+	if invalid then
+		ipv4_template = mc:maxhost()
```
G10h4ck: Are you sure this will not cause easier collision in common situation?

ilario: Do you mean that the lack of prefix resulting in a /32 was intentional?
I'm not sure I got your point, do you mean that there will be a lot of collisions on the maxhost() IP? In this case, sure there will be, but for avoiding this we should entirely redesign this section of the code. If you think my PR is deepening the problem, please explain this point.

ilario:
@p4u @G10h4ck I moved the discussion about anygw on #313

Issue #306, opened 2018-01-30
separate Mesh and AP radio based on frequencies 


aparcar:
Using an encrypted 11s mesh and encrypted AP on the same radio (2.4GHz or 5GHz) does not work as reported [here](https://github.com/libremesh/lime-packages/issues/208).

To archive this feature it's [possible](https://github.com/ilario/lime-packages/blob/dee5ba5da8d54b0fec744070269b4db33cba7de4/packages/lime-docs/files/lime-example#L76) to set individual configurations, but due to the usage of [foreach](https://github.com/libremesh/lime-packages/blob/6e152fc2f08838f722ac5240a1c2ec81c11b1994/packages/lime-system/files/usr/lib/lua/lime/config.lua#L48) in the [wireless.lua](https://github.com/libremesh/lime-packages/blob/6e543c79a08e41482f1fdd0214dbf93e97f5d312/packages/lime-system/files/usr/lib/lua/lime/wireless.lua#L78) module it's not possible to set that in `lime-defaults'.

p4u:
Follows #307

p4u:
My proposal is a module named (for instance) lime-smart-wifi-config

This module must run before lime-config (via uci-defaults script).

The module looks for the existing WiFi radios in the system (for such task, LUA funcions from lime-system might be used) and writes to /etc/config/lime the specific sections according to a configuracion file or a set of heuristics (for those which are not in the config file).

This is a proposal for the default heuristics (empty config file).

If only 1 radio
1. If 2.4GHz or 5Ghz -> Set it to the correct channel and AP + MESH
2. If Dual Band -> Set it to 2.4GHz and AP + Mesh

If more than 1 radio
1. If only one at 2.4GHz -> set it to only AP
2. If more than one at 2.4GHz -> set the first found to only AP
3. Set the rest of radios at 5GHz (prefered) or 2.4Ghz -> only Mesh 

Channels must not collide! So if two radios at 2.4Ghz, the first will get one channel and the second another. lime-defaults file can be used for getting the first channel, for instance "channel_2ghz = 11" the next will be 1 and the next 6 (always 1, 6 or 11).

Let's say config file is /etc/config/lime-smart-wifi-config

These are a set of options that might exist in this file

```
option all_to_mesh = 1/0
option all_to_ap = 1/0
option mesh_2ghz = 1/0
option mesh_5ghz = 1/0
option ap_2ghz = 1/0
option ap_5ghz=1/0
[...]

Issue #208, opened 2017-08-10
How to configure an encrypted mesh based on ieee80211s


netzcoop:
hello there,
(we also asked this on lime-users@lists.libremesh.org on 09.08.17 23:29)

we try to setup different customized lime installations for a camp.

our routers:  tplink 4300.

1) the following configuration is working for ap encryption but not for
the mesh:

* cooking in branch stable (we also tried develop with the latest
working commit):

 V=s J=1 ./cooker -c ar71xx/generic --flavor=lime_camp
--profile=tl-wdr4300-v1 --community=camp/commoniee80211s

* flavors.config

lime_camp="lime-full -dnsmasq -wpad-mini wpad authsae"

* etc/config/lime_defaults:

config lime system
    option hostname 'CA-%M4%M5%M6'
    option domain 'mesh.camp.local'
    option keep_on_upgrade 'libremesh base-files-essential
/etc/sysupgrade.conf'

config lime network
    option primary_interface eth0
    option main_ipv4_address '10.5.0.0/21'
    option main_ipv6_address '2801:01e8:2::/64'
    option bmx6_mtu '1500'
    list protocols ieee80211s
    list protocols lan
    list protocols anygw
    list protocols batadv:%N1
    list protocols bmx6:13
    list protocols olsr:14
    list protocols olsr6:15
    list protocols olsr2:16
    list resolvers 4.2.2.2   # b.resolvers.Level3.net
    list resolvers 141.1.1.1 # cns1.cw.net
    list resolvers 2001:470:20::2 # ordns.he.net
    option bmx6_over_batman false
    option bmx6_pref_gw none
    option bmx7_over_batman false
    option anygw_mac "aa:aa:aa:%N1:%N2:aa"

config lime wifi
    option channel_2ghz '11'
    option channel_5ghz '48'
    option htmode_5ghz 'HT40'
    list modes 'ap'
    list modes 'apname'
    list modes 'ieee80211s'
    option ap_ssid 'Unsecure'
    option apname_ssid 'Unsecure | %H'
    option adhoc_ssid 'Mesh | Camp'
    option adhoc_bssid '16:27:F6:27:57:14'
    option adhoc_mcast_rate_2ghz '24000'
    option adhoc_mcast_rate_5ghz '6000'
    option ieee80211s_mesh_fwding '0'
    option ieee80211s_mesh_id 'Mesh_Camp'
    option ap_encryption 'psk2+ccmp'
    option ap_key 'somepassword'
[...]

PR #307, opened 2018-01-30
lime/wireless: activate wifi modes based on suffix

aparcar:
Add the option to add `_2ghz` and `_5ghz` as a suffix to the
activated wifi modes (ap, apname, ieee80211s).

This is usefull as the specific radio configuration (radio0, radio1,
etc) only works with /etc/config/lime but not with lime-defaults.

When using encrypted mesh and encrypted AP it comes to an error
mentioned [here](https://github.com/libremesh/lime-packages/issues/208).

Fixes https://github.com/libremesh/lime-packages/issues/306

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -60,8 +60,11 @@ config lime wifi
 	list modes 'ap'
 	list modes 'apname'
 	list modes 'ieee80211s'
-#	list modes 'adhoc'				# See below for adhoc configuration
-#	list modes 'client'				# See below for client configuration
+#	list modes 'adhoc'            # See below for adhoc configuration
+#	list modes 'client'           # See below for client configuration
+#	list modes 'ap_2ghz'          # Enable Access Point only on 2.4GHz Radio
+#	list modes 'apname_2ghz' 
+#	list modes 'ieee80211s_5ghz'  # Enable Mesh network only on 5Ghz Radio
 	option ap_ssid 'LibreMesh.org'
 #	option ap_key 'SomeWPA2PskKey'
 #	option ap_encryption 'psk2'
@@ -72,6 +75,7 @@ config lime wifi
 	option ieee80211s_mesh_id 'LiMe'
 #	option ieee80211s_encryption 'psk2/aes'
 #	option ieee80211s_key 'SomePsk2AESKey'
+
 
 
 # The following interface specific options have to be included in /etc/config/lime, not in /etc/config/lime-defaults
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -113,31 +113,35 @@ function wireless.configure()
 			uci:save("wireless")
 
 			for _,modeName in pairs(modes) do
-				local args = {}
-				local mode = require("lime.mode."..modeName)
+				if not modeName:match(ignoredSuffix) then
+					modeName = modeName:gsub(freqSuffix, "")
 
-				for key,value in pairs(options) do
-					local keyPrefix = utils.split(key, "_")[1]
-					local isGoodOption = ( (key ~= "modes")
-					                and (not key:match("^%."))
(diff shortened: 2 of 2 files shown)


p4u:
Great, I like the idea to be able to configure the radios according to the band. However:

1. What happens if a radio is Dual Band (2.4 and 5) ?
2. What happens if "mode ap" and "mode ap_5ghz" are enabled at same time, radio will be configured both times and the last will survive?

aparcar:
> What happens if a radio is Dual Band (2.4 and 5) ?

If never seen such device nor how it's configured. Can you give more details?
Also it's possible two have two 5GHz bands (like LibreRouter) and the configuration wouldn't cover that case neither. Still it covers basically all dual band routers which I'd put in focus here. For more specific configuration the `lime` config could be used.

> What happens if "mode ap" and "mode ap_5ghz" are enabled at same time, radio will be configured both times and the last will survive?

I think the last one will survive and I'd keep it like that to avoid complexity. Second choice would be to rename the interface to make both possible. To illustrate:

`
list modes ap
list modes ap_5ghz
option ap_ssid 'LibreMesh.org'
option ap_ssid_5ghz 'LibreMesh.org 5'
`
This would result in 1 AP on 2.4Ghz and 2 APs on 5GHz, allowing clients to automatically switch between 2.4 and 5GHz when ssid 'LibreMesh.org' is selected or explicitly use 5Ghz.

p4u:
> If never seen such device nor how it's configured. Can you give more details?

The market is full of dual band WiFI radios, just look for "ABGN" radio. Atheros QCA9557 for instance (example router https://www.8devices.com/products/rambutan). Look here for more examples https://wikidevi.com/wiki/Qualcomm_Atheros

> Also it's possible two have two 5GHz bands (like LibreRouter) and the configuration wouldn't cover that case neither. Still it covers basically all dual band routers which I'd put in focus here. For more specific configuration the lime config could be used.

This is a pending task (@G10h4ck was gonna do it), to have lists of channels instead of single options so "list channel_5ghz 48" "list channel_5ghz 64" would set the first 5Ghz radio to 48 and the second to 64. But is not yet implemented.

> This would result in 1 AP on 2.4Ghz and 2 APs on 5GHz, allowing clients to automatically switch between 2.4 and 5GHz when ssid 'LibreMesh.org' is selected or explicitly use 5Ghz.

I'm thinking now that this is getting too complex. Specific WiFi sections exist and they must be used.

Maybe it would be better to develop a libremesh module "lime-smart-wifi" (or something like this) which can be installed (or not) and applies some "smart" heuristics in the first boot to let the WiFi devices configured properly creating the WiFi specific sections in lime config file.

aparcar:
> The market is full of dual band WiFI radios [...]

Thanks, I'll check that out.

> I'm thinking now that this is getting too complex. Specific WiFi sections exist and they must be used.

So can we stick to the herewith proposed implementation (for now)?

p4u:
As some uses cases are not covered and the general configuration section is getting complex and a bit inconsistent, I'd not follow your approach. Instead I would do as I propose (a separate module to autoconfigure independently the radios in specific sections).

p4u:
By _inconsistent_ I mean that your approach is mixing Radio Mode (AP/Managed/Mesh/Adhoc) with Frequency channel. The design of lime config is to have these two layers separated by different options (you can select Ap and Mesh but only one channel per radio (i.e 11), but you cannot select AP at 1 and Mesh at 11).

aparcar:
> The market is full of dual band WiFI radios, just look for "ABGN" radio. Atheros QCA9557 for instance (example router https://www.8devices.com/products/rambutan). Look here for more examples https://wikidevi.com/wiki/Qualcomm_Atheros

@rogerpueyo do you have anything like that in the lab?

@p4u so a smart wifi plugin could have an option like separate mesh + ap and then does the configuration instead of wireless.lua?

rogerpueyo:
I have two https://mikrotik.com/product/RBwAPG-5HacT2HnD, which are dual-band (2,4 GHz radio + 5 GHz radio).

p4u:
Let's keep on #306

Issue #304, opened 2018-01-29
RFC: mesh cloud configuration


aparcar:
To simplify the configuration of a mesh cloud I'd like to hear your comments and thoughts of the following setup. 

In the following I'd like to distinguish between the *initial* and the *running* configuration. *Initial* means they run any OpenWrt based software where a `sysupgrade` is easily possible. *Running* means the nodes are already in place and only (or mostly) connected by wireless mesh.

Before going into any details, with the configuration tool the following aspects should be changeable:

* Mesh wide
  * AP + Mesh
    * Name
    * Encryption
    * Channel
  * *Routing protocols*
  * *11s/Ad hoc*
  * *Login credentials*
  * *SSH Keys*
* Device specific
  * Hostname

*Italic* values are optional or unsure.

# Initial

There are already *easy* to use tools like [Chef](https://chef.libremesh.org) and the LibreMesh [Cooker](https://github.com/libremesh/lime-sdk). While these tools can use existing [Network Profiles](https://github.com/libremesh/network-profiles/) to preinstall a specific node configuration, it can be challenging for some users to create these profiles. In the following possible solutions to simplify that process.

A possible solution would be to create a *Chef* like web interface to create *Network Profiles* and output packages like the [Network Profiles Builder](https://github.com/libremesh/network_profiles_builder). These packages are useable to use by *Chef*. And offer versioning for later configuration changes.

Another approach could be to use generic images for *all* nodes and let them fetch a configuration on first boot. This was implemented in the [FirstBootWizard](https://github.com/libremesh/FirstBootWizard/) Instead of *only* pulling the [lime-defaults](https://github.com/libremesh/lime-packages/blob/develop/packages/lime-system/files/etc/config/lime-defaults) it would be possible to install a [*config.ipk*](http://downloads.libremesh.org/network-profiles/) containing a more complex setup (including SSH keys etc.).

## Individual hostnames

This approach does not cover setting names to specific devices. This could be done by creating a table of [*MAC*](https://en.wikipedia.org/wiki/MAC_address) and *hostname*. The node starts up and read the table, compares it with it's own active mac addresses. On matching a *MAC* it will set the *hostname* accordingly. 

The *hostname* table could be distributed with the help of [BMX7 SMS plugin](https://github.com/bmx-routing/bmx7#sms-plugin). 

# Running

[...]

PR #311, opened 2018-02-03
lime-bmx7-rc: add package & RCF

aparcar:
lime **c**loud **c**ontroller has two parts one is lime-cc-client and one is lime-cc-server

* lime-cc-client
check every 5 minutes for a new configuration received by bmx7-sms.
if the configuration is valid and the node is trusted the configuration
is used as default, lime-config and lime-apply runs.
if not, the configuration is ignored.

* lime-cc-server
distributes it own lime-defaults to the cloud. If changes happen to the
config it will be distributed to the cloud.

this is a prove of concept for one of the approaches mentioned
[here](#304)

bmx7-sms is required to make this working

lime-cc: use uci options

uci option to set trusted nodes in lime-defaults

```
config trusted_nodes 'trusted_nodes'
    list node_id '<node_id1>'
    list node_id '<node_id2>'
    list node_id '<node_id3>'
```

if a lime-defaults file is received it will automatically overwrite the
existsting lime-defaults and lime-{config,apply} runs.


packages/lime-cc/files/lime-cc-client.sh
```diff
@@ -0,0 +1,34 @@
...
+if [[ $? == 1 ]]; then
+    echo "please add section trusted_nodes in lime-defaults"
+    exit 1
+fi
+
+echo "checking"
+# get all recevied configs
+for config_path in /var/run/bmx7/sms/rcvdSms/*lime-defaults; do
+    # check if installed config is older then received config
+    if [[ /etc/config/lime-defaults -ot "$config_path" ]]; then
```
p4u: IMO it would be better to compare MD5SUM or SHA256SUM instead. Cause if bmx7 is restarted files will be recreated and config will be applied. Also on each reboot the file would be new so executed again.

This sum might be stored into /etc/config/lime so the even after reboot it will be saved.

packages/lime-cc/Makefile
```diff
@@ -0,0 +1,60 @@
...
+		$(INSTALL_DIR) $(1)/etc/init.d
+		$(CP) files/lime-cc-client.procd $(1)/etc/init.d/lime-cc-client
+endef
+
+define Package/lime-cc-server
+  SECTION:=lime
+  CATEGORY:=Lime
+  TITLE:=lime-cc-server
+  PKGARCH:=all
+  DEPENDS:=+bmx7-sms
```
p4u: It should also depend on bmx7 and lme-system.

p4u: Also busybox cron (see smonit package)

aparcar: `bmx7-sms` depends on `bmx7` afaik, however I'll add `lime-system` thanks

packages/lime-cc/files/lime-cc-client.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+
+trusted_ids=$(uci get lime-defaults.trusted_nodes.node_id 2>/dev/null)
```
p4u: Please use "uci -q get" (instead of redirecting to dev null)

packages/lime-cc/files/lime-cc-client.sh
```diff
@@ -0,0 +1,34 @@
...
+
+if [[ $? == 1 ]]; then
```
p4u: I would better use [ -z "$trusted_ids" ] && ( echo "...." ; exit 1)

packages/lime-cc/files/lime-cc-client.sh
```diff
@@ -0,0 +1,34 @@
...
+    # check if installed config is older then received config
+    if [[ /etc/config/lime-defaults -ot "$config_path" ]]; then
+        echo "found new config"
+        # get filename without path
+        config_file="$(basename $config_path)"
+        # parse node id
+        node_id="${config_file%%:*}"
+        # check if node is trusted
+        for trusted_id in $trusted_ids; do
+            if [[ $node_id == $trusted_id ]]; then
```
p4u: use "$var" when comparing strings or an empty var will crash the sentence

packages/lime-cc/files/lime-cc-server.defaults
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+uci set bmx7.sms='plugin'
+uci set bmx7.sms.plugin='bmx7_sms.so'
+
+uci set bmx7.lime_defaults='syncSms'
+uci set bmx7.lime_defaults='lime-defaults'
+
+
```
p4u: It looks like you like empty lines... I would try to use less empty lines in general for the code

packages/lime-cc/Makefile
```diff
@@ -0,0 +1,60 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lime-cc
+PKG_VERSION:=0.0.2
+PKG_RELEASE:=2
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/lime-cc-client
```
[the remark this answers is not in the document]
aparcar: As you're the naming guru, you have a suggestion? `lime-cc-bmx7`? It could be split in different modules, one for `bmx7`, another for `bmx6` + `usign` signature, one for server polling?

packages/lime-cc/files/lime-cc-client.sh
aparcar: Good idea, I'll look into that. Genereally I thought of installing a `opkg` package instead of just the `lime-defaults` to actually enable ssh-key adding or whatever is required. THis would also enable versioning without the need of "age" checks.

packages/lime-cc/files/lime-cc-client.procd
```diff
@@ -0,0 +1,21 @@
...
+start() {
+        if ! ( grep -q /usr/bin/lime-cc-client /etc/crontabs/root 2>/dev/null ) ; then
+                echo "*/5 * * * * /usr/bin/lime-cc-client > /tmp/log/lime-cc-client.log" >> /etc/crontabs/root
+                /etc/init.d/cron enable
+                /etc/init.d/cron start
+        fi
+}
+
+stop() {
+        sed -r -i -e '\|^([^\t ]+[\t ]+){5}/usr/bin/lime-cc-client > /tmp/log/lime-cc-client.log$|d' /etc/crontabs/root
```
p4u: please, trunkate long lines like this one

packages/lime-cc/Makefile
p4u: Even if bmx7-sms depends on bmx7 I would add bmx7 so the package do not depend on bmx7-sms package. Also it's a good practice to easy show what are the real dependencies.

packages/lime-cc/Makefile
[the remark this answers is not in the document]
p4u: I'd set it to something like lime-bmx7-remote-controller Since the conept "cloud" applies to link-local domains, so bat-adv.

p4u:
Please add or reply the requested modifications.

G10h4ck:
this has an hard dependency on bmx7 protocol so the package name should be more explicit on this for example including bmx7 in the name. Also cc is very hermetic a more flat name should be used

aparcar:
@p4u if you find the time please let me know what you think

aparcar:
I'll close this PR as I'm developing a more generic tool [here](https://github.com/aparcar/meshrc/tree/master/packages/meshrc)

PR #316, opened 2018-02-05
lime-smart-wifi: add package, fix #306

### LINKED ISSUE #306: separate Mesh and AP radio based on frequencies

aparcar:
fix separate Mesh and AP radio based on frequencies #306

1 radio with dual band is not supported due to missing device for
  testing, this will be added.

automatically swaps channels if multible radios on same freq exist

tested on tp link wdr 4300

also introduce new function to utils to check if an element is inside an
array

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -102,12 +102,13 @@ function wireless.configure()
 			--! up to 10km links by default
 			local distance = options["distance"..freqSuffix] or options["distance"] or 10000
 			local htmode = options["htmode"..freqSuffix] or options["htmode"]
+			local channel = options["channel"..freqSuffix] or options["channel"]
 
 			local uci = libuci:cursor()
 			uci:set("wireless", radioName, "disabled", 0)
 			uci:set("wireless", radioName, "distance", distance)
 			uci:set("wireless", radioName, "noscan", 1)
-			uci:set("wireless", radioName, "channel", options["channel"..freqSuffix])
+			uci:set("wireless", radioName, "channel", channel)
 			if options["country"] then uci:set("wireless", radioName, "country", options["country"]) end
 			if htmode then uci:set("wireless", radioName, "htmode", htmode) end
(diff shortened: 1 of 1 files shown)


aparcar:
@p4u if you find the time please cleanup the code a bit, I have no clues on what style is appropriate in *lua*

If I understood correctly *wpad-mesh* can do the channel selection automatically, maybe an additional check should if it's installed and change the default "channels" to none

aparcar:
depends on https://github.com/libremesh/lime-packages/pull/318

packages/lime-smart-wifi/src/lime-smart-wifi.lua
```diff
@@ -0,0 +1,82 @@
...
+local wireless = require("lime.wireless")
+local utils = require("lime.utils")
+
+local all_radios = wireless.scandevices()
+
+smart = {}
+
+smart.channels = {}
+smart.channels["2ghz"] = config.get("wifi", "channels_2ghz", { 1, 11, 6 })
+smart.channels["5ghz"] = config.get("wifi", "channels_5ghz", { 36, 48, 149 })
```
p4u: Should be based on country code, you can get this information (available channels) from iwinfo library

PR #317, opened 2018-02-05
config.lua: get_bool & defcnf hotfix

aparcar:
config.get_bool: 
If used with real boolean values like 1 or true instead of strings like "1" and "true" the function fails. 

config.get:
when the default value is a array/table the print message fails


aparcar:
@p4u done

PR #318, opened 2018-02-05
wireless.lua: fallback channel option

aparcar:
If specific radio information are set without a suffix the current configuration fails.
To fix this if *channel_<suffix>* ins't found it will fall back to *channel*

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -102,12 +102,13 @@ function wireless.configure()
 			--! up to 10km links by default
 			local distance = options["distance"..freqSuffix] or options["distance"] or 10000
 			local htmode = options["htmode"..freqSuffix] or options["htmode"]
+			local channel = options["channel"..freqSuffix] or options["channel"]
 
 			local uci = libuci:cursor()
 			uci:set("wireless", radioName, "disabled", 0)
 			uci:set("wireless", radioName, "distance", distance)
 			uci:set("wireless", radioName, "noscan", 1)
-			uci:set("wireless", radioName, "channel", options["channel"..freqSuffix])
+			uci:set("wireless", radioName, "channel", channel)
 			if options["country"] then uci:set("wireless", radioName, "country", options["country"]) end
 			if htmode then uci:set("wireless", radioName, "htmode", htmode) end
 			uci:save("wireless")


packages/lime-system/files/usr/lib/lua/lime/wireless.lua
```diff
@@ -102,12 +102,13 @@ function wireless.configure()
 			--! up to 10km links by default
 			local distance = options["distance"..freqSuffix] or options["distance"] or 10000
 			local htmode = options["htmode"..freqSuffix] or options["htmode"]
+			local channel = options["channel"..freqSuffix] or options["channel"]
```
p4u: I don't get whan options["channel"] is, AFAIK there not exist such option, channel is 2.4Ghz or 5Ghz, but not "any". You might use "auto" instead to let the radio decide what channel to use.

aparcar:
```
# /etc/config/wireless
config wifi-device 'radio0'
        option type 'mac80211'
        option hwmode '11a'
        option path 'pci0000:00/0000:00:00.0/0000:01:00.0'
        option disabled '0'
        option noscan '1'
        option htmode 'HT40'
        option distance '10000'
        option channel '36'

# /etc/config/lime
config wifi 'radio0'        
        option channel '36'  
        list modes 'ieee80211s'    
```

p4u: Fine, you are talking about specific radio configuration.

BTW, distance=10000 is a bad value as default, it should be 1000 (another PR?)

aparcar: > BTW, distance=10000 is a bad value as default, it should be 1000 (another PR?)

https://github.com/libremesh/lime-packages/commit/400ff710b54edd66c79b6959d904fa0d8a1ee471

Maybe important?

G10h4ck: yes it is important that come out of a complex consensus, the rough motivation is that if you set it to 1Km it will work better for short links but will not work at all for a bit longer links, while if you put it to 10Km it will be slower for short links but long links will works ok by default.

p4u: We got this discussion last summer and the consensus was to reduce it to 1KM, which is in general a more common scenario

https://github.com/libremesh/lime-packages/commit/d69f1aa5d0db49061b7f0901c9547ff768c93158
https://github.com/libremesh/lime-packages/issues/201

PR #319, opened 2018-02-06
wireless.lua: reduce default distance to 1000

aparcar:
following the [lime-defaults](https://github.com/libremesh/lime-packages/blob/develop/packages/lime-system/files/etc/config/lime-defaults) the hard coded default shouldn't be that big as most connections are under 10km.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -100,7 +100,7 @@ function wireless.configure()
 			end
 
 			--! up to 10km links by default
-			local distance = options["distance"..freqSuffix] or options["distance"] or 10000
+			local distance = options["distance"..freqSuffix] or options["distance"] or 1000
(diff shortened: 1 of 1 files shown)


ilario:
+1 (we could even think of different distances for radios at 2.4 GHz usually used for AP, distance 100 m, and radios at 5 GHz usually used for long distance links, distance 1000 or more).

aparcar:
@p4u comments? I can just add that as well

G10h4ck:
Please see https://github.com/libremesh/lime-packages/pull/318#discussion_r166285228

p4u:
The modification from 10 to 1KM was done here https://github.com/libremesh/lime-packages/commit/d69f1aa5d0db49061b7f0901c9547ff768c93158 but that hardcoded value was not changed (by mistake).

Part of the discussion was here https://github.com/libremesh/lime-packages/issues/201 but also during the meetings in Summer.

So, unless you want to open this discussion again I would keep the previous decision which is keep it to 1KM.

However IMO Ilario's proposal is probably the best solution.

PR #320, opened 2018-02-07
lime-system: add function arrayConcat to fix wifi

aparcar:
Currently setting specific radio options lead to a total overwrite of
all settings provided in lime and lime-defaults.

Changing just the channel results in a bad config as ssid, etc. are empty.

Instead, this patch introduces the tableConcat function where the
specific settings are attached to the defaults and so overwrite them.

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -48,7 +48,6 @@ config lime network
 #	option autoap_enabled 0				# Requires lime-ap-watchping installed. If enabled AP SSID is changed to ERROR when network issues
 #	option autoap_hosts "8.8.8.8 141.1.1.1"		# Requires lime-ap-watchping installed. Hosts used to check if the network is working fine
 
-
 ### WiFi general options
 
 config lime wifi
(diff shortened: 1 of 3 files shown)


p4u:
I exeprienced the same _problem_ several times, i.e I just want to set the channel for a specific radio but then I've to copy all the existing required options.

IMO your approach is much better and confortable for the user. So I agree on this modification.

Lets give @G10h4ck a couple of days to check it.

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -154,4 +154,12 @@ function utils.is_installed(pkg)
 	return utils.file_exists('/usr/lib/opkg/info/'..pkg..'.control')
 end
 
+function tableConcat(t1,t2)
+	local _,i
+	for _,i in ipairs(t2) do
+		table.insert(t1,i)
```
G10h4ck: This actually discard the key, so it has a concatenation semantic only in case the two tables are array, otherwise i believe it would degenerate to unexpected behaviors.
It is probably better to rename the function to arrayConcat ?

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
```diff
@@ -86,7 +86,7 @@ function wireless.configure()
 
 		if specRadio then
 			modes = specRadio["modes"]
-			options = specRadio
+			utils.tableConcat(options, specRadio)
```
G10h4ck: What happens if an option is present both in general and in specific section?

G10h4ck:
Beyond the code review this will probably cause big changes in the behavior of lime on unchanged configuration file, I believe this topic deserve a broader discussion and analysis before merge

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -154,4 +154,12 @@ function utils.is_installed(pkg)
 	return utils.file_exists('/usr/lib/opkg/info/'..pkg..'.control')
 end
 
+function tableConcat(t1,t2)
```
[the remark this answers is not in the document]
aparcar: just forgot to add it, done.

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -154,4 +154,12 @@ function utils.is_installed(pkg)
...
+	local _,i
```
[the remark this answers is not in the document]
aparcar: afaik unnecessary, I just copied took the function from qmpinfo

packages/lime-system/files/usr/lib/lua/lime/utils.lua
aparcar: renamed

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
aparcar: as it's added at the end the "default" options are overwritten with specific ones.

aparcar:
> Beyond the code review this will probably cause big changes in the behavior of lime on unchanged configuration file, I believe this topic deserve a broader discussion and analysis before merge

As  @p4u [mentioned above](https://github.com/libremesh/lime-packages/pull/320#issuecomment-364053985) this would imo improve the user experience: If the defaults are to use an encrypted mesh and a change of the ssid wouldn't result in a open AP. 
The new logic is, *if not defined, stick to default* instead of the current approach of *if individual, redefine **everything** yourself again*.

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
G10h4ck: Better add a documentation comment there explaining this as this behavior is implicit

aparcar: Apart from the missing documentation you're okay with this PR?

G10h4ck: Yeah for me it is ok.
I am wondering if similar behavior is something one would have on specific network interfaces too, I was considering this also because of homogeneity between the two specific behaviors

aparcar:
@G10h4ck I updated the `lime-example` file. It couldn't harm to clean it a bit and format it to our precious text width of 80 - but that may be another PR.

Please merge to continue with https://github.com/libremesh/lime-packages/pull/321

aparcar:
this patch can actually be a problem in case we use the client mode: the openwrt default value of 'auto' is overwritten to a lime-default channel. An optional if statement could handle that case for clients, that if no radio specific channel is set, it defaults to 'auto' ignoring the lime-defaults.

p4u:
> this patch can actually be a problem in case we use the client mode: the openwrt default value of 'auto' is overwritten to a lime-default channel. An optional if statement could handle that case for clients, that if no radio specific channel is set, it defaults to 'auto' ignoring the lime-defaults.

That's right, 'auto' should also be an option as a valid channel.

aparcar:
> That's right, 'auto' should also be an option as a valid channel.

https://github.com/libremesh/lime-packages/pull/320/files#diff-192457068239fe711325c38fafa5f441R107

p4u:
ping @G10h4ck is it fine for you? Can it be merged?

aparcar:
@G10h4ck ping

packages/lime-docs/files/lime-example
```diff
@@ -78,35 +77,46 @@ config lime wifi
 
 # The following interface specific options have to be included in /etc/config/lime, not in /etc/config/lime-defaults
 
-### WiFi interface specific options ( override general option )
+### WiFi interface specific options ( override defaults options )
+
+## use radio0 only for mesh
+config wifi radio0
+    list modes 'ieee89211s'
```
G10h4ck: Please use tab for indentation not spaces

packages/lime-docs/files/lime-example
```diff
@@ -78,35 +77,46 @@ config lime wifi
...
+
+## change ssid for radio1
+config wifi radio1
+    option ap_ssid 'Special'
```
G10h4ck: Please use tab for indentation not spaces

packages/lime-docs/files/lime-example
```diff
@@ -78,35 +77,46 @@ config lime wifi
...
-#	list modes 'adhoc'
-#	option channel_2ghz '1'
-#	option channel_5ghz '48'
-#	option adhoc_ssid 'LiMe'			# Parametrizable with %M, %H
-#	option adhoc_bssid 'ca:fe:00:c0:ff:ee'
-#	option adhoc_mcast_rate_2ghz '24000'
-#	option adhoc_mcast_rate_5ghz '6000'
+## diable lime-config for radio2
+config wifi radio2
+    option modes 'manual' # If you use manual protocol you must not specify other protocol, or your configuration will be broken!
```
G10h4ck: Please use tab for indentation not spaces

aparcar:
@G10h4ck ping

G10h4ck:
Thanks you!

PR #321, opened 2018-02-07
lime-smart-wifi: add package, fix #306

### LINKED ISSUE #306: separate Mesh and AP radio based on frequencies

aparcar:
fix separate Mesh and AP radio based on frequencies #306

1 radio with dual band is not supported due to missing device for
  testing, this will be added.

automatically swaps channels if multible radios on same freq exist

tested on tp link wdr 4300

also introduce new function to utils to check if an element is inside an
array

reopend due to git messup

### DIFF
--- a/packages/lime-smart-wifi/Makefile
+++ b/packages/lime-smart-wifi/Makefile
@@ -1,0 +1,34 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lime-smart-wifi
+PKG_VERSION:=0.2
+PKG_RELEASE=2
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+ SECTION:=net
+ CATEGORY:=LiMe
+ TITLE:=Configure wifi mesh and ap hardware based
+ MAINTAINER:=Paul Spooren <paul@spooren.de>
+ URL:=http://libremesh.org
+ DEPENDS:=+lime-system
+endef
+
+define Package/$(PKG_NAME)/description
+ Automatically configure wifi mesh and ap based on existing radios
+endef
+
+define Build/Prepare
+endef
+
+define Build/Compile
+endef
+
+define Package/$(PKG_NAME)/install
+	$(INSTALL_DIR) $(1)/usr/bin
+	$(CP) ./src/lime-smart-wifi.lua $(1)/usr/bin/lime-smart-wifi
+	$(CP) ./files/* $(1)/
+endef
+
+$(eval $(call BuildPackage,$(PKG_NAME)))
--- a/packages/lime-smart-wifi/files/etc/uci-defaults/90_lime-smart-wifi
+++ b/packages/lime-smart-wifi/files/etc/uci-defaults/90_lime-smart-wifi
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+uci -q get lime-defaults.smart_wifi || \
+    uci set lime-defaults.smart_wifi='smart_wifi'
+uci -q get lime-defaults.smart_wifi.ap_2ghz || \
+    uci set lime-defaults.smart_wifi.ap_2ghz=1
+uci -q get lime-defaults.smart_wifi.ap_5ghz || \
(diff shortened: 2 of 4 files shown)


aparcar:
Depends on https://github.com/libremesh/lime-packages/pull/320

aparcar:
Just tested the app with a xiami miwifi router and it works as intended. 
In the default we could enable 2.4Ghz mesh per default so the devices stay compatible to single 2.4Ghz band devices.

@p4u @G10h4ck please merge this patch and the #320

p4u:
I'd suggest a change on the configuration section. Instead of adding the smart-wifi options to the lime.wifi section (which is used by lime-system) I think it's much better to separate it in another section. For instance

```
config lime smart-wifi
 list channels_2ghz 11
 list channels 2ghz 6
 list channels 5ghz 36
 list channels 5ghz 48
 option mesh_5ghz 1
 option ap_5ghz 0
 option mesh_2ghz 0
 option ap_2ghz 1
 option all_to_mesh 0
 option all_to_ap 0
```

However this options cannot be added in compile time to /etc/config/lime-defaults because this file is owned by lime-system package. Then a uci-defaults script is required to check if the option exist and set it if not. Example for an entry

> uci -q get lime-defaults.smart-wifi.mesh_2ghz || uci set lime-defaults.smart-wifi.mesh_2ghz=0

It is better to have all default values in config files instead of hardcoding it into the code (I'm not saying not to hardcode but they must also be visible in the config file).

aparcar:
@p4u i really dislike the idea of using UCI scripts. All settings should be configurable in the defaults package. I don't want the user to understand the uci-defaults system plus the UCI command syntax, don't you think?

p4u:
> @p4u i really dislike the idea of using UCI scripts. All settings should be configurable in the defaults package. I don't want the user to understand the uci-defaults system plus the UCI command syntax, don't you think?

I don't think you understand what I'm proposing. User must NOT deal with uci-defaults, but if an option is not set (right now your commit does not include any option) a uci-default script would create it on first boot. 

The user still can have its own /etc/config/lime-defaults file with the options, as usual.

p4u:
The other approach to not make this options invisible is to add them in lime-defaults but as I said before, this config file is owned by another package (lime-system). Adding options of one package to another is not a good aproach IMO.

p4u:
> The other approach to not make this options invisible is to add them in lime-defaults but as I said before, this config file is owned by another package (lime-system). Adding options of one package to another is not a good aproach IMO.

Or let's just do it... my concern is to make the lime-defaults file too big if we start adding all options from all libremesh related packages

aparcar:
Good I'll add defaults via UCI in it's own section. That simplifies the removal as well

aparcar:
> the default we could enable 2.4Ghz mesh per default so the devices stay compatible to single 2.4Ghz band devices

@p4u

p4u:
ok

aparcar:
@p4u all done hopefully

aparcar:
@p4u ping

ilario:
Congrats for the amazing package :D 
Can we have a commented configuration example in packages/lime-docs/files/lime-example?

aparcar:
sure, will create that later

PR #324, opened 2018-02-11
wireless: default distance 5:1000m and 2.4:100m

aparcar:
following the idea here:
https://github.com/libremesh/lime-packages/pull/319#issuecomment-363400722


ilario:
We will have to split this also in the web interface.
And stress a lot the users on the fact that they will have to increase the distance_5ghz for links above 1 km.

@G10h4ck @nicoechaniz @altergui @rogerpueyo @p4u @nicopace @egon0 @FreifunkUFO @dangowrt 
do you like the 1 km default distance for 5 GHz and 100 m for 2.4 GHz?

p4u:
I like the idea to be able to specify it depending on the WiFi band.

About the default distance, I think 100 and 1KM are quite realistic.

PR #329, opened 2018-02-14
wireless: fix ignored freqs + default htmode

aparcar:
fix error due to local ignored freqs if 2.4ghz


aparcar:
@p4u please check this PR the `lime-config` is broken for now

PR #333, opened 2018-02-19
lime-proto-bmx6,7: fix maxRate for wifi

aparcar:
max rate is configured in Bits not KBits. 54000 would result in
a estimated max connection of 54KBit while it's really 54Mbit.

### DIFF
--- a/packages/lime-proto-bmx6/src/bmx6.lua
+++ b/packages/lime-proto-bmx6/src/bmx6.lua
@@ -198,7 +198,7 @@ function bmx6.setup_interface(ifname, args)
 
 	-- BEGIN [Workaround issue 40]
 	if ifname:match("^wlan%d+") then
-		uci:set(bmx6.f, owrtInterfaceName, "rateMax", "54000")
+		uci:set(bmx6.f, owrtInterfaceName, "rateMax", "54000000")
 	end
 	--- END [Workaround issue 40]
 
--- a/packages/lime-proto-bmx7/src/bmx7.lua
+++ b/packages/lime-proto-bmx7/src/bmx7.lua
@@ -196,7 +196,7 @@ function bmx7.setup_interface(ifname, args)
 
 	-- BEGIN [Workaround issue 40]
 	if ifname:match("^wlan%d+") then
-		uci:set(bmx7.f, owrtInterfaceName, "rateMax", "54000")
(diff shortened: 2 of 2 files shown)


aparcar:
I'm not fully certain about the impact of this miss configuration but if I'm not mistaken then all mesh/ad hoc connections with a better connection then 56Kbit where treated the same by BMX, not distinguishing between a 1Mbit or 50Mbit connection as best path.

G10h4ck:
Can you point to the documentation specifying the unis is Kbit/s and not bit/s?

Cheers!

aparcar:
https://github.com/bmx-routing/bmx7
>txRate: Calculated transmit rate in bps for the link.

G10h4ck:
https://github.com/bmx-routing/bmx7#dynamic-reconfiguration

dangowrt:
@G10h4ck the example above denotes rates in kbit/s, ie.
```
bmx7 -c dev=-eth1 dev=eth2 /rateMax=100000
```
to set the rate to 100MBit/s.
Either this documentation is wrong or @aparcar assumption that the rate is denoted in Bit/s rather than kBit/s is wrong.

dangowrt:
Taking the iwinfo plugin as an example, I reckon having an ethtool plugin to get the actual speed and status of an Ethernet link shouldn't be hard at all.
Also note that on 802.11a and 802.11g the highest **physical** rate is 54MBit/s -- the typical payload transmission rate when using this standard is about half of that around 25MBit/s and **varies** even when using the same phyrate depending on re-sends necessary due to collisions. That's why we got minstrel to estimate the transmission rate because it cannot be predicted reliably otherwise...

p4u:
By the way, just to point out that the decision of coding and merging this PR was not arbitrary,  there was a previous discussion on this topic in Riot bmx-routing channel with @axn 

> Regarding rxRate in general I can't find anything wrong. With perfect links and default dev=... /rateMax setting it shows rxRate=56000K which reflects 56M. A reasonable assumption unless more detailed physical capabilities are known for a given interface
> If you change /rxRate=100000000 then also 'bmx7 -c links' reflects this

Also this issue on bmx7 github https://github.com/bmx-routing/bmx7/issues/11

It looks like this example on the bmx7 documentation was wrong.

Issue #257, opened 2017-11-15
Commit access for TranslateWiki in LibreMesh repos


patogit:
(Continuing the conversation from PR #254)

@nicopace and @Nikerabbit (and anyone else who's interested), let's see if we can agree on how to arrange commit access for translation files from TranslateWiki. I imagine that whatever we work out here can apply to the LiMeApp too ( @gmarcos87 ), and maybe other repos if they're ready to translate before January.

How about if commits go to a protected branch for the first two months, and then we revisit the idea of TranslateWiki having direct push commit access?

In other repos, @TranslateWiki makes about one commit per month (usually on a Monday or Thursday). So after two commits maybe there will be more confidence to allow direct push... or maybe the LibreMesh development flow really depends on translations going through protected branches just like all the other code.

@Nikerabbit, would that work for you? Do you have other ideas?

nicopace:
Noone in the LibreMesh community has direct commit access (or shouldn't have) to protect and collectivize the knowledge, and to ensure that what gets merged works by doing integration tests.

The Pull Requests are reviewed within a week tops, so doesn't seem to be conflicting with the times of translatewiki.

ilario:
For me no problem to give direct commit access to a TranslateWiki member. 
Our mechanism of no one has direct commit access is for avoiding people to push bad code or useless branches, but this is not going to happen for contributions from TranslateWiki, just translations will come from them.

Nikerabbit:
> How about if commits go to a protected branch for the first two months, and then we revisit the idea of TranslateWiki having direct push commit access?

I am not sure what do you mean. We currently do not have support for pull requests and I'm not fond of doing pull requests manually.

> In other repos, @translatewiki makes about one commit per month (usually on a Monday or Thursday).

I try to push translations every Monday and Thursday. Commits in each repository does not happen every time if there are no changes to translations.

Translation updates are generally very safe, except for mobile apps where the app doesn't build if translations have errors with variables or mark-up.

patogit:
@Nikerabbit, what would be required to add support for pull requests? Where is the code that automates the process you use now to commit to GitHub repositories?

LiMeApp is a mobile app.

The software in lime-packages, LuCI, is the web interface for router firmware. Do you (or anyone else on this thread) know how OpenWRT / LEDE derivatives handle errors with variables or markup in translation files?

Nikerabbit:
[...]

PR #337, opened 2018-02-21
add qqq language code for msg documentation

patogit:
this language code is for documentation messages for translators.


patogit:
If I need to `git rebase` or something like that, please explain, as I'm learning how to use git.

patogit:
cancel this pull request. Turns out it's not a good idea, see comment at https://github.com/libremesh/lime-packages/issues/257#issuecomment-367413975

PR #347, opened 2018-04-20
lime-hwd-openwrt-wan: fix script if missing wan

aparcar:
currently a missing wan interfaces screws the config, the following
string will be inserted:

    WARNING: Variable 'wan' does not exist or is not an array/object

this happens for instance on a raspberry pi with only one eth port.

If no wan is detected, just leave it unconfigured.


aparcar:
@G10h4ck okay?

aparcar:
Thanks

PR #349, opened 2018-04-27
[RFC]utils: introduce %m in IPs to use macid

aparcar:
macid is like netid a md5sum hash of mac. The idea is to have static IP
addresses for devices which sustain after a reset or reflash, while
keeping collision chance as lower as using %M aka the real mac.
This is due to the fact that when using similar devices, they tend to
have nearly equal mac addresses.

A possible network could look like this: 10.%m1.%m2.%m3/25 which would
result in a collision chance of `n/(2 * 256^2)` where `n` is the number of
used devices.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/utils.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/utils.lua
@@ -44,11 +44,15 @@ end
 
 function utils.applyMacTemplate16(template, mac)
 	for i=1,6,1 do template = template:gsub("%%M"..i, mac[i]) end
+	local macid = utils.get_id(mac)
+	for i=1,6,1 do template = template:gsub("%%m"..i, macid[i]) end
 	return template
 end
 
 function utils.applyMacTemplate10(template, mac)
 	for i=1,6,1 do template = template:gsub("%%M"..i, tonumber(mac[i], 16)) end
+	local macid = utils.get_id(mac)
+	for i=1,6,1 do template = template:gsub("%%m"..i, tonumber(macid[i], 16)) end
 	return template
 end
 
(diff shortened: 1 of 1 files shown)


gmarcos87:
For a test in a real environment I isolated the functions and ran it with all the macs of our network in quintanalibre using 10.5.%m1.%m2/16. The result is as follows:
```
➜  $ for mac in `cat all_br-lan_macs`; do lua /tmp/macToIp.lua 10.5.%m1.%m2/16 $mac | uniq --count; done
      1 10.5.183.158/16
      1 10.5.208.55/16
      1 10.5.166.97/16
      1 10.5.2.97/16
      1 10.5.249.240/16
      1 10.5.249.201/16
      1 10.5.223.216/16
      1 10.5.76.67/16
      1 10.5.72.71/16
      1 10.5.127.233/16
      1 10.5.53.14/16
      1 10.5.128.124/16
      1 10.5.229.185/16
      1 10.5.208.77/16
      1 10.5.164.180/16
      1 10.5.217.31/16
      1 10.5.37.96/16
      1 10.5.124.102/16
      1 10.5.221.123/16
      1 10.5.238.61/16
      1 10.5.72.202/16
      1 10.5.235.160/16
      1 10.5.73.231/16
      1 10.5.7.63/16
      1 10.5.225.249/16
      1 10.5.214.54/16
      1 10.5.171.26/16
      1 10.5.173.92/16
      1 10.5.153.55/16
      1 10.5.227.147/16
      1 10.5.13.186/16
      1 10.5.156.47/16
      1 10.5.48.189/16
      1 10.5.12.144/16
      1 10.5.212.221/16
      1 10.5.22.128/16
      1 10.5.199.177/16
      1 10.5.174.241/16
      1 10.5.162.126/16
      1 10.5.9.44/16
      1 10.5.125.114/16
      1 10.5.142.143/16
      1 10.5.155.143/16
      1 10.5.82.146/16
      1 10.5.187.183/16
      1 10.5.251.173/16
      1 10.5.162.21/16
      1 10.5.159.69/16
      1 10.5.52.35/16
      1 10.5.203.232/16
      1 10.5.40.184/16
      1 10.5.208.115/16
      1 10.5.79.190/16
```
There is no collision. Can you make the function respect not so wide ranges? like /21 for example. How would you calculate the probability of collision in that case?

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -57,18 +61,25 @@ function utils.applyHostnameTemplate(template)
 	return template:gsub("%%H", system.get_hostname())
 end
 
+function utils.get_id(string)
+	local id = {}
+	local fd = io.popen('echo "' .. string .. '" | md5sum')
+	if fd then
+		local md5 = fd:read("*a")
+		id[1] = md5:match("^(..)")
```
G10h4ck: It's a bit perverse to use match instead of for + strsub i see this thing was already present in our code but it's not good motivation to keep doing it wrong (it could start smell like chicken at some point )... in this cases, morover why not exposing the full md5 instead of only 6 couples?

aparcar: @G10h4ck less chicken taste now? https://github.com/libremesh/lime-packages/pull/349/files#diff-bf0ed9c5850e447bab27e71c3db02c03R70

aparcar:
@G10h4ck changed again to use `j` instead of `i` multiplication to save precious CPU cycles.

PR #356, opened 2018-05-14
[RFC] proto/lan: handle bridged primary_interface fix #355

### LINKED ISSUE #355: bridging all eth interfaces results in random MAC
TL;DR: If all interfaces are bridged, should `primary_interface` become `br-lan` for static MAC addresses?

Using the IP placeholder `%m` to use a hash based of the *static* MAC I realized that the MAC address changes if eth0 and eth1 are bridged together. That's the automatic setup of `lime-system` if `lime-hwd-openwrt-wan` isn't installed.

The used MAC is based on the option `primary_interface` (default eth0) which becomes random as the bridge `br-lan` receives the "real" MAC.

aparcar:
If eth0 is bridged on br-lan the bridge receive the devices MAC address
and the real eth0 (seems to) get a random generated MAC address.

To avoid random MAC addresses the primary interface of lime is setup to
br-lan in these cases.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua
@@ -55,6 +55,9 @@ function lan.setup_interface(ifname, args)
 			table.insert(bridgedIfs, iface)
 		end
 	end
+    if ifname == config.get("network", "primary_interface") then
+        config.set("network", "primary_interface", "br-lan")
+    end
 	table.insert(bridgedIfs, ifname)
 	uci:set("network", "lan", "ifname", bridgedIfs)
 	uci:save("network")


aparcar:
@dangowrt it's a Buffalo WDR-1166DHP, and I think it's fairly well supported. The *odd* behavior appears once eth0 and eth1 are bridged (no eth separation between lan and wan).

Once bridged, the `br-lan` however uses the MAC address printed on the device - so I guess no hardware error.

dangowrt:
Generally, the bridge inherits its MAC from the first interface added -- I reckon that'd be eth0 or eth0.1 in your case. Yet, no physical ethernet interface should **ever** come up with a random MAC whatsoever, this always hints that something has gone wrong (exception: very crappy devices designed for client-mode)

dangowrt:
@aparcar I had a look at the board support for this Buffalo device and it seems like the part relevant for the ethernet MAC address is missing in the device-tree and only handled in userspace (which is an option we only have for legacy reasons). I'm not sure which device-tree file actually belongs to the board you are talking about, is it `WSR-1166.dts`?
The problem seems to the that the device MAC is not stored at a constant offset inside an MTD partition but rather as a variable inside the `board_data` partition -- and unfortunately we lack the kernel support for parsing that env partition, currently only constant offsets inside partitions are managed by the kernel. This is then worked around by setting the address in userspace in `/etc/board.d/02_network` which generates `/etc/board.json`.

Once again: PARSING `/etc/config/network` FOR ANY PURPOSE WILL ALWAYS LEAD TO PROBLEMS AND HAS ONLY WORKED COINCIDENTALLY.
We should use `/etc/board.json` instead, it got everything we need to generate the network configuration for the device.

aparcar:
> This seems quite dangerous to me, should not alter programmatically that configuration...

Fair enough

> We should use /etc/board.json instead, it got everything we need to generate the network configuration for the device.

Are you talking about [the loop iterating over uci options](https://github.com/libremesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua#L53)?

Parsing JSON with lua introduces a new dependence (which is fine for me) or re implementing stuff with shell scripts (using jshn). @G10h4ck @dangowrt you choose - I'm happy to implement that

dangowrt:
Parsing JSON in Lua shouldn't be too tricky, I believe we already drag in a library for that at some other point.
Generally, as a rule of thumb: If some code **generates** UCI this should **never** also read UCI, but **always** take that information from a level lower, where it originally came from (see `config_generate` and such).

aparcar:
@dangowrt I'm a bit confused as in `board.json` the same MAC is used for wan and lan
```

        "network": {                                          
                "lan": {                                  
                        "ifname": "eth0.1",                
                        "protocol": "static",                 
                        "macaddr": "74:03:BD:31:0E:80"    
                },                                         
                "wan": {                                  
                        "ifname": "eth0.2",           
                        "protocol": "dhcp",                   
                        "macaddr": "74:03:BD:31:0E:80"    
                }                                     
        } 
```

PR #357, opened 2018-05-14
change .po file encoding to UTF-8

patogit:
In order to work with TranslateWiki, @Nikerabbit asked to change the .po files from ISO-8859 to UTF-8. Two files (po/limeweb-ui.pot and po/qqq/limeweb-ui.po) were already in UTF-8. In this commit, I changed the remaining file po/es/limeweb-ui.po to UTF-8.

@nicoechaniz thinks it will be ok with the routers, but says to ask @G10h4ck .
@G10h4ck would be surprised if UTF-8 works on the routers.

Opinions? Solutions? Ways to test without bricking routers?

### DIFF
--- a/packages/lime-webui/po/es/lime-webui.po
+++ b/packages/lime-webui/po/es/lime-webui.po
@@ -35,7 +35,7 @@ msgid "Distance"
 msgstr "Distancia"
 
 msgid "Max distance for the links in meters (big impact on performance)"
-msgstr "Distancia m√xima de los enlaces WiFi en metros (impacto importante)"
+msgstr "Distancia m√°xima de los enlaces WiFi en metros (impacto importante)"
 
 msgid "Default channel used for 2.4 GHz radios"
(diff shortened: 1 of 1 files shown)


G10h4ck:
I think this should be easy to test without briking the router, at worst the app which uses the .po will display strange characters, so please test it and report back ;)

patogit:
ok. I've used the cooker once, and I know there are ways to include local files in what gets cooked. Is this the proper way to do it? https://github.com/libremesh/lime-sdk/blob/master/README.md#for-a-casual-cooking-on-a-existing-feed-repository Following those instructions, and just changing the one file that I'm interested in?

patogit:
note that even though i left the "keep settings" box checked, various things were reset to default:
- SSIDs of WiFi networks
- what SSIDs are enabled and what are enabled
- hostname

what settings were preserved:
- country code of SSIDs
- root password
- timezone

Maybe I can report this somewhere else?

PR #358, opened 2018-05-16
changed priorize to prioritize in two strings in translation template file

patogit:


patogit:
also changed "Attended Sysupgrade" to "Attended System Upgrade" since that seems easier to translate. I know that the file names are sysupgrade.bin or factory.bin, but in this case it seems to make sense to actually say "system upgrade".

PR #362, opened 2018-05-24
[RFC]travis: build packages and upload them

aparcar:
current upload to easy.planetexpress.cc - easily changeable.

I think snapshot packages are more important as snapshot images: The images can easily be updated if a snapshot repository exists. Reflashing (and producing) new images is more complicated and requires much more computation time.

Maybe it's possible to unite the image creation for PRs and the snapshot building on updates of dev branch. What do you think?


.travis.yml
```diff
@@ -2,36 +2,65 @@ sudo: false
...
 notifications:
   email: false
 language: c
+cache:
+  directories:
+    - $TRAVIS_BUILD_DIR/sdk
 install:
 - cd $TRAVIS_BUILD_DIR
-- git clone https://github.com/libremesh/lime-sdk.git sdk || true
+- git clone https://github.com/aparcar/lime-sdk.git sdk || true
```
nicopace: this should be libremesh

.travis.yml
```diff
@@ -2,36 +2,65 @@ sudo: false
...
 - sed 's|https://github.com/libremesh/lime-packages.git.*|'$TRAVIS_BUILD_DIR'|g' feeds.conf.default > feeds.conf.default.local
-- ./cooker -f -d x86/generic
+- ./cooker -f --sdk $TARGET
 script:
 - cd $TRAVIS_BUILD_DIR/sdk
-- export J=$(($(nproc)+1))
-- ./cooker -b x86/generic
-- travis_wait ./cooker --flavor=lime_default -c x86/generic --profile=Generic
-- travis_wait ./cooker --flavor=lime_mini -c x86/generic --profile=Generic
-- travis_wait ./cooker --flavor=lime_zero -c x86/generic --profile=Generic
```
nicopace: if we remove the building images thing, we need to add a a messages on our bot that shows anyone how to build the images based on the packages.

.travis.yml
aparcar: Were needed as I used a patched cooker

.travis.yml
aparcar: I'd suggest we add a link to the cooker + chef

aparcar: > if the case is that noone is using it, we need to review the case for it... probably we could automate more in order for it to do something useful, for example to do testing

I think devs likely have some kind of setup where they run a VM/device. Once a PR is created I'd like to fetch the resulting package and install it in my setup and see if it works. For me that seems more convenient then downloading the image. 

However if that's a required and used feature, the approach could be changed to build 1 x86/64 image + packages.

Also we should distinguish between snapshot and release PRs.

aparcar:
> if the case is that noone is using it, we need to review the case for it... probably we could automate more in order for it to do something useful, for example to do testing

I think devs likely have some kind of setup where they run a VM/device. Once a PR is created I'd like to fetch the resulting package and install it in my setup and see if it works. For me that seems more convenient then downloading the image.

However if that's a required and used feature, the approach could be changed to build 1 x86/64 image + packages.

Also we should distinguish between snapshot and release PRs.

aparcar:
This was merged to master

PR #364, opened 2018-05-25
lime-defaults: configure bmx7 if installed

aparcar:


G10h4ck:
please set make default vlan explicit in the config file, take care that id 17 is reserved already for babeld

aparcar:
@dangowrt @G10h4ck how should we solve this? The lime-proto-bmx7 defaults to VID 13 just like bm6.
We can choose VID 18 or run both bmx version on the same VID...

https://github.com/aparcar/lime-packages/blob/0ca6c8242d34fce16de21e38f6a8293fc3f7e4e6/packages/lime-proto-bmx7/src/bmx7.lua#L193

aparcar:
Changed in `lime-proto-bmx7` as well to stay compatible.
We actually break things here, but well it's snapshots and people have their own `lime-defaults` anyway

aparcar:
@G10h4ck nice RTT, thanks

PR #370, opened 2018-06-01
treewide: set PKGARCH:=all

aparcar:
Set PKGARCH to all where possible (only Lua/Bah code). This results in images install able on all devices.

This will dramatically reduce the time to compile all packages.


aparcar:
Please merge this, it's tested and all packages still compile.

PR #374, opened 2018-06-04
add option babeld_over_librenet6

aparcar:
make babeld run on librenet6 tinc interface

### DIFF
--- a/packages/lime-proto-babeld/src/babeld.lua
+++ b/packages/lime-proto-babeld/src/babeld.lua
@@ -34,6 +34,12 @@ function babeld.configure(args)
 
 	local uci = libuci:cursor()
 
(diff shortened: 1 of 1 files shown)


G10h4ck:
seems that babel+uci is pretty sensitive to order of config sections, please move this before redistribute rules, moreover i would specify that the interface is tunnel if possible (check if the script which convert from the UCI file to babel config support that option)

aparcar:
Moved it up before the redistribute statements. Also added `type tunnel`. The docs mention if not added, it's detected automatically, maybe we shoudl trust on that?

G10h4ck:
I have tested the autodetection and it wasn't reliable for vlan interfaces, so better to specify when we know for sure

PR #379, opened 2018-06-05
replace dnsmasq-dhcpv6 with odhcpd-ipv6only

aparcar:
eventually solving a whole stack of issues

### DIFF
--- a/packages/batman-adv-auto-gw-mode/Makefile
+++ b/packages/batman-adv-auto-gw-mode/Makefile
@@ -19,7 +19,7 @@ define Package/$(PKG_NAME)
   TITLE:=Set batman-adv gw_mode by internet availability
   MAINTAINER:=Gui Iribarren <gui@altermundi.net>
   URL:=http://libremesh.org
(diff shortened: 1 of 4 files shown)


G10h4ck:
have you tested if everything keeps working as expected?

nicopace:
As this are critical features, we need to test on devices and on networks.
Would say @G10h4ck @nicoechaniz and me at least to test on devices before merge.
Still a WIP although, so just let us know @aparcar

nicopace:
A lot of changes got into this @aparcar ... 
hope you can clean up.

aparcar:
I just messed up the PR with some rebasing. Now it's clean again.
On the snapshot server I created some images running with the modified version, works here for me
https://snapshots.libremesh.org/staging/aparcar/targets/

aparcar:
Generally I'd suggest to remove the distributed hosts thing and replace it with the to be created `mesh-dns`.

gmarcos87:
Now that alfred was replaced by shared-state we should change that too. Also pirania depends on dnsmasq to redirect traffic.
I would like to know a little more about the advantages of this migration, currently the behavior of dnsmasq is not being conflictive.

PR #380, opened 2018-06-06
[WIP] let travis handle pull request

aparcar:
if push to dev branch: build packages and upload
if pull request: build slowly and show full log to see errors

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,8 @@ install:
   - /$TRAVIS_BUILD_DIR/.travis_do.sh setup
 script:
   - /$TRAVIS_BUILD_DIR/.travis_do.sh build
+before_cache:
+  - rm -rf "$HOME/sdk/$SDK_TARGET/sdk/key-build"
 after_success:
   - /$TRAVIS_BUILD_DIR/.travis_do.sh upload
 branches:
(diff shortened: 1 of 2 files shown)


aparcar:
@nicopace this works but the log output is simply to long...

nicopace:
I would sed out the outputs that are not relevant.
Like all the compilation that are not errors are not needed.

aparcar:
The openwrt/packages travis config does this quite nice by just compiling packages which actually where changed. Reducing the overall compiling output to a minimum. We could use this approach as well.

@nicopace However, I'm a bit unsure if the compiling of PRs makes sense after all. They're not uploaded and as LibreMesh is 99% Lua, there won't be any errors anyway, except you break a Makefile

nicopace:
The important part of doing travis builds is:
1. we ensure we don't break the repo (has happened before)
2. we have a ready-made image to download and test the functionality without having to have the environment. Much easier for others to contribute.

aparcar:
No longer needed due to https://github.com/libremesh/lime-packages/commit/29a7bf743529e5ddb97c90602798b3ef2b9f2f99

aparcar:
Generally, PR building is somewhat useless  as we can't upload PR packages, it would allow anyone to access our servers. Instead if you're a LiMe dev, push to a different (unprotected) branch and within some 20 minutes your packages appear [here](https://snapshots.libremesh.org/branches/)

PR #382, opened 2018-06-17
utils: add function tableAppend

aparcar:
I don't know what lead to the arrayContact function but it can't be used
in that case: `options` are loaded via `config.get_all()` and result in
an dict like table. Specific options are also a dict and can't be added
as an array.

Also both function return the value rather then manipulating the
incomming table. Therebey wireless.lua is adjusted.

Also cleans some whitspace errors

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/utils.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/utils.lua
@@ -4,11 +4,10 @@ utils = {}
 
 local config = require("lime.config")
 
-
 function utils.split(string, sep)
-    local ret = {}
-    for token in string.gmatch(string, "[^"..sep.."]+") do table.insert(ret, token) end
-    return ret
+	local ret = {}
+	for token in string.gmatch(string, "[^"..sep.."]+") do table.insert(ret, token) end
+	return ret
 end
 
 function utils.stringStarts(string, start)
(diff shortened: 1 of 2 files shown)


G10h4ck:
hahahahahah dejavu, I remember to say similar things when you made the past tableConcat PR that then end up beign named arrayAppend, I would rename the function into tableMeld because in case t2 has some key which overlap t1 t1 get overwritten and please say tat on the documentation too

aparcar:
@G10h4ck done

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -176,10 +175,18 @@ function utils.has_value(tab, val)
...
+--! contact array a2 to the end of array a1
+function utils.arrayConcat(a1,a2)
+	for _,i in ipairs(a2) do
+		table.insert(a1,i)
+	end
+	return a1
+end
+
+--! melt table t1 into t2, if keys exists in both tables use value of t2
+function utils.tableMelt(1,t2)
```
G10h4ck: This seems a fatal type 1 instead of t1 ;)

aparcar: man you're picky today ;)

PR #393, opened 2018-06-29
lime-proto-bmx7: use auto rate

aparcar:
bmx7 estimates the maxRate of wifi interfaces by iteself, thereby don't
set a static value of 54M in times of AC mesh connections


nicopace:
Don't have a way to test it.
Do you have AC equipment to test that this works?

aparcar:
BMX7 is able to determine maxRates on its own via `iwinfo` plugin, if that's not available it uses the hardcoded value of 54M, so no need to make this redundant in our code.

aparcar:
@G10h4ck I think we fixed by Sr. Dango and Axel, some time ago https://github.com/bmx-routing/bmx7/commit/23610b979d571a59d1914c1aa4ae06cfd5de9750

If not I'd ask @axn to fix this in BMX7 rather then implementing a workaround here...

G10h4ck:
We done both things I thinks, ask to @axn and implement the workaround in the meanwhile

PR #394, opened 2018-06-29
Update README.md

aparcar:


nicopace:
Why would you change the README for the develop and master branch?
I don't see the point of having divergent branches between these two.
develop should be what master will end up being once it stabilizes, right?

aparcar:
@nicopace develop is now master branch, which did not exist before. Stable releases get their own branches, just like OpenWrt does. To keep the code available I'd suggest to mark the changes in the README and not delete it all.

nicopace:
> To keep the code available I'd suggest to mark the changes in the README and not delete it all.

What do you mean by this?

G10h4ck:
It's the develop -> master move completed also in other components like lime-sdk? In that case i would merge this

aparcar:
Yes I'd say it was done a long time ago

PR #398, opened 2018-08-07
lime-system: use lime-profile if available

aparcar:
Network profiles are somewhat popular, but cause problems when they're
used as packages, as lime-system and the packaged profile both supply
the file `/etc/config/lime-defaults`. A solution is to introduce
`lime-profile` which is then used instead of lime-defaults.

This has the advantage that new releases with different defaults don't
break compatiblity.


nicopace:
@G10h4ck and @nicoechaniz should check this

PR #399, opened 2018-08-12
add lime-librenet6 package

aparcar:

### DIFF
--- a/packages/lime-librenet6/files/etc/tinc/librenet6/tinc-up
+++ b/packages/lime-librenet6/files/etc/tinc/librenet6/tinc-up
@@ -1,0 +1,2 @@
+#!/bin/sh
+ifconfig $INTERFACE up txqueuelen 1000
--- a/packages/lime-librenet6/files/etc/uci-defaults/tinc-setup
+++ b/packages/lime-librenet6/files/etc/uci-defaults/tinc-setup
@@ -1,0 +1,33 @@
+#!/bin/sh
+
+# tell tinc to use the librenet6 network
+uci -q show tinc.librenet6 || {
+    touch /etc/config/tinc
+    uci batch <<-EOF
+    set tinc.librenet6='tinc-net'
+    set tinc.librenet6.enabled='1'
+    commit tinc
+EOF
+}
+
+# disable ra
+uci set network.lan.accept_ra=0
+uci commit network
+
+# set IPv6 subnet of librenet6.librenet.org
+uci set lime.network.main_ipv6_address '2002:8aff:5905:%N1%N2::/64'
+
+# be sure to enable babeld
+uci get lime.network.protocols | grep "babeld" || {
+    uci add_list lime.network.protocols="babeld:17"
+}
+
+# commit modified lime config
+uci commit lime
+
+# apply all changes
+lime-config
+lime-apply
+
+# restart tinc to be sure it and babeld are running
+/etc/init.d/tinc restart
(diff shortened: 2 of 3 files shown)


packages/lime-librenet6/files/etc/tinc/librenet6/tinc-up
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+ifconfig $INTERFACE up txqueuelen 1000
```
G10h4ck: ifconfig is deprecated! Isn't ip installed by default on openwrt at least some minimal version from busybox ?

packages/lime-librenet6/files/etc/uci-defaults/tinc-setup
```diff
@@ -0,0 +1,33 @@
...
+    commit tinc
+EOF
+}
+
+# disable ra
+uci set network.lan.accept_ra=0
+uci commit network
+
+# set IPv6 subnet of librenet6.librenet.org
+uci set lime.network.main_ipv6_address '2002:8aff:5905:%N1%N2::/64'
```
G10h4ck: This seems to interfere with normal lime behavior or am I missing something?

packages/lime-librenet6/files/etc/uci-defaults/tinc-setup
```diff
@@ -0,0 +1,33 @@
...
+# be sure to enable babeld
+uci get lime.network.protocols | grep "babeld" || {
+    uci add_list lime.network.protocols="babeld:17"
+}
+
+# commit modified lime config
+uci commit lime
+
+# apply all changes
+lime-config
```
G10h4ck: This two line seems could interfere with normal lime behavior too

dangowrt:
On Tue, Aug 28, 2018 at 08:47:26AM -0700, G10h4ck wrote:
> G10h4ck requested changes on this pull request.
> 
> 
> 
> > @@ -0,0 +1,2 @@
> +#!/bin/sh
> +ifconfig $INTERFACE up txqueuelen 1000
> 
> ifconfig is deprecated! Isn't ip installed by default on openwrt at least some minimal version from busybox ?

Just tried
ip link set dev eth0 txqueuelen 1000
and realized that busybox' ip applet indeed doesn't support setting
txqueuelen. We should change that...

packages/lime-librenet6/files/etc/uci-defaults/tinc-setup
aparcar: Well you have to use an IP of that subnet if you want a working connection. To keep it simple for users I decided to add this

G10h4ck:
same comments apply to https://github.com/libremesh/lime-packages/pull/400 I wonder if creating a librenet6 lime-proto instead of an uci-default could integrate better with libremesh, being a lime-proto it could also be configurable from /etc/config/lime lime-app etc.

dangowrt:
> +ifconfig $INTERFACE up txqueuelen 1000

please use
```
ip link set $INTERFACE qlen 1000
```
instead (qlen is a synonym for txqueuelen).

ilario:
@aparcar @G10h4ck once the comments are addressed, can we merge this or there is something else?

gmarcos87:
Hi @ilario, I don't think this is suitable for merge. If I run this on our libremesh network it won't work. You should not change the ipv6 settings from lime but add new ones. I think the requests for changes made by gio are quite correct. It would be interesting to reformulate this as lime-proto-librenet6 instead of a script that replaces other configurations already used.
Anyway the package was already added in #400. Maybe we should close this pull-request,

ilario:
I'm confused, does the issues raised by @G10h4ck apply also on #400?
If so, should we open an issue for tracking the need of changing the currently merged code?

gmarcos87:
That's right, we have to open the issue.

PR #401, opened 2018-08-13
remove lime-defaults from lime-system package

aparcar:
this should be done via a network profile instead


aparcar:
related https://github.com/libremesh/network-profiles/pull/33

aparcar:
I don't see why not use this approach, the libremesh-default profile could simply be installed per default

aparcar:
@p4u this is what I talked about. Since the lua module lime-config supports a fallback value, this solution would also work if somebody creates a network profiles without supplying the lime-defaults file

nicopace:
Could you test it?
Woulf be interesting to publish the defaults so people can understand what are they

On October 23, 2018 9:08:42 AM CDT, Paul Spooren <notifications@github.com> wrote:
>@p4u this is what I talked about. Since the lua module lime-config
>supports a fallback value, this solution would also work if somebody
>creates a network profiles without supplying the lime-defaults file
>
>-- 
>You are receiving this because you modified the open/close state.
>https://github.com/libremesh/lime-packages/pull/401#issuecomment-432261209

-- 

G10h4ck:
About people understanding, we have lime-example in lime-doc package, I agree in crating a libremesh-default that get installed in case no community profile is chosen

gmarcos87:
What would happen to network profiles that do not have lime-defaults, those that currently rest on default settings and only use the profile for other issues.

aparcar:
Maybe we can add `libremesh-profile` as a dependence of `libremesh-full` and each network profiles `PROVIDES` the `libremesh-profile`, see the idea [here](https://github.com/openwrt/packages/blob/864fd2bfee14d4decf8c1db03380a3d7aa080f15/sound/mpd/Makefile#L52).
If no profile is selcted it could fallback to `libremesh-default` or raise an error

aparcar:
Okay the whole thing is somewhat a pain. I added `libremesh-profile` as a dependence of `lime-system`. The result is initially as intended, when building a package with `wunderkammer-common lime-system` the profile is installed, when installing only `lime-system` the default profile called `zz-default` is installed. However, when installing `lime-system wunderkammer-common` the `lime-system` already *grabs* a *libremesh-profile* resulting in a conflict between `zz-default` and `wunderkammer-common`. 

A possible solution is the following: We install the profile `libremesh-default` via cooker and chef if no other is selected. People building from a build root usually know what they're doing and manage their own files.
This can however result in a situation where no `lime-defaults` are present, which is somewhat solved by offering fallback values.

As long as there is no clean solution for that, there is no working chef for LibreMesh. Please comment and think of a solution.

nicopace:
this conversation has been in general defined by @G10h4ck and @p4u .
I would love:
* for a default to exist in a file instead of in code (cause it is easy to reference to it),
* to have a set of configs belonging to the network (not specific to any node),
* to have a set of configs belonging to the node (specific to the node).

these three would be essentially the same file. the precedence would be if the more specific one doesn't have a certain info, it goes to the less specific: node -> network -> default

that way, we can use the FirstBootWizard to share the Mesh information (that will only have the customizations over the default), new defaults will be applied on the default without a problem.

I am thinking on the case were the LibreRouter (or any other router that will come with LibreMesh) will join a network or create a network.

Also, if we create a network config file, it is easier to upload it to a network-profile cause it is not mixed with the nodes config.

G10h4ck:
@aparcar do you have the companion PR ready? I think it will be easier to merge if the package that provides lime-default is there and a proof of it working out of the box whene no profile is selected and, it being not installed when a profile is selected would be great!

p4u:
https://github.com/libremesh/lime-packages/pull/434

aparcar:
This is currently solved via a different PR by @p4u, closing for now.

PR #404, opened 2018-09-19
batman-adv-auto-gw-mode: use /tmp/dnsmasq.d/ instead of -P/var/state

altergui:
Believe it or not, someone has been using this package ...for years! Reinhard from i4free-gr / 3zonelabs
The code is running on the ~15 nodes that comprise the local community network.
Reinhard reported that it does not work anymore in OpenWrt 18.06, so here's a hotfix.

using "uci set -P/var/state" was deprecated and now officially abandoned in OpenWrt 18.06.
But, now dnsmasq reads any file inside /tmp/dnsmasq.d/ as part of the config file, so use that instead.


aparcar:
lgtm

PR #405, opened 2018-09-21
Indicate to which kind of radios we want to apply wireless modes

gmarcos87:
These modifications allow to indicate to which radios we want to apply the modes.
For example if I want to open an ap but only in the 2ghz radio the configuration in lime-config is:

```
config lime 'wifi'
	list modes 'ap_2ghz'
	list modes 'adhoc'
	...
```
In the same way it could indicate that I only want to raise the mesh in 5ghz

```
config lime 'wifi'
	list modes 'ap_2ghz'
	list modes 'adhoc_5ghz'
	...
```
If no particular frequency is indicated, the mode apply to all radios.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -95,6 +95,7 @@ function wireless.configure()
 		local ignoredSuffix
 		local distance
 		local htmode
+		local modeSuffix
 		if modes[1] ~= "manual" then
 			if wireless.is5Ghz(radioName) then
 				freqSuffix = "_5ghz"
@@ -131,6 +132,8 @@ function wireless.configure()
 
 			for _,modeName in pairs(modes) do
 				local args = {}
+				modeSuffix = utils.split(modeName, "_")[2]	
+				modeName = utils.split(modeName, "_")[1]	
 				local mode = require("lime.mode."..modeName)
 
 				for key,value in pairs(options) do
@@ -154,7 +157,10 @@ function wireless.configure()
 					end
 				end
 
-				mode.setup_radio(radio, args)
+				if ( modeSuffix == nil ) or ("_"..modeSuffix == freqSuffix) then
+					mode.setup_radio(radio, args)
+				end
+
 			end
 		end
 	end


G10h4ck:
Seems nice, Have you tested it both with and without the configuration?

altergui:
please add the missing quote after 'wifi

(just in case someone copypastes your snippet.

Am 21. September 2018 18:11:38 OESZ schrieb Marcos Gutierrez <notifications@github.com>:

gmarcos87:
> Seems nice, Have you tested it both with and without the configuration?

I tried different options and always created configuration correctly. The code does not add any additional functionality just check if the mode has any added frequency and if that frequency matches the radio you are configuring.

aparcar:
This was proposed a while ago but denied, see here https://github.com/libremesh/lime-packages/pull/307

gmarcos87:
> This was proposed a while ago but denied, see here #307|
@aparcar It's funny that we came up with almost the same code without knowing it. 
It is necessary to solve this somehow, not just close the pull-request. Is there any way to detect if the radio is single band or dual band?

Adding a channel list instead of a single channel per radio type is also something to do.
If we think of Librerouter, we're going to have one 2.4ghz radio and two 5ghz radios. It would be interesting to be able to put the ap in 2.4ghz, mesh in 5ghz and even be able to indicate in which channel each radio.

gmarcos87:
Ok, I'll try lime-smart-wifi. I think the configuration scheme is a bit twisted, instead of options with 1/0 I could detect if the mode is selected.

Instead of
```
option mesh_5ghz 1
option all_to_ap 1
list channels_2ghz 1 11 6
list channels_5ghz 36 48 149
....
```

could be
```
option mode 'mesh_5ghz'
option mode 'ap'
list channels_2ghz 1 11 6
list channels_5ghz 36 48 149
....
```

gmarcos87:
This is not the same as lime-smart-wifi, it is simply to be able to set fixed modes for different radios. I am aware of the drawbacks of multiband radios, but anyway this does not break the experience of lime, does not interfere with the use of other modules or generate major issues. It is not a function that comes activated by default, it is simply to extend the possibilities of configuration in lime-config.

PR #406, opened 2018-09-28
Indicate to which radio we want to apply wireless modes

G10h4ck:
Move marcos patch to correct branch


nicopace:
In order to accept this PR, I would invite you to add documentation in lime-example, lime-web or as comments in the code.
That will make things easier for others to use this functionality.

PR #407, opened 2018-09-29
Feature/channel list

G10h4ck:
I haven't tested if UCI config system always return the channel list in same order but it should

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -52,7 +52,7 @@ config lime network
 
 config lime wifi
 #	option country 'ES'
-	option channel_2ghz '11'
+	option channel_2ghz '11'                    # May be either a list or a single option, in case of a list channel will be selected according to radio index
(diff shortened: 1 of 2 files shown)


aparcar:
I think we could remove the lime-smart-wifi package once this is merge to avoid redundant code

ilario:
I got into the channel list feature, it is cool and nicely implemented!
But I think that the if else in:
`chanIndex = chanIndex > 0 and chanIndex or 1`
is always true, so the line can be removed.
Do you agree? @G10h4ck

ilario:
Another question:
let's take the case that we have a list with two channel_5ghz and our wlan0 is 5 GHz, wlan1 is 2.4 GHz and wlan2 is 5 GHz.
Both wlan0 and wlan2 would get the first channel in the list, right?
Do we consider this as something to solve or we can ignore it?
(we can fix this in a dirty way adding a couple of counters out of the for loop, tell me if you want me to try)

PR #413, opened 2018-10-16
Fix sprunge evaluating commands with redirects

G10h4ck:
evaluating the command directly as ${2} didn't work well with commands
  that included redirects, use eval "${2}" for correct behaviour
Improve readability by breaking lines longer then 80 chars

### DIFF
--- a/packages/sprunge/src/sprunge.sh
+++ b/packages/sprunge/src/sprunge.sh
@@ -8,23 +8,41 @@ COMMAND="$0 $@"
 
 usage()
 {
-	echo -e "sprunge: command line paste bin\n\nSuggested ways of using $0 are mainly 4\nyourCommand arg1 arg2 | $0\n$0 [ -c \"yourCommand arg1 arg2\" ]\n$0 [ -f file ]\n$0; then EOF using Ctrl+D" 1>&2 ; exit 0
+	echo "\
+sprunge: Command line pastebin
+
+Suggested ways of using $0 are mainly 4:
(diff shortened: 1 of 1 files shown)


nicopace:
Question... why is this part of lime-packages? don't see it part of the lime stack (completely low priority question)

aparcar:
@G10h4ck shouldn't we move this package to the official OpenWrt repo?

Also instead of implementing the args, why not using `cat` (do one thing & do it good, right?)
```
command | sprunge
cat <file> | sprunge
cat | sprunge (interrupt with Ctrl+D)

G10h4ck:
> @G10h4ck shouldn't we move this package to the official OpenWrt repo?
> 
> Also instead of implementing the args, why not using `cat` (do one thing & do it good, right?)
> 
> ```
> command | sprunge
> cat <file> | sprunge
> cat | sprunge (interrupt with Ctrl+D)
> ```

the `command | sprunge` semantic is already implemented, the args are optional and have the extra effect to prepend to the paste the command which generated the output, with the `command | sprunge`  syntax sprunge has no way to know which command generated the output and so cannot paste it together with the output

as an example this has been generated by you guess it http://sprunge.us/Sk7sw0 while this http://sprunge.us/OZnQjj has been generated by  `npm install &> /dev/stdout | sprunge` having the command which generated the output together with the output help a lot when you are looking at debugging information provided by someone else, so you never miss-match the paste ;) so I think we are doing 4 things and doing it good :p

PR #414, opened 2018-10-22
lime-example improved documentation for wireless client mode

### LINKED ISSUE #383: broken client mode for wireless
```
config wifi 'radio0'
	list modes 'client'
	option client_key 'pass'
	option client_ssid 'ssid'
	option client_encryption 'psk2'

config net 'wirelessclient'
	option linux_name 'wlan0-client'
	list protocols 'wan'
```

creates the following output in `/etc/config/wireless`
```

config wifi-iface 'lm_wlan0_sta_radio0'
	option mode 'sta'
	option device 'radio0'
	option ifname 'wlan0-sta'
	option network 'lm_net_wlan0_sta'
	option ssid 'ssid'
	option encryption 'psk2'
	option key 'pass'
```

However I think `lm-net_wlan0_sta` isn't created. The official script simplify creates an `wwan` interface with dhcp and wan firewall enabled, I think we should do the same.

ilario:
The documentation of wireless client mode was a bit incorrect: the wireless interface is not wlan0-client but wlan0-sta, this should fix #383 
This also adds a comment recommending to check if the used radio name exists

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -80,20 +80,20 @@ config lime wifi
 ### WiFi interface specific options ( override defaults options )
 
 ## use radio0 only for mesh
-config wifi radio0
+config wifi radio0 # check existing radios with "wifi status" command
 	list modes 'ieee80211s'
 
 ## change ssid for radio1
-config wifi radio1
+config wifi radio1 # check existing radios with "wifi status" command
 	option ap_ssid 'Special'
 
 ## diable lime-config for radio2
-config wifi radio2
+config wifi radio2 # check existing radios with "wifi status" command
 	option modes 'manual' # If you use manual protocol you must not specify other protocol, or your configuration will be broken!
 
 
 ## set radio3 to adhoc with specific channel
-config wifi radio3
+config wifi radio3 # check existing radios with "wifi status" command
 	list modes 'adhoc'
 	option channel_2ghz '1'
 	option channel_5ghz '48'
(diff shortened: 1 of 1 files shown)


packages/lime-docs/files/lime-example
```diff
@@ -80,20 +80,20 @@ config lime wifi
 ### WiFi interface specific options ( override defaults options )
 
 ## use radio0 only for mesh
-config wifi radio0
+config wifi radio0 # check existing radios with "wifi status" command
```
nicopace: Shouldn't this be automatic?
Is this something that can be predicted?
If not, this setting can't come within the firmware, right?
Just curious on what would be the steps to use this functionality, and if we can make it simpler enough for the user to use (like adding wan connection on the lime-app for manual configs (wifi and manual ip and pppoe info)).

nicopace: and what about the usecase?
I never had to do this, so I don't see the case

ilario: some usecases of specific radio configuration:

. use a radio as client
. use a radio just for mesh and another radio just for ap

G10h4ck: The comment is a bit misleading to me, i undestard it as if the option is meant to check for the existence of a radio, while it is not like that, I think the comment should explain what is the paramether for and then as an extra add "you should ensure the radio exists, for exaple usign `wifi status` command"

nicopace: @ilario those are not usecases, that is what you want to accomplish.
A usecase would be a case when a user needs that way of using the radio.
Like: As a user i want to get internet over wireless to share with the mesh
... Or something like that.
And with that, would be very interesting to know how frequent it is. Thanks!

ilario: @G10h4ck improved

ilario: @G10h4ck improved

ilario:
The new #405 and #419 reduces a bit the need to use radio specific configuration, making this much easier for some scenarios.

PR #419, opened 2018-10-24
Add documentation about pull/405

gmarcos87:
In #405 (merged) we add the ability to indicate to which type of radios we want to apply the wireless modes.
In this pull request I add that information to lime-example


ilario:
That's very convenient!
In case of more than one radio for a band (for example LibreRouter will have 2 radios at 5 GHz) is the configuration applied to both? Is there a way to specify just one of those or the radio specific configuration is still needed for this?

G10h4ck:
Specific configuration still needs to be specifics.

Issue #409, opened 2018-10-07
Static proto: improve setting IPv6 and implement setting gateway


ilario:
Static proto does support IPv6 but I didn't get how to do that...
In lime-example it says:
```
list protocols 'static:172.23.0.1/24' # Set up a static IP (both IPv4 and IPv6 supported, specify twice one per IP type)
```
but setting IPv6 fails due to the colon (:) being interpreted as arguments separator instead of part of the IPv6, escaping them with backslash does not work neither.

nicopace:
@G10h4ck can you jump on this?

G10h4ck:
This comment on the code https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/usr/lib/lua/lime/proto/static.lua#L21 suggests it is supported, I don't know if Pau tested it,  @ilario can you give more information about the error you get?
Anyway the code seems a bit strange to me and I would be surprised if it worked... moreover it miss how to set a gateway and other things usually needed in static setups

nicopace:
What is strange about it?

how would you use it?

Can you give us an example?
Like, one per option:
* ipv4
* ipv6
* ipv4+ipv6

thanks!

ilario:
> moreover it miss how to set a gateway and other things usually needed in static setups

Exactly, see #408 

> can you give more information about the error you get?

The colon is interpreted as an arguments separator instead that being interpreted as part of an IPv6.
For example, if I set:
```
list protocols 'static:192.168.1.5/24'
list protocols 'static:aaaa::1111/64'
```
I obtain the following error:
```
# lime-config
Clearing wireless config...
Clearing network config...
Disabling odhcpd
Cleaning dnsmasq
Disabling 6relayd...
/usr/lib/lua/lime/network.lua:276: attempt to index local 'addr' (a nil value)
stack traceback:
	/usr/lib/lua/lime/network.lua:252: in function </usr/lib/lua/lime/network.lua:252>
	/usr/lib/lua/lime/network.lua:276: in function 'createStaticIface'
	/usr/lib/lua/lime/proto/static.lua:23: in function 'setup_interface'
	/usr/lib/lua/lime/network.lua:251: in function </usr/lib/lua/lime/network.lua:251>
	[C]: in function 'xpcall'
	/usr/lib/lua/lime/network.lua:251: in function </usr/lib/lua/lime/network.lua:210>
	[C]: in function 'xpcall'
	/usr/bin/lime-config:45: in function 'main'
[...]

PR #420, opened 2018-10-27
lime.network pass down to lime.proto specific section data

G10h4ck:
this way lime.proto can reliably access to specific section data when used in a specific section

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -265,6 +265,7 @@ function network.configure()
 		if owrtIf then
 			deviceProtos = owrtIf["protocols"] or {"manual"}
 			flags["specific"] = true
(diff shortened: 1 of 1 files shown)


nicopace:
Can we add documentation on the PR?
Guess there is some relation with #409

G10h4ck:
> The args._specific_section does not end up containing just the string of the section name, indeed it is a table with a lot of content, for example this comes out from the lime-hwd-openwrt-wan section where I added a "static_ipv4" option:

That is on purpose passing a table in lua is very cheap (only a reference to the table is passed under the hood so the table is not copied) so any protocol can easily access all information it might eventually need

ilario:
Seems that no more documentation is being provided by @G10h4ck, the comments here are some sort of documentation, very difficult to find one anyway. I'm closing, @nicopace if you think of a better place for documenting this please feel free to re-open.

PR #425, opened 2018-11-10
Added warning for invalid IPv4 settings

ilario:
In case an invalid IPv4 was set in /etc/config/lime, network.lua silently sets a fallback one.
The users can have troubles in understanding what happened ([example](https://lists.libremesh.org/pipermail/lime-users/2018-November/001333.html)), this could be mitigated by a warning message.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -87,6 +87,7 @@ function network.primary_address(offset)
 	if invalid then
 		ipv4_template = mc:maxhost()
 		ipv4_template:prefix(tonumber(ipv4_maskbits))
+		print("INVALID IPv4 DETECTED, CHECK YOUR SETTINGS. USING " ..tostring(ipv4_template))
 	end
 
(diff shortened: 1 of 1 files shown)


packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -87,6 +87,7 @@ function network.primary_address(offset)
 	if invalid then
 		ipv4_template = mc:maxhost()
 		ipv4_template:prefix(tonumber(ipv4_maskbits))
+		print("INVALID IPv4 DETECTED, CHECK YOUR SETTINGS. USING " ..tostring(ipv4_template))
```
G10h4ck: I would put in the log message the exact name of the option line network.ipaddr or similar

ilario:
@G10h4ck I tried to improve the message, now it's exhaustive in my opinion

G10h4ck:
Did you tested if it works as expected both in case the address is valid and in case is not with the addition of `and "SOMETHING"` ?

ilario:
Yes I did. I tested the good, the anygw and the boradcast case. Thanks to my router's MAC address (ending with :00) I even managed to test the network address case :)

Issue #415, opened 2018-10-22
Wifi station mode in backbone via routing protocols instead than br-lan


ilario:
Currently the wifi client mode works just if used as WAN, so with a natting and a DHCP client for having the internet access.
To build a point to multipoint (as opposed to mesh) network can be a great option in many cases (for example when the topology is a tree or when the hardware does not support neither adhoc nor ieee80211s or if we want to use the better supported ap-sta mode rather than mesh) and being able to use wifi client in LAN network is what's missing.
The main trouble in doing this is described [here](https://wiki.openwrt.org/doc/howto/clientmode#bridged_client_mode_issues).
A proposed solution is to include the wlan0-sta interface in bat0 bridge instead than in br-lan bridge.
Related issues: #127, #383, #47

nicopace:
Hi ilario,
I never came across this setup... can you share a more detailed example on how would you expect to use it, and if you have a rough idea on how to implement it?

ilario:
I will copy here a comment I made originally elsewhere, tell me if it is clear to you:

> a user could want to build a point to multipoint network with hardware that does not support adhoc nor ieee802.11s (I was planning to do this with a lot of modem-routers we received as donations with the local hacklab), for doing this some nodes should be configured as client via radio specific configuration or general configuration (unsupported, I think), see #415 for some problems hindering this

Another reason for using AP-STA as a backbone could be a better support in the drivers for these two modes rather than for adhoc or for ieee802.11s, so for performances and stability reasons.

If the inclusion in bat0 is a solution (it worked for me, maybe there's a better one? @G10h4ck?) I would suggest to expand [this line in lan.lua](https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua#L46) into a section managing the interface inclusion in bat0 (is there also the risk that this interface configuration happens before bat0 is up?)

nicopace:
Awesome.
No particular interest of having this kind of networks, neither know anyone that uses them (other than traditional WISPs), but please, no problem on supporting it, sound like falls into something we could be interested in having as a collective.

G10h4ck:
wifi client can't be used in LAN because there is no standard way to bridge a wifi STA interface

ilario:
@G10h4ck read above

ilario:
[...]

Issue #127, opened 2017-05-18
Documentation needed for WiFi client mode


ilario:
Currently [wifi client mode](/libremesh/lime-packages/blob/develop/packages/lime-system/files/usr/lib/lua/lime/mode/client.lua) allows you to configure in `/etc/config/lime` both WAN or LAN network (for discussion see issue #47 and description of commit 6ded289).
We need to add in `/etc/config/lime-example` (the location for basic documentation) a reminder and an example for this mandatory LAN/WAN setting.

ilario:
@p4u @G10h4ck how can I predict the linux_name of the client interface?
This is needed for writing an example of documentation.
For making it clear, an example of minimal client configuration could be:

```
config wifi wificlientradio
	list modes 'client'
	option channel_2ghz 'auto'
	option client_ssid 'SomeWiFiNetwork'
	option client_key 'SomeWPApskPassword'
	option client_encryption 'psk'

config net wificlientnet
	option linux_name '??????????????????'
	list protocols 'wan'
```


what to use for linux_name?

G10h4ck:
> what to use for linux_name?
wlanX-client remember to substitute X with the number of the interface

in general the names are composed this way

wlan${RADIO_NUMBER}-${LIME_WIFI_MODE_NAME}

ilario:
So it would be safe to use `wlan0-client` in lime-example and explain in comments that the user have to manually verify this via ssh or web interface?

ilario:
Correct me if I'm wrong:

* when using WiFi client mode to connect to a **non-LiMe** AP, `list protocols 'wan'` should be used;
* when connecting to a **LiMe AP** (AP-sta could be a good idea for node to node connections, mainly when using hardware which doesn't suppot ieee80211s or adhoc), `list protocols 'wan'` should be **avoided** as a NATting would create loops in the network, instead something like `list protocols lan
	list protocols anygw
	list protocols batadv:%N1
	list protocols bmx6:13` should be used.

ilario:
Can we discuss why the assumptions in my last comment are wrong?
I suppose they are 'cause 12ec6b5 deleted the wan/lan distinction, but this was not explained anywhere.
@G10h4ck pls

G10h4ck:
You use wan to get upstream connectivity, not to connect to a libremesh network, while using lan would require bridging and wifi client interface are not capable of that in a standard way

ilario:
[...]

PR #426, opened 2018-11-11
Wifi client mode in backbone via BMX6

ilario:
For the original discussion see #415, of which this PR is a partial solution.
Partial as it allows to use BMX6 on client-AP links, but does not allow yet BATMAN-adv and even less BATMAN-adv+BMX6.

TL;DR
--------
As including client interfaces in bridges is not possible (see #415 and #127), we can include them between the interfaces used by the routing protocols and obtain the same result. This can be done via specific interface configuration.

BMX6 on client-AP link
-----------------------------
This PR allows this. Adding a `proto/client.lua` (I just copied `proto/ieee80211s.lua` and replaced "ieee80211s" with "client") was needed in order to set up the interface. For running this new proto, a `list protocols 'client'` has to be included in the specific interface configuration. On the AP side the wlanX-ap interface is included in br-lan, which is used by BMX6. As the wlanX-ap interface does not have a VLAN ID, we have to explicitly set 0 as the VLAN ID also on the client side, with a configuration resulting in:

```
option wifi radio0
	list modes 'client'
	option channel_2ghz 'auto'
	option client_ssid 'LibreMesh.org'
	option distance 1000
config net wirelessclientBackbone
	option linux_name 'wlan0-sta'
	list protocols 'client'
	list protocols 'bmx6:0'
```

In case VLAN is needed, it has to be set also as specific AP configuration, which is not implemented yet. 

BATMAN-adv on client-AP link
--------------------------------------
BATMAN-adv would need specific AP configuration as currently is not included in bat0 (decided in 2014 213efd0). Specific AP configuration is not implemented yet.

BATMAN-adv+BMX6 on client-AP link
-----------------------------------------------
Additionally to what just exposed, having both BMX6 and BATMAN-adv is hindered by the fact that BMX6 sets the `proto` field of the interface as `static`, while BATMAN-adv requires it to be set as `batadv` otherwise the interface does not get included in bat0. This behaviour of BMX6 comes from a 2014 workaround 83e5a2d trying to fix [this issue](http://old.libremesh.org/issues/38), which wouldn't be present if the proto is set by BATMAN-adv.

The problem of this is that it does not setup the sta interface, so the resulting /etc/config/network is broken.
We need a proto/client.lua (copying the proto/ieee80211s and replacing all ieee80211s occurrences with client works) and then, important, the specific configuration has to include a list protocols 'client' in order to activate proto/client.lua

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -113,6 +113,7 @@ option wifi radio4 # you should ensure that the chosen radio name exists, for ex
 	option client_ssid 'SomeWiFiNetwork'
 	option client_key 'SomeWPApskPassword'
 	option client_encryption 'psk' # psk for WPA or psk2 for WPA2
+	option distance 300 # maximum distance to AP, heavily affects performances
 
 config net wirelessclientWAN
 	option linux_name 'wlan0-sta' # the client interface name could be named differently, like wlan1-sta
@@ -123,6 +124,7 @@ option wifi radio5 # you should ensure that the chosen radio name exists, for ex
 	list modes 'client'
 	option channel_2ghz 'auto'
 	option client_ssid 'LibreMesh.org'
+	option distance 1000 # maximum distance to AP, heavily affects performances
 
 config net wirelessclientBackbone
 	option linux_name 'wlan0-sta' # the client interface name could be named differently, like wlan1-sta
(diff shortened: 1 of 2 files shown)


ilario:
2 months passed and no reactions...
This PR already allows to setup a working point to multipoint or point to point network (non 802.11s nor adhoc, simply AP-client, the most reliable and performing option I think) having at least BMX6 (the presence of BATMAN-adv is not a problem but will not be used on the AP-client links).

Can anyone review please? @G10h4ck @aparcar @p4u

Issue #408, opened 2018-10-07
Static proto: specify gateway


ilario:
When a static IPv4 address is set via specific interface configuration, there should be a way to set also the gateway, needed when configuring it for a WAN.

In /docs/lime-example one can read:

```
list protocols 'static:172.23.0.1/24'		# Set up a static IP (both IPv4 and IPv6 supported, specify twice one per IP type)
```
The gateway could be accepted as a second optional argument, like this:

```
list protocols 'static:192.168.1.2/24:192.168.1.1'
```

nicopace:
This is implemented in here: 
https://github.com/libremesh/lime-packages/blob/d2ba7a7b85ac6f4bb26741fd365016dccb39545d/packages/lime-system/files/usr/lib/lua/lime/proto/static.lua#L16

anyone who might want to implement what @ilario is proposing should start from there.

I am not sure if it is fine to define that there... @G10h4ck ?

ilario:
Closing this and continuing on #409: from the discussion there seems that the issue is basically the same.

PR #427, opened 2018-11-12
Static proto: add setting for gateway and clean IPv6 setting

### LINKED ISSUE #409: Static proto: improve setting IPv6 and implement setting gateway

ilario:
Depends on #420 
This should fix #409 and #408 
Changes the settings of a static IP from
```
	list protocols 'static:1.2.3.4/24'
	list protocols 'static:1:2:3:4:5:6:7:8/64'
```
to
```
	list protocols 'static'
	option static_ipv4 '192.168.1.2/24'
	option static_gateway_ipv4 '192.168.1.1'
	option static_ipv6 '2a00:1508:0a00::1234/64'
	option static_gateway_ipv6 'fe80::1'
```
the advantages are: possibility to set a gateway, support for any IPv6 format (compressed with :: or not).

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -126,7 +126,11 @@ config net wirelessclient
 #config net port5					# Do not put any "." in the section name
 #	option linux_name 'eth1.5'			# Put here the actual name of the interface
 #	list protocols 'wan'				# Some of these protocols require the relative package "lime-proto-..."
-#	list protocols 'static:172.23.0.1/24'		# Set up a static IP (both IPv4 and IPv6 supported, specify twice one per IP type)
+#	list protocols 'static:172.23.0.1/24'		# Set up a static IP (both IPv4 and IPv6 supported)
+#	option static_ipv4 '192.168.1.2/24'
+#	option static_gateway_ipv4 '192.168.1.1'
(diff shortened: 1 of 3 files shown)


ilario:
Tested, it works as expected.

The limitations are:
1) just one IPv4 and one IPv6 per interface can be set. If setting more than a static IP per type is a needed feature, the parsing in proto/static.lua should be done with a for loop.
2) only IP and gateway IP can be set, no other option (there are [a few more for IPv6](https://openwrt.org/docs/guide-user/network/ipv6/start#protocol_static_with_ipv6), but I never used them).

G10h4ck:
You could avoid the limitations by naming the options like this static_openwrtoptionname and then just loop trought static_* option and "copy/pasting" them into openwrt configurations, tell me if you will do that soon, otherwise I'll merge this PR as is

ilario:
I cannot work on this before January :/

PR #428, opened 2018-11-16
proto-anygw filter rogue router advertisement

G10h4ck:
untested


G10h4ck:
The rogue routes doesn't get announced by libre-mesh routers, but by home routers that often people attach in their network as network extenders and similar, this prevent the rogue announcement to spread to the whole community, so only the one who attached the rogue router get it's piece of network broken

aparcar:
@G10h4ck looks somewhat nice. Don't you think it's better to add this to `/etc/config/firewall` as UCI? This way it's way more readable and integrates better in the rest of the system, not?

G10h4ck:
No this is ebtables, not iptables, also I avoid openwrt firewall with all my heart, and we have the agreement to not have an hard dependency on it

ilario:
From @nicoechaniz:

> if you plug a modern tp-link to a LibreMesh network it will start doing a rogue Router Advertisement
> and will break IPv6
> it is enabled out of the box
> in the current tp-link firmware

aparcar:
@G10h4ck @ilario  are you gonna test this?

ilario:
@aparcar nope, I'm not going to test this as I am not able to.

PR #439, opened 2018-12-18
Fix lime-docs package

### LINKED ISSUE #438: lime-docs doesn't build
lime-docs fails to build due to wrong source URL (`svn` proto on a github subdirectory -- that cannot ever have worked)

p4u:
Instead of using svn for checking out a subdirectory of lime-web
repository, use git and checkout a specific branch named onlydocs which
contains only the documents (not the whole web).
Fix #438

### DIFF
--- a/packages/lime-docs/Makefile
+++ b/packages/lime-docs/Makefile
@@ -13,9 +13,9 @@ PKG_NAME:=lime-docs
 PKG_VERSION=$(GIT_COMMIT_DATE)-$(GIT_COMMIT_TSTAMP)
 
 PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)-$(PKG_VERSION)
-PKG_SOURCE_URL:=https://github.com/libremesh/lime-web/trunk/docs
-PKG_SOURCE_PROTO:=svn
-PKG_SOURCE_VERSION:=HEAD
+PKG_SOURCE_URL:=https://github.com/libremesh/lime-web
+PKG_SOURCE_PROTO:=git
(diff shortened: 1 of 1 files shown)


ilario:
I still have to test, but: doesn't this download the whole lime-web anyway?

If this is correct (that downloads the whole lime-web and then checks out the onlydocs branch), why not using directly master branch (so we do not have to worry about updating the onlydocs branch)?

ilario:
#438 has been fixed on Github side, so in my opinion is better to stay with the current solution (which does not need updating `onlydocs` branch) until we find a better one based on git.
Do you agree @p4u @aparcar ?

p4u:
> If this is correct (that downloads the whole lime-web and then checks out the onlydocs branch), why not using directly master branch (so we do not have to worry about updating the onlydocs branch)?

I don't know exactly what git command is executed by openwrt-buildroot, maybe it fetch only the branch files or maybe it downloads all the branches. So I cannot actually answer this question :(

But it's true that download the whole web (4MB) is not that problematic cause it's done in the host side (not in the router side). So IMO the best solution would be to use git, download the master branch and copy only the docs/* to the package.

aparcar:
I think [this](https://github.com/openwrt/openwrt/blob/master/include/download.mk#L199) is the buildroot method, which also removes the `.git` folder

p4u:
Ok, I like using wget instead of git and dowload https://codeload.github.com/libremesh/lime-web/tar.gz/HEAD but not gonna work on it soon, so feel free to send your PR :)

Close this.

PR #441, opened 2018-12-29
Update lime-example

p4u:
Fix client configuration example

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -107,7 +107,7 @@ config wifi radio3 # you should ensure that the chosen radio name exists, for ex
 # network specific one like in the following example.
 
 ## set radio4 as client of access point, both the following "wifi" and "net" sections are required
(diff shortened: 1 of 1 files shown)


p4u:
You'r right in the three points, I just saw this "mistake" and made a quick commit (via GitHub web interface). So I did not check all the lines in the document neither think too much about this change.

Feel free to PR yourself or let's open an issue.

G10h4ck:
The purpose of such high number is to force someone who attempt to copy paste things from the example to understanding what is she doing before it have any effect, to avoid someone drunk to brick their router just by copy/pasting things that she doesn't understand from an example file.

I don't see what improvement bring substituting 12 with 0 in a documentation file, even IP RFC use reserved IPs for documentation probably to avoid collisions by drunk copy paste, so I think that using radio12 in the documentation instead of radio0 is a good practice. And I would prefer to not change it as it force to understand before any effect (except if you got a router with at least 12 radios :p ).

PR #444, opened 2019-01-08
Introduce bmx7-auto-gw-bw-mode package

p4u:
If the node is an Internet gateway, checks the bandwidth (downloads a
10MB file) and announces it to the mesh. It integrates with watchping
and bmx7-auto-gw-mode.

The check is automatically done via crontab every 6 hours.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,8 +49,10 @@ deploy:
       branch: master
   - provider: script
     script:
-    - rsync --delete-after -L -r -v -e "sshpass -p $SSH_PASS ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -p $CI_PORT"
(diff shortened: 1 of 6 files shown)


p4u:
It's the best way I found to run a bw test to Internet without requiering extra tools.

ilario:
Sounds like a good method for the bandwidth check, but shouldn't we include wget as dependency? Or is it included in Busybox also?

p4u:
wget is a requirement for base openwrt

p4u:
I mean, busybox wget

aparcar:
I'll fox the ci soonish so we can test the plugin. please add your tool to the libremesh.sdk.config to actually build the package

p4u:
I'd not add the package to lime-sdk until it's merged in master.

aparcar:
https://github.com/libremesh/lime-packages/blob/master/libremesh.sdk.config i mean this one

PR #447, opened 2019-01-10
Introduce new package bmx7-mdns

p4u:
bmx7 mesh DNS is a tool to provide simple distributed DNS functionality to the mesh network.
Uses bmx7-sms plugin to announce domains and IPs and dnsmasq for local resolve.
TLD are by default ".mesh" and ".mesh6" but can be changed, not matching TLDs are ignored (to avoid poisoning).
Node hostname is automatically announced, but also other domains can be published using syntax name.tld@IP
Comes with a simple LUCI interface to monitor announced domains and add new ones to the mesh network.
It's based on 2013 qmp.cat mDNS for bmx6 code.

### DIFF
--- a/libremesh.sdk.config
+++ b/libremesh.sdk.config
@@ -30,6 +30,7 @@ CONFIG_PACKAGE_lime-system=m
 CONFIG_PACKAGE_lime-webui=m
 CONFIG_PACKAGE_bmx7-auto-gw-mode=m
 CONFIG_PACKAGE_bmx7-auto-gw-bw-mode=m
+CONFIG_PACKAGE_bmx7-mdns=m
 CONFIG_LIMEWEBUI_ES=y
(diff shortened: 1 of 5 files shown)


dangowrt:
if that can support records other than type A and AAAA, that'd be
super nice. like TXT records or SRV (think: _sip._udp....) it could
allow service announcements as well... or E164/ENUM records to announce
numeric phone numers and have mesh-potato-like analog phone network
functionality with asterisk resolving those records...

p4u:
Agree, that would be super nice to have, but it's out of scope for this PR. This simple DNS (A/AAAA) system will solve a current need for libremesh networks. In the future a better and more complex system might be implemented, but for such features you propose I think we would need to implement our own DNS resolver with bmx7-sms/alfred/blah support or use something like named/bind instead of dnmasq.

PR #448, opened 2019-01-12
If bmx6 not installed do not crash lime-webui

p4u:


p4u:
So I bet I'm the first trying to use lime-webui without bmx6....

aparcar:
thanks

PR #449, opened 2019-01-13
Allow use first IP of ipv4 range if anygw not enabled

p4u:
IP like 10.0.0.1/24 should be possible as main address if anygw not installed nor enabled.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -80,8 +80,16 @@ function network.primary_address(offset)
 	local mc = ipv4_template
 	--! Generated address is network address like 192.0.2.0/24 ?
 	local invalid = ipv4_template:equal(mc:network()) and "NETWORK"
-	--! Generated address is the one reserved for anygw like 192.0.2.1/24 ?
-	invalid = invalid or ipv4_template:equal(mc:minhost()) and "ANYGW"
+	--! If anygw enabled, generated address is the one reserved for anygw like 192.0.2.1/24 ?
+	if utils.isModuleAvailable("lime.proto.anygw") then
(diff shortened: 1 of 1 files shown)


packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -80,8 +80,15 @@ function network.primary_address(offset)
...
 	--! Generated address is network address like 192.0.2.0/24 ?
 	local invalid = ipv4_template:equal(mc:network()) and "NETWORK"
-	--! Generated address is the one reserved for anygw like 192.0.2.1/24 ?
-	invalid = invalid or ipv4_template:equal(mc:minhost()) and "ANYGW"
+	--! If anygw enabled, generated address is the one reserved for anygw like 192.0.2.1/24 ?
+	if utils.isModuleAvailable("lime.proto.anygw") then
+		local generalProtocols = config.get("network", "protocols")
+		for _,protocol in pairs(generalProtocols) do
+			if protocol == 'anygw' then
+				invalid = invalid or ipv4_template:equal(mc:minhost()) and "ANYGW"
```
G10h4ck: At first glance I would say a `break` should be added just after this line ;)

p4u: Upss, you'r right. Fixed

p4u:
ping @G10h4ck

G10h4ck:
Looking at it again I realized should't it be the same for IPv6 too ?

p4u:
Probably. But this PR refers specifically to the IPv4 range. Another PR for the IPv6 might be sent.

@G10h4ck merging is blocked until you mark the requested changes as solved (which IMO they are).

ilario:
The check for invalid IPv6 has not been implemented yet, as the probability of a collision is less likely.

Issue #144, opened 2017-06-07
Ath9k radio, MCS stuck workaround


p4u:
According to the information provided by @nicoechaniz (https://github.com/libremesh/network-profiles/blob/master/quintanalibre.org.ar/comun/etc/uci-defaults/92-cron-workaround-tasks), some Atheros radios (the ones from Tplink 3500 and Tplink 842) stock in a strange state where the driver thinks MCS can be higger than 7 but it might be not. In consecuence throughput is slowly reducing until it almost lost connectivity with other nodes. 

Performing a scan (in more than one frequency) fixes this issue and the radio goes to the normal state again.

It would be needed to find a way to identify this problem (when is it happening) and perform an acction to fix it and save some debug information to fix it upstream. Smonit might do this job (#137).

Current workaround is a cron executed every 5 minutes:

```
iw wlan1-adhoc scan freq 5745 5825
iw wlan0-adhoc scan freq 2412 2457
```

nicoechaniz:
this is a very tricky bug.

Here in QuintanaLibre we are able to tell that the MCS rate negotiated is unrealistic because of the split MIMO chains scenario. A node with its chains splitted in opposite directions should never negotiate an MCS15 rate with any distant node, but we have found this to happen regularly.

A node acting properly will show something like this when doing an iw station dump:

```
root@oncelotes:~# iw wlan1-adhoc station dump
Station 64:70:02:3d:84:03 (on wlan1-adhoc)
	inactive time:	10 ms
	rx bytes:	1297596979
	rx packets:	6589765
	tx bytes:	2513743403
	tx packets:	2049296
	tx retries:	449374
	tx failed:	18
	rx drop misc:	12870
	signal:  	-64 [-64, -85] dBm
	signal avg:	-67 [-67, -87] dBm
	tx bitrate:	135.0 MBit/s MCS 6 40MHz short GI
	rx bitrate:	108.0 MBit/s MCS 5 40MHz
	expected throughput:	40.191Mbps
```

It's visible there that the chains are split (signal level is very different on each chain) and the negotiated MCS in this scenario can not go beyond MCS7.

So, a not so tidy way to monitor this effectively would require parsing iw station dump output and reacting to not invalid MCS rates.

dangowrt:
@ecsv @nbd168 This ticket does provide some interesting hints towards where to look for the bug...
I recon it is related to #130 .
[...]

PR #451, opened 2019-01-17
Remove sleepy radio hook

p4u:
The implementation of this "dirty fix" should not affect all targets,
but only those which are afected.

The implementation is just wrong, smonit executes quick-fix threee times
and then 1 time long fix if the problem persists.


p4u:
This quick fix was added in some release to avoid a problem on wdr3500/3600/4300 routers. But it was a patch just for that release, I should have never been merged to master.

In addition the implementation is wrong, so it does not even works (only the first 20 minutes)...

So I'd remove it until a better one is implemented (if still needed). If some community wants to keep using it, they can use the network-profiles.

gmarcos87:
Hi @p4u, Can you give more information that this bug is resolved in 18.06.1?

p4u:
I don't know (actually I never undestood this bug).

But I know that this patch does not fix it, so there is no point of having this dirty script here.

p4u:
I'm not saying it does not exist, I'm just saying I have never experienced it and the removed file on this PR does not fix. 

If "cotonete" workaround works, perfect! But this PR is still valid.

gmarcos87:
The difference is that cotonete is an optional package and smonit is included in all libremesh by default

p4u:
Then, rather cotonete should be enabled by default or smonit script must be fixed. But the current one just does not work.

In addition this workaround should be executed **only** when the device is affected by the bug, it must not be enabled always by default.

aparcar:
Why don't add cotonete per default? Does it add any overhead on *not affected* devices?

ilario:
Related #144

nicopace:
This package is an improvement of the implementation done in smonit, that shouldn't be removed.
The issue with the smonit implementation is that it takes time since the moment it detects the issue till the moment it implements a solution, though leaving the node unconnected during that time.
The specific implementation within smonit is part of trunk, so i don't see why that would be a reason to take it out of the repository.
In fact, this is a rework of the smonit implementation that the original author of the solution the smonit implementation was based on (@nicoechaniz ) and I invested a lot of time in polishing it.
The minimal task I woud say, instead of removing it, would be to see why in some cases is not working and fixing it, cause this provides a much faster (so more effective) way of dealing with the issue smonit (and the workarounds in quintanalibre profile for example) deal with.

nicopace:
For now it doesn't implement the fix, only logs when the problem is reproducible, so no harm done on interfaces that don't have the issue.

Please, don't remove.

nicopace:
Sorry @p4u, I rushed into it.
Will review now...

p4u:
Regarding smonit, her the context: https://github.com/libremesh/lime-packages/commit/e8a27e19d6cb470e035d5437df9adabf8240e926


```
Small and modular monitor for daemons.
It executes every 5 minutes and run a set of hooks to check the
daemons and status of the system. Hooks are installed on /etc/smonit and
must define a function for quickfix (like daemon restart) and longfix
(like rebuild configuration for daemon). Longfix is launched only if
after 3 quickfixes the problem is not solved. Logs are stored on /tmp/smonit/<daemon_name>.
```

p4u:
And here an example of what happens with sleepy_radio hook on a libremesh node. See last message `[2019/01/18 16:30] sleepy_radio -> No recovery after longfix, doing nothing else`

```
$ cat /tmp/smonit/smonit.log | grep sleepy
[2019/01/18 16:05] sleepy_radio -> Executing
[2019/01/18 16:05] sleepy_radio -> Status: error
[2019/01/18 16:05] sleepy_radio -> Fails:0, executing quick fix
[2019/01/18 16:10] sleepy_radio -> Executing
[2019/01/18 16:10] sleepy_radio -> Status: error
[2019/01/18 16:10] sleepy_radio -> Fails:1, executing quick fix
[2019/01/18 16:15] sleepy_radio -> Executing
[2019/01/18 16:15] sleepy_radio -> Status: error
[2019/01/18 16:15] sleepy_radio -> Fails:2, executing quick fix
[2019/01/18 16:20] sleepy_radio -> Executing
[2019/01/18 16:20] sleepy_radio -> Status: error
[2019/01/18 16:20] sleepy_radio -> Fails:3, executing long fix
[2019/01/18 16:25] sleepy_radio -> Executing
[2019/01/18 16:25] sleepy_radio -> Status: error
[2019/01/18 16:25] sleepy_radio -> No recovery after longfix, doing nothing else
[2019/01/18 16:30] sleepy_radio -> Executing
[2019/01/18 16:30] sleepy_radio -> Status: error
[2019/01/18 16:30] sleepy_radio -> No recovery after longfix, doing nothing else

```

gmarcos87:
As longfix we could call wifi on all ad-hoc or mesh interfaces. This is the fix that takes the longest but guarantees the result.

p4u:
> As longfix we could call wifi on all ad-hoc or mesh interfaces. This is the fix that takes the longest but guarantees the result.

Yes, that would make more sense. Quick fix makes the scan and longfix recreates the wifi devices... but **that requires a very trusty way of detecting the problem**, cause that's very drastic (actually the scann is also quite dramatic in therms that the radio stops working during a period). Please, take into account that many people around the world does not use tplink-wdr neither ath9k driver.

PR #454, opened 2019-01-21
build lime-app via ci

aparcar:


aparcar:
The CI already build it https://snapshots.libremesh.org/branches/lime-app-ci/lime-app_v0.1.0-alpha.4-1_all.ipk

PR #466, opened 2019-02-07
Add safe-upgrade package

spiccinini:
Provides dual-boot functionality with automatic revert safety, mainly for >=16MB devices like the LibreRouter

### DIFF
--- a/packages/safe-upgrade/bootscript_gen.py
+++ b/packages/safe-upgrade/bootscript_gen.py
@@ -1,0 +1,31 @@
+script = """
+run preboot;
+boot_part=${stable_part};
+if test ${testing_part} -ne 0; then
+    echo Testing part ${testing_part};
+    boot_part=${testing_part};
+    # saving environment so next boot it defaults to stable_part
+    set testing_part 0;
+    saveenv;
+fi;
+if test ${boot_part} -eq 2; then
+    fw_addr=${fw2_addr};
+    run boot_2;
+else
+    fw_addr=${fw1_addr};
+    run boot_1;
+fi;
+# if some error happened
+run boot_1;
+# everythong else failed, raw booting
+bootm ${fw1_addr};
+"""
+
+
+def onelinerize(data):
+    import re
+    # remove comments
+    data = re.sub(r'#.*', '', data)
+    return data.replace("\n", " ").replace("    ", "").replace("  ", " ").strip()
+
+print("set bootcmd '%s'" % onelinerize(script))
--- a/packages/safe-upgrade/Makefile
+++ b/packages/safe-upgrade/Makefile
@@ -1,0 +1,50 @@
+#
+# Copyright (C) 2019 Santiago Piccinini <spiccinini@altermundi.net>
+#
+# This is free software, licensed under the GNU General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
+GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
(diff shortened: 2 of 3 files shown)


packages/safe-upgrade/bootscript_gen.py
```diff
@@ -0,0 +1,31 @@
+script = """
```
aparcar: may add a *She-Bang* aka `#!/bin/python3` here

nicopace: @spiccinini couldn't this be done in lua? it is too simple to have it in Python ... looks unnecesary.
There is one example (very similar to yours) here: http://lua-users.org/wiki/TextTemplate
And one not that similar but much simpler here: http://lua-users.org/wiki/StringInterpolation

ilario: @nicopace I think (please @spiccinini comment on this) this file is not going to be run on the routers, indeed it is not going to be installed on the router. So in my opinion having it in Python is not a problem

nicopace:
It makes it easier for other to maintain, as it is all in one single language.
Not mandatory... But the less language diversity, the easier to get the skills to maintain or contribute

On February 7, 2019 11:15:32 AM CST, Ilario Gelmetti <notifications@github.com> wrote:
>ilario commented on this pull request.
>
>
>
>> @@ -0,0 +1,31 @@
>+script = """
>
>@nicopace I think (please @spiccinini comment on this) this file is not
>going to be run on the routers, indeed it is not going to be installed
>on the router. So in my opinion having it in Python is not a problem
>
>-- 
>You are receiving this because you were mentioned.
>https://github.com/libremesh/lime-packages/pull/466#discussion_r254784670

-- 

dangowrt:
@spiccinini I'm not a friend of dual-booting by changing bootcmd every time. What if that itself fails by interrupted write or any other kind of breakage of the uboot environment? Then (best case!) router will default to factory uboot environment not boot if the default value of bootcmd points to the wrong firmware. And once again there is single point of failure. Please reconsider that approach.

packages/safe-upgrade/bootscript_gen.py
nicopace: It makes it easier for other to maintain, as it is all in one single language.
Not mandatory... But the less language diversity, the easier to get the skills to maintain or contribute

spiccinini: Hi! Yes, this code is only for the develpment in case modification of the boot script is needed as it is easier to maintain in a readable form but u-boots needs it collapsed.

Yes I can port it to lua, although my confidence of the correctness will drop and as this is not central to the package is low priority for me.

spiccinini:
@dangowrt Hi! Bootcmd only change when you perform an update in 3 steps, the installation/testing step, the new boot (removing the testing state) and then at confirmation step.
And yes it can break if power is lost in any of this moments with the single u-boot env config that the bootloader is using. I plan to use dual env config in u-boot (I left some space prepared for that at the end of the Flash before the ART). Using dual-env if  the env is corrupted it will automaticaly  fall-back to the other env.

In any case it is much more probable that someone installs a bad firmware than power lost in the tiny window of env modifications. 

If you have other ideas on how to implement this, please share :)

spiccinini:
I plan to write a technical documentation on how all this works, but I couldn't find the time yet. If you want to understand, I think that you can build up a good idea of how this work reading the boot script in `bootscript_gen.py` as I tried to make it very simple and it has comments.

Issue #345, opened 2018-03-11
anygw: alfred lease sharing is broken


dangowrt:
Doesn't work with dnsmasq on current develop branch. Maybe alfred is broken or doesn't work well on non-batman-adv bridge ports, I never really saw it working, so I have a hard time figuring out the expected behaviour...

nicopace:
@dangowrt will get to check this later hopefully... it is a thing that has had issues in the past, and also a very useful feature!

@aparcar bmx and batman-adv are part of a solution that work together.
Despite LibreMesh is flexible enough by design to support different kinds of setups, the collective decision has been for the default to be a collaboration between batman-adv (layer2 mesh routing) and bmx6 (layer 3 mesh routing).

Implementing the functionality of bmx7-sms can be done the same way there is bmx6-sms, and the same that there is for alfred... but as of now, there has not been a decision to change the default, that is the one @dangowrt is talking about.

aparcar:
related https://github.com/libremesh/lime-packages/issues/369

dangowrt:
Just to add my two cents to this:
 * lease sharing is only required for layer-2 mesh, so using bmx6/7-sms for this doesn't make sense to me, because for layer-2 mesh we use batman-adv (and potentially could use 802.11s with `mesh_fwding=1`), so alfred or something which works on top of either Ethernet layer-2 or IPv6 link-local multicast (like alfred) is the right thing here imho.
 * replacing our current hacky solution with ddhcpd could be nice, it looks like it is made for exactly this purpose. I didn't have a deep look at it nor try it myself, but it looks like it's DHCP (v4) only and no support for DHCPv6 what-so-ever (and we need DHCPv6 to work on prefixes longer then `/64`, because SLAAC/RA only works for at `/64` networks.
 * fixing alfred also looks like a good option because it will allow to use the same tool also for distributed name resolution and other purposes (and already supports v4 and v6 and it's rather easy to hook this into any DHCP/DHCPv6/DNS server daemon). So hence my question: Is this just about a bug or do you see a fundamental design flaw in our currentl solution or when using alfred+odhcpd in future?

p4u:
Agree with @dangowrt just add that alfred is not a good solution for DNS since it only works on the collision domain network. For DNS bmx-sms seems a much better option (actually is already implemented on bmx7-mdns package).

[...]

PR #467, opened 2019-02-07
Add lime-no-proto collection

gmarcos87:
It creates a new collection in which it does not force the selection of any mesh protocol layer 2 or layer 3. Currently all collections contain batman and bmx6 as dependencies. This collection is meant to develop and test other protocol convinations.

### DIFF
--- a/packages/lime-basic/Makefile
+++ b/packages/lime-basic/Makefile
@@ -46,6 +46,20 @@ define Package/$(PKG_NAME)-no-ui
 	         +bmx6-auto-gw-mode +lime-hwd-ground-routing +smonit
 endef
 
+define Package/$(PKG_NAME)-no-proto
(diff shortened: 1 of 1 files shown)


aparcar:
isnt dnsmasq-lease-share broken af?

ilario:
I love the idea!
A few comments:

1. why didn't you include `lime-webui` dependency as in `lime-basic` package?
2. in the Makefile the package is named in three different ways:
`lime-basic-no-proto`, `lime-basic--no-proto`, `lime-basic-level2`
looks wrong, could you check? 

> isnt dnsmasq-lease-share broken af?

@aparcar let's keep that on #345

gmarcos87:
Perhaps it would be better to work the dependencies of the lime-package packages a little more rather than create another meta package. I'll close this pull-request for now and analyze the situation a little better.

PR #469, opened 2019-02-16
Add fft-eval again

gmarcos87:
Some time ago there were a couple of changes from fft-eval that were sent to upstream, after that the Makefile was broken and as it block the build of libremesh was removed. This pull request includes a Makefile that works.


ilario:
What about https://github.com/libremesh/FFT_eval can we delete that?

PR #465, opened 2019-02-04
remove libremap legacy

aparcar:
Signed-off-by: Paul Spooren <mail@aparcar.org>

PR #471, opened 2019-02-16
ubus-lime-location: remove libremap-agent

aparcar:
This causes trouble as libremap-agent seems outdated and is no longer available.

### DIFF
--- a/packages/ubus-lime-location/Makefile
+++ b/packages/ubus-lime-location/Makefile
@@ -13,7 +13,7 @@ define Package/$(PKG_NAME)
   MAINTAINER:=Marcos Gutierrez <gmarcos87@gmail.com>
   SUBMENU:=3. Applications
   TITLE:=Libremap ubus status module
-  DEPENDS:= +lua +libubox-lua +libubus-lua +libuci +libremap-agent
+  DEPENDS:= +lua +libubox-lua +libubus-lua +libuci
   PKGARCH:=all
 endef
 


p4u:
It makes lime-sdk and chef to crash, so IMO this is an urgent issue. When libremap-agnt is fixed we can add it back.

aparcar:
So please approve and merge this, the CI does the rest

gmarcos87:
Hi, if we do this we must resolve libremap-agent as soon as possible, since the images that will be cooked will be defective (at least in relation to libremap). Is there more details about what happens with libremap-agent?

gmarcos87:
https://github.com/libremesh/lime-packages/pull/465 why not revert this?

spiccinini:
I think that the problem is that lime-sdk and/or chef and/or the travis of lime-packages doesn't have the libremap feed installed and somehow this was working by chance.

So I think that we still want to depend on libremap-agent but fix the feeds

p4u:
Some times openwrt-sdk is not working as expcted, so if lime-full is selected all dependencies must be compiled, but thats not always the case. So yes, we should revert #465

aparcar:
is libremap actually used? the Website and all related repos looked dead

spiccinini:
maintained version is here https://github.com/libremesh/libremap-agent (at least for now until the original repo is active again)

aparcar:
merge it into lime-packages and add to ci auto build, then close this PR here.

aparcar:
no value to have plenty of repos which are neither maintained nor auto build

p4u:
Is this already fixed?

egon0:
i fixed the build error on lime-agent ( see libremesh/libremap-agent/pull/2 ).

aparcar:
https://github.com/libremesh/lime-packages/pull/474 this adds libremesh to our repo

egon0:
there is no need to make whole lime-stack unusable because of libremap-related problems when someone won't need the libremap-stuff. a map app is not needed to run a lime-based mesh network. also, the serverside is mostly unmaintained - the only point to keep the POSSIBILITY to install libremap-agent is the legacy case at some communities. there is no point in discussing and delaying this pr this much. libremap is defacto not needed to run a lime-based mesh-network. when there is no need, it shouldn't be a DEPENDENCY at any point imho. lime should be a framework, anything is possible, nothing that is not directly mesh related should not be a hard dependency.

gmarcos87:
Okay, I think there's a misunderstanding. The problem is not in libremap-agent. It's stable software and it works well. The problem appears when someone tried to remove it from libremesh (by a personal decision). The problem is not libremap-agent, the problem is how its removal was handled.
Blaming libremap-agent is outdated is a way to avoid the real causes and intentions.
When I asked for explanations about why it is considered outdated or what changes are required I never got any answers. I recommend going back with the changes that really broke chef or lime-sdk and closing this pull-request.

egon0:
The basic problem was a commit from 31.12.2018 in the libremap-agent-repo - I fixed the problem. But you are forcing a hard dependency on libremap-agent that is not necessary for a basic mesh network. There is simply no need for it. Also the initial issue that resulted in this pr is fixed, there is no need to have Libremap-agent as dependency whatsoever to run a lime based mesh network. Everybody who wants to use libremap and so it's agent in the firmware could simply add it while building their firmware. Currently we are all forced to have it as it is a fixed dependency.

egon0:
And all this pr is doing, is removing this dependency. So anybody who really needs libremap-agent is free to add it to their firmware builds.

gmarcos87:
I understand, but libremesh is not the minimum and basic firmeware for a mesh network. As this has many more extras on several sides and is not solved by removing everything. If we remove this dependency from here libremap-agent disappears from all libremesh since no other package has it as dependency.
If you really want the basics and do not have libremap-agent or other extras can **start from lime-basic-no-ui and add the packages you want.** 

That option always existed, no one is forcing you to use a package you don't want. Simply in **lime-full it has libremap-agent, because it is full, with all features**. If you want the basic start from the basic and not withdrawing from the full.

gmarcos87:
It would be different if you modify the ubus-lime-location code to detect whether or not libremap-agent is installed and act differently. So if it would be an optional dependency.  Currently this pull-request eliminates a dependency that the code needs without taking into account what it generates in the user experience.

aparcar:
> It would be different if you modify the ubus-lime-location code to detect whether or not libremap-agent is installed and act differently. That way libremap-agent would be an optional dependency. 

Could you do that? Chef is broken for days

gmarcos87:
It's not in my scope right now.I don't understand why this breaks chef anyway. Add the libremap feed and you're done.

ilario:
See #644

PR #476, opened 2019-02-22
Add shared-state to replace alfred

G10h4ck:
shared-state provide a way to keep maps of data in sync between hosts
in the network, the design is minimal and completely insecure, it aims
to be a quick and dirty replacement for alfred which tend to fail
unpredictably and flood our network to death with unnecessary traffic
when there is enough alfred master nodes


dangowrt:
Hm, I have mixed feelings about that because it requires a web-server to run on every node -- on 4MB flash devices in my local environment we have made the trade to not have a webserver but WPA3/SAE no 802.11s support using wolfssl instead (and currently rely in batman-adv DAT to keep DHCP more or less in sync, but that works only in limited cases and only for IPv4)

PR #477, opened 2019-02-23
safe-reboot via Ubus

gmarcos87:
# Safe reboot in ubus-lime-utils
Exposing safe-reboot methods via ubus is going to be useful to make changes from the ui safely.
It would be very interesting to have the timestamp of the next scheduled restart. Currently safe-reboot doesn't have how to get this information or I don't know where to extract it from.

| Path       | Procedure       | Signature                                                         | Description                                                                                                                                                                                                                                                                                                                           |
| ---------- | --------------- | ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| lime-utils | safe_reboot     | { "action": "status" }                                            | Get safe-reboot status                                                                                                                                                                                                                                                                                                                |
| lime-utils | safe_reboot     | { "action": "start", "value": {"wait": TIME, "fallback": TIME } } | After backing up /overlay/upper/etc, wait for TIME (value.wait) before reboot (Default: 5min). After boot, wait for TIME (value.fallback) before reverting /overlay/upper/etc from backup found in /overlay/upper/.etc.last-good.tgz (Default: 10min).<br> _TIME examples: 1hour 60min 60m 3600sec 3600 (all of them are equivalent)_ |
| lime-utils | safe_reboot     | { "action": "now" }                                               | Do not make /overlay/upper/etc backup; instead check that there's one already in place (/overlay/upper/.etc.last-good.tgz,then reboot and wait for fallback timeout.                                                                                                                                                                  |
| lime-utils | safe_reboot     | { "action": "cancel" }                                            | Remove /overlay/upper/.etc.last-good.tgz (useful after a successful reboot)                                                                                                                                                                                                                                                           |
| lime-utils | safe_reboot     | { "action": "discard" }                                           | Restores /overlay/upper/etc from /overlay/upper/.etc.last-good.tgz (useful to discard changes)

### DIFF
--- a/packages/ubus-lime-utils/README.md
+++ b/packages/ubus-lime-utils/README.md
@@ -1,9 +1,13 @@
-
 # Utils Libremesh ubus status module
 
-|Path   	|Procedure   	|Signature   	|Description
-|---  |---  |---  |---
-|lime.utils |get_cloud_nodes   	|{}   	| Get cloud nodes
+| Path       | Procedure       | Signature                                                         | Description                                                                                                                                                                                                                                                                                                                           |
+| ---------- | --------------- | ----------------------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| lime-utils | get_cloud_nodes | {}                                                                | Get cloud nodes                                                                                                                                                                                                                                                                                                                       |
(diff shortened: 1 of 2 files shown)


gmarcos87:
@nicoechaniz? @G10h4ck? Anybody?

PR #481, opened 2019-02-26
LimeApp updated to v0.2.0-alpha.5

gmarcos87:
The current version of LimeApp in lime-packages is outdated with the current state of development.
I managed to implement a **continuous integration between libremesh/lime-app and libremesh/lime-packages**. When I publish a release (via tag) it automatically creates a branch with the Makefile updated in lime-packages.
This version improves the handling of calls to UBUS to avoid congestion and the first screens of FirstBootWizard in case the ubus module is installed.
There are also several fixes and minor features like text2speach in the alignment screen.

### DIFF
--- a/packages/lime-app/Makefile
+++ b/packages/lime-app/Makefile
@@ -39,7 +39,7 @@ define Package/$(PKG_NAME)/install
 	$(INSTALL_DIR) $(1)/etc/uci-defaults
 	$(INSTALL_BIN) ./files/lime-app.defaults $(1)/etc/uci-defaults/90_lime-app
 	$(INSTALL_BIN) ./files/95-lime-app-rpc-acl $(1)/etc/uci-defaults/
-	@mkdir -p $(1)/usr/share/rpcd/acl.d || true
+	$(INSTALL_DIR) $(1)/usr/share/rpcd/acl.d || true
 endef
 
 define Package/$(PKG_NAME)/postinst
--- a/packages/lime-app/Makefile
+++ b/packages/lime-app/Makefile
@@ -1,4 +1,4 @@
-# 
+#
 # Copyright (C) Libremesh 2017
 #
 # This is free software, licensed under the GNU General Public License v3.
@@ -6,11 +6,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lime-app
-PKG_VERSION:=v0.1.0-alpha.4
+PKG_VERSION:=v0.2.0-alpha.5
 PKG_RELEASE:=1
 
 PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
-PKG_HASH:=582148b14ccde1eabb28cafae85d8cc44232a7e223ad8b0e352b7d0842c7cbb4
+PKG_HASH:=fe54150286f82f19fce29b064df2e64e9787495c532b7941be8257e0d909fd40
 PKG_SOURCE_URL:=https://github.com/libremesh/lime-app/releases/download/$(PKG_VERSION)
 
 include $(INCLUDE_DIR)/package.mk


packages/lime-app/Makefile
```diff
@@ -48,3 +48,53 @@ define Package/$(PKG_NAME)/postinst
...
 $(eval $(call BuildPackage,$(PKG_NAME)))
+#
+# Copyright (C) Libremesh 2017
+#
+# This is free software, licensed under the GNU General Public License v3.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lime-app
+PKG_VERSION:= 3ea6d26c7c789b0224d0cf9a9f500b7a93b34c9985277d61f798e35022412741
```
G10h4ck: this seems swapped with PKG_HASH

packages/lime-app/Makefile
```diff
@@ -48,3 +48,53 @@ define Package/$(PKG_NAME)/postinst
...
+PKG_RELEASE:=1
+
+PKG_SOURCE:=$(PKG_NAME)-$(PKG_VERSION).tar.gz
+PKG_HASH:=v0.2.0-alpha.5
```
G10h4ck: this seems swapped with PKG_VERSION

packages/lime-app/Makefile
```diff
@@ -48,3 +48,53 @@ define Package/$(PKG_NAME)/postinst
...
+define Build/Compile
+endef
+
+define Package/$(PKG_NAME)/install
+	$(INSTALL_DIR) $(1)/www/app/
+	$(CP) $(BUILD_DIR)/build/* $(1)/www/app/
+	$(INSTALL_DIR) $(1)/etc/uci-defaults
+	$(INSTALL_BIN) ./files/lime-app.defaults $(1)/etc/uci-defaults/90_lime-app
+	$(INSTALL_BIN) ./files/95-lime-app-rpc-acl $(1)/etc/uci-defaults/
+	@mkdir -p $(1)/usr/share/rpcd/acl.d || true
```
G10h4ck: Shouldn't this be $(INSTALL_DIR) ?

packages/lime-app/Makefile
gmarcos87: Good catch. That's why we do peer review.

packages/lime-app/Makefile
gmarcos87: INSTALL_BIN set chmod +x after copy?

ilario: @gmarcos87 yes, that's the difference between INSTALL_BIN and INSTALL_DATA

gmarcos87: Ok, those files are executable scripts so I think it's okay INSTALL_BIN

G10h4ck: > Shouldn't this be $(INSTALL_DIR) ?

I refer to this line
`@mkdir -p $(1)/usr/share/rpcd/acl.d || true`

packages/lime-app/Makefile
```diff
@@ -48,3 +48,53 @@ define Package/$(PKG_NAME)/postinst
...
 $(eval $(call BuildPackage,$(PKG_NAME)))
+#
+# Copyright (C) Libremesh 2017
+#
+# This is free software, licensed under the GNU General Public License v3.
+
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=lime-app
+PKG_VERSION:= v0.2.0-alpha.5
```
G10h4ck: I am not sure if Makefile ignore it, but I would remove the space between = and v

gmarcos87: solved

packages/lime-app/Makefile
gmarcos87: INSTALL_DIR will create the whole path if there is no intermediate directory?

spiccinini: INSTALL_DIR is defined in rules.mk as  `INSTALL_DIR:=install -d -m0755`. It creates all the intermediate directories (like mkdir -p)

PR #482, opened 2019-02-26
Add First Boot Wizard and ubus-lime-fbw packages

gmarcos87:
# first-boot-wizard
First Boot Wizar allows you to analyze nearby mesh nodes and extract the base configuration to be part of the network. Once copied the configurations the user can select in LimeApp which one to use to configure his node or if he wants to create a new network.

**How does it work?**

If when turning on the node detects that it does not have a configuration applied (/etc/first_run exist) it scans all the networks and selects the mesh and ad-hocs. Then switch to the channel of each network and try to make a wget via fe80 to extract the configuration. After doing the full scan it goes back to the previous configuration.
The downloaded lime-default files are saved in /tmp and are exposed via lime-app to the user.
There are still many things to improve and optimize, but this version meets the MVP requirements we had planned.

# ubus-lime-fbw
| Path     | Procedure       | Signature          | Description                               |
| -------- | --------------- | ------------------ | ----------------------------------------- |
| lime-fbw | status          | {}                 | Get FBW status (scanning, lock, disabled) |
| lime-fbw | create_network  | {"name": "string"} | Create network                            |
| lime-fbw | search_networks | {"scan": true      | false}                                    | Get all networks (true force rescan) |
| lime-fbw | set_network     | {"file": "string"} | Use one of the results                    |

### ubus -v list lime-fbw

```
''lime-fbw' @4c5b89e0
	"status":{}
	"create_network":{"name":"String"}
	"search_networks":{"scan":"Boolean"}
	"set_network":{"file":"String"}
```

### DIFF
--- a/packages/first-boot-wizard/files/bin/firstbootwizard
+++ b/packages/first-boot-wizard/files/bin/firstbootwizard
@@ -1,0 +1,6 @@
+#!/usr/bin/lua
+
+local fbw = require('firstbootwizard')
+
+print("[FBW] Scanning...")
+get_all_networks()
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -1,0 +1,364 @@
+#!/usr/bin/lua
+
+require "ubus"
+local json = require 'luci.json'
+local ft = require('firstbootwizard.functools')
+local iwinfo = require("iwinfo")
+local wireless = require("lime.wireless")
+local fs = require("nixio.fs")
+local uci = require("uci")
+local nixio = require "nixio"
+
+local conn = ubus.connect()
+if not conn then
+    error("Failed to connect to ubus")
+end
+
+local function execute(cmd)
+    local f = assert(io.popen(cmd, 'r'))
+    local s = assert(f:read('*a'))
+    f:close()
+    return s
+end
+
+local function eui64(mac)
+    local cmd = [[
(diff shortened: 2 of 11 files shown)


packages/first-boot-wizard/files/bin/firstbootwizard
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/lua
+
+local fbw = require('firstbootwizard')
+
+print("[FBW] Scanning...")
+get_all_networks()
```
nicopace: it doesn't seem that this command is useful for anything... is it?
if the get_all_networks stores the info somewhere for later use, should change the signature of the method.
Also, would add a print("[FBW] Scanning done") sentence

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -0,0 +1,364 @@
...
+end
+
+local function execute(cmd)
+    local f = assert(io.popen(cmd, 'r'))
+    local s = assert(f:read('*a'))
+    f:close()
+    return s
+end
+
+local function eui64(mac)
```
nicopace: A non issue really... but it would ve very interesting to get this one out into a library, because it is a very general funcitonality that can be used in many spaces

packages/first-boot-wizard/files/bin/firstbootwizard
gmarcos87: The get_all_networks method prints that information and more.

PR #488, opened 2019-03-14
Add stale github bot

gmarcos87:
# Automatic housekeeping for pull-request and issues
I was recently reviewing the issues and pull-request of lime-packages. It is a very arduous and repetitive task. It occurred to me to look for some tool to do that monitoring and to warn of issues or pull-request abandoned, since many of them can be important. I found that github has its own bot to perform this task: stale.
This pull request incorporates the configuration for this bot, the proposed operation is as follows:
- Alert about issues without activity for X days (one year, two months, whatever we want)
- If after that warning there is no activity after Y number of days the issue or pull request is closed and a particular tag (autoclose, wontfix, etc) is applied.
- The bot ignores the issues marked with the tags we want (release-blocker, bug, user history, etc).

I think it's a very good tool, if you agree we can use this pull-request to polish the options of days and tags.

## My proposal is:
- Days for the first alert: 60
- Subsequent days to close if there is no activity: 15
- Tags to ignore: release-blocker, bug, user history
- Closed Tag: closed-by-bot

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,58 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Number of days of inactivity before an Issue or Pull Request becomes stale
+daysUntilStale: 356
+
+# Number of days of inactivity before an Issue or Pull Request with the stale label is closed.
+# Set to false to disable. If disabled, issues still need to be closed manually, but will remain marked as stale.
+daysUntilClose: 30
+
+# Issues or Pull Requests with these labels will never be considered stale. Set to `[]` to disable
+exemptLabels:
+  - pinned
+  - security
+    release-blocker
+    "user story"
+
+# Set to true to ignore issues in a project (defaults to false)
+exemptProjects: false
+
+# Set to true to ignore issues in a milestone (defaults to false)
+exemptMilestones: false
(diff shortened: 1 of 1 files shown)


ilario:
This sounds like a bad idea to me.
There is **plenty** of stale issues that are still valid, we're just too lazy to discuss them. 
Auto-closing would just mask our careless management of the issues.

nicopace:
I agree with you @ilario that tons of issues are relevant and valid to discuss.
I guess relevancy is already being stated by not giving attention to them.
I guess also that the most relevant issues will be waken up again and again if they are relevant enough, and by tagging them we will not loose those that are relevant.
I support it.

gmarcos87:
@ilario I think you're right illegally that the antiquity of the issue is not a ground for invalidity. It can mean our momentary inability to give an answer or even a problem that must be thought out and discussed in depth. But leaving it open on page 3 of a list of 70 issues is also not functional. I think what this bot does is just to shake us up by reminding us that we are leaving things on the road that may be important to keep in mind. It doesn't close for old, it closes them for lack of reaction after reminding us of their existence.

ilario:
I'm still sceptical but ok, let's try it

PR #489, opened 2019-03-16
Fix tags list

gmarcos87:
Fix typos in stale bot configuration


ilario:
It did make sense to have release-blocker as an exception... Nevermind, not so important.

PR #490, opened 2019-03-18
yggdrasil: interface: ygg0 | After first boot, yggdrasil-network starts with interface ygg0

wfleurant:
Yggdrasil has interesting encryption & bonding features in it's networking implementation. The algorithm builds a spanning tree network topology.

Features similar to stream bonding (mptcp) or link aggregation (lagg)
- increases throughput between nodes with additional network links with neighbor
- switching-queues tolerant of interface flapping / obscene packet loss

### DIFF
--- a/packages/yggdrasil/Makefile
+++ b/packages/yggdrasil/Makefile
@@ -1,0 +1,109 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=yggdrasil
+PKG_RELEASE:=1
+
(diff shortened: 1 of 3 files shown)


aparcar:
@wfleurant Hi, thanks for your work! The LibreMesh build CI only support multiarch packages, aka no compiled code. Go however has different binaries per target. Therefore, we have to open this PR again openwrt/packages

aparcar:
Moved here https://github.com/openwrt/packages/pull/8452

PR #491, opened 2019-03-21
LimeApp updated to v0.2.0-alpha.9 - Fix api calls

gmarcos87:
Remove window.crypto. Chrome only allows its use in secure environments (https or localhost)
Instead I implement a simple form of hash that fulfills the function.


gmarcos87:
https://github.com/libremesh/lime-app/commit/bf2c98965e742fbc21d3e0e2ad8678e4fed2eb3c

PR #493, opened 2019-03-28
Fix adhoc and mesh id in FBW

gmarcos87:
Change name/%H in favor of name.%H

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -328,8 +328,8 @@ function apply_user_configs(configs)
     -- nixio.syslog("crit", "FBW apply_user_configs ssid "..ssid)
     uci_cursor:set("lime-defaults", 'wifi', 'ap_ssid', name)
     uci_cursor:set("lime-defaults", 'wifi', 'apname_ssid', name..'/%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', name..'.%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'ieee80211s_mesh_id', name)
+    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', 'LiMe.%H')
+    uci_cursor:set("lime-defaults", 'wifi', 'ieee80211s_mesh_id', 'LiMe')
     uci_cursor:commit("lime-defaults")
 
     -- Apply config and reboot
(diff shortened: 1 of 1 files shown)


packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -328,8 +328,8 @@ function apply_user_configs(configs)
     -- nixio.syslog("crit", "FBW apply_user_configs ssid "..ssid)
     uci_cursor:set("lime-defaults", 'wifi', 'ap_ssid', name)
     uci_cursor:set("lime-defaults", 'wifi', 'apname_ssid', name..'/%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', 'LiMe.'..name..'/%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'ieee80211s_mesh_id', 'LiMe.'..name..'/%H')
+    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', name..'.%H')
+    uci_cursor:set("lime-defaults", 'wifi', 'ieee80211s_mesh_id', name)
```
G10h4ck: this should be independent for the network name and constast accross all LibreMesh devices of al networks

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -328,8 +328,8 @@ function apply_user_configs(configs)
     -- nixio.syslog("crit", "FBW apply_user_configs ssid "..ssid)
     uci_cursor:set("lime-defaults", 'wifi', 'ap_ssid', name)
     uci_cursor:set("lime-defaults", 'wifi', 'apname_ssid', name..'/%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', 'LiMe.'..name..'/%H')
-    uci_cursor:set("lime-defaults", 'wifi', 'ieee80211s_mesh_id', 'LiMe.'..name..'/%H')
+    uci_cursor:set("lime-defaults", 'wifi', 'adhoc_ssid', name..'.%H')
```
G10h4ck: I would not include the name of the community network there as it may create confusion in client devices

gmarcos87: I have seen the usefulness of having the name of the node at the time of scans and identify nodes that have the apname and apname closed. This is the case of nodes that for some reason does not link to the mesh (bad signal, configuration problem, etc). Or to identify if a node belongs to a community or another. Does LiMe.hostname make more sense than community.hostname?

G10h4ck: Yeah I think it would generate less confusion as it is more different to the ap(name)_ssid, take care that having different ssid works only for adhoc, while ieee80211s_mesh_id must be the same on all libremesh nodes, otherwise it won't connect

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
gmarcos87: https://github.com/libremesh/lime-packages/pull/493/commits/3f50ca16dda23b02831cd2eec23cfa6e16b7e1a7

PR #503, opened 2019-04-20
use channels 157 and 48 for 5.8GHz

spiccinini:
channel 157 is FCC approved for outdoor use

Signed-off-by: Santiago Piccinini <spiccinini@altermundi.net>

TODO: test with devices with only one radio

### DIFF
--- a/packages/lime-system/files/etc/config/lime-defaults-factory
+++ b/packages/lime-system/files/etc/config/lime-defaults-factory
@@ -45,7 +45,8 @@ config lime network
 
 config lime wifi
 	option channel_2ghz '11'
-	option channel_5ghz '48'
+	list channel_5ghz '157'
+	list channel_5ghz '48'
 	option htmode_5ghz 'HT40'
 	option distance_2ghz '100'
 	option distance_5ghz '1000'


aparcar:
I guess it's better to have 48 as first entry to stay compatible with existing setups, not?

Regarding dual radios, I once wrote lime-smart-wifi, however would be better not be implemented in the core, not?

G10h4ck:
> I guess it's better to have 48 as first entry to stay compatible with existing setups, not?
> 
> Regarding dual radios, I once wrote lime-smart-wifi, however would be better not be implemented in the core, not?

because the first 5Ghz interfaces is usually named wlan1 it will get channel 48 by default

spiccinini:
> > I guess it's better to have 48 as first entry to stay compatible with existing setups, not?
> > Regarding dual radios, I once wrote lime-smart-wifi, however would be better not be implemented in the core, not?
> 
> because the first 5Ghz interfaces is usually named wlan1 it will get channel 48 by default

Thanks Gio for commenting on this. In my tests the first 5GHz interface get's the 48 channel.

aparcar:
so when using a dual 5GHz only device, it's wlan0 157 and wlan1 48?

G10h4ck:
> so when using a dual 5GHz only device, it's wlan0 157 and wlan1 48?

yes

Issue #101, opened 2017-03-31
Setup Wizard: setup hostname


nicopace:
A user wants to give a host name to a node that is setting up via an easy to use setup wizard.
This involves thinking about how to implement a wizard in LuCI.

ilario:
Maybe this is already implemented in lime-app? @gmarcos87

gmarcos87:
done in FBW #506

PR #506, opened 2019-04-21
[FBW] Avoid duplicated donwloads, create networks and get hostname

gmarcos87:
## Avoid duplicated downloads
Before it downloaded all the configurations from the associated devices each time it configured the network. Now it does it once per mac address.

## Get remote hostname
Obtains the remote hostname the moment it gets the configurations and saves the data in the file.

## Set username in create and join network functions
The user can input a hostname so that it changes the moment the new network configuration is set.


packages/ubus-lime-fbw/files/www/cgi-bin/hostname
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+echo "Status: 200"
+echo "Content-type: text/plain;"
+echo "Cache-control: max-age=0, no-cache;"
+echo "Access-Control-Allow-Origin: *"
+echo "Access-Control-Allow-Methods: GET;"
+echo "Access-Control-Allow-Headers: X-PINGOTHER, Content-Type;"
+echo
+echo $HOSTNAME
```
spiccinini: Just a thought, in the future if we start adding more of this we may refactor into something more maintainable than than bash/cgi.

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -219,22 +250,31 @@ function read_configs()
...
     utils.execute("/rom/etc/uci-defaults/91_lime-config")
-    utils.execute("rm /etc/first_run")
-    os.execute("(( /usr/bin/lime-config && /usr/bin/lime-apply && reboot 0<&- &>/dev/null &) &)")
-    return { configs = configs }
+    uci_cursor:set("lime", 'system', 'hostname', hostname)
+    uci_cursor:commit('lime')
+    -- Start sharing lime-defaults
+    -- Apply new configuration
+    os.execute("/usr/bin/lime-config")
+    os.execute("/usr/bin/lime-apply")
```
G10h4ck: Why is lime-apply needed if we reboot just after?

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -219,22 +250,31 @@ function read_configs()
...
+    -- Start sharing lime-defaults and reboot
+    share_defualts()
+    remove_lock_file()
+    os.execute("reboot 0")
```
G10h4ck: what the 0 after reboot does?

packages/ubus-lime-fbw/Makefile
```diff
@@ -25,6 +25,7 @@ define Package/$(PKG_NAME)/install
 	$(CP) ./files/* $(1)/
 	@chmod a+x $(1)/usr/libexec/daemon/lime-fbw
 	@chmod a+x $(1)/etc/init.d/ubus-lime-fbw
+  @chmod a+x $(1)/www/cgi-bin/hostname
```
G10h4ck: AFAIR  some makefile macro should be used instead con plain chmod to install executables @spiccinini do you remember if it is like that?

packages/ubus-lime-fbw/files/www/cgi-bin/hostname
G10h4ck: missing new line at the end of the file

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
```diff
@@ -153,4 +153,8 @@ function utils.append_network(dev)
     end
 end
 
+function utils.filter_alredy_scanned(hosts, results)
+    return ft.filter(function(host) return results[host.host] == nil end, hosts)
+end
+
 return utils
```
G10h4ck: missing new line at the end of file

packages/ubus-lime-fbw/Makefile
spiccinini: Yes, `INSTALL_BIN` macro does `chmod +x` (equivalent to `chdmo a+x`)

spiccinini: This package is using "the trick" using `CP ./files/*` so using `INSTALL_* ` macros goes against that practicity. I don't know if there is a preferred way. I see mixed uses in other packages (from openwrt and libremesh).

spiccinini:
This fixes #101  ?

packages/ubus-lime-fbw/files/www/cgi-bin/hostname
gmarcos87: Yeah, we can do a lot more about it. For now I would leave it at that and see if I could find a more general solution.

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
gmarcos87: done in 8dc3f5ee3d782a71d0149ba6cf2095b4d86a6f92

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
gmarcos87: fix typo in 50582c1

packages/ubus-lime-fbw/files/www/cgi-bin/hostname
gmarcos87: fixed in 42b9ba9c7cdc476f4e4c890b606adda409aed1ab

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
gmarcos87: fixed in 42b9ba9

gmarcos87:
yes @spiccinini close https://github.com/libremesh/lime-packages/issues/101

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
G10h4ck: Still I see this missing new line at EOF

gmarcos87: ready, Im thinking on adapting this script  https://gist.github.com/johnjohndoe/4024222 :)

PR #507, opened 2019-04-25
[pirania] shared-state-priania - Fixes

gmarcos87:
When we started testing the distributed system of vouchers we noticed that we needed to be able to edit the objects (expire their validity, add macs addresses to the list, etc).
Now we use the md5 hash of the voucher as a key, so if the voucher changes, both the old and the new status are emitted. Then in the pirania hook the different results of the same vauchers are reduced according to the corresponding criteria and builds the database.

It also adds a line to the cronlist that triggers the synchronization and some readability conerreciones proposed by @spiccinini


G10h4ck:
Seems fine to me, keep an eye on CPU usage running that each 30 seconds ;)

packages/shared-state-persist/files/usr/bin/shared-state-persist
```diff
@@ -28,7 +28,12 @@
 local nixio = require('nixio')
 local fs = nixio.fs
 
-local sharedStateDir = "/var/shared-state"
+local sharedStateFolder = '/shared-state
+local sharedStateDefaultLocation = "/var/shared-state"
+
+NEW_DEVICE = 0
```
G10h4ck: better make those locals too ?

PR #509, opened 2019-05-08
Add shared-state.options.candidates list

gmarcos87:
Allows you to specify additional IPs addresses using uci configurations.
Until we advance in a more fully functional scheme this option will allow us to maintain one or more masters or replicators. It is not necessary that they are routers, it is possible to program a http server for some more powerful device or destined to that end.
These ips are determined by a list in /etc/config/shared-state. 
A commented option is included in this pull request as an example.

### DIFF
--- a/packages/shared-state/files/etc/config/shared-state
+++ b/packages/shared-state/files/etc/config/shared-state
@@ -1,2 +1,2 @@
 config shared-state 'options'
-	#list candidates '10.13.0.2'
+	#list candidates 'http://10.13.0.2/cgi-bin/shared-state/'
--- a/packages/shared-state/files/etc/config/shared-state
+++ b/packages/shared-state/files/etc/config/shared-state
@@ -1,0 +1,2 @@
+config shared-state 'options'
+	#list candidates '10.13.0.2'
--- a/packages/shared-state/files/usr/bin/shared-state
+++ b/packages/shared-state/files/usr/bin/shared-state
@@ -21,6 +21,7 @@ local http = require("luci.httpclient")
 local JSON = require("luci.jsonc")
 local nixio = require("nixio")
 require("nixio.util")
+local uci = require("uci")
 
 sharedState = {}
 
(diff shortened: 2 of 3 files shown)


packages/shared-state/files/etc/config/shared-state
```diff
@@ -0,0 +1,2 @@
+config shared-state 'options'
+	#list candidates '10.13.0.2'
```
G10h4ck: I think complete URLs should be provided instead of IP, it is more comform with the command line option and more flexible. If the config is an URL it is much easier to put a cgi script into whatever web-service at whatever path of which the use have less degree of control, instead with the IP the web-service must behave more stringently like a shared-state daemon.

gmarcos87: You're right, I already made the change.

packages/shared-state/files/etc/config/shared-state
```diff
@@ -0,0 +1,2 @@
+config shared-state 'options'
+	#list candidates 'http://10.13.0.2/cgi-bin/shared-state/'
```
nicopace: shouldn't this be uncommented, and in a file called shared-state.example or something like that?
this is in the right place but with a config that is broken? or at least it looks like... so better to provide a sane default, or no default and expect it not to be there at all.

packages/shared-state/files/usr/bin/shared-state
```diff
@@ -83,6 +84,15 @@ function sharedState.sync(urls)
 	urls = urls or {}
 
 	if #urls < 1 then
+
+		local uci_cursor = uci:cursor()
+		local fixed_candidates = uci_cursor:get("shared-state", "options","candidates") or {}
+		for _, line in pairs(fixed_candidates) do
+			table.insert(
```
nicopace: could it be that this gets inserted every time that is called, and then the urls in the file get piled up?
first time it runs it is 1 time, 2nd time 2 times, and on and on?

packages/shared-state/files/etc/config/shared-state
G10h4ck: it is not broken, and empty list is the correct default

nicopace: good... thanks for the clarification @G10h4ck

packages/shared-state/files/usr/bin/shared-state
nicopace: looks like today i am a little bit distracted :)
thx gio

PR #511, opened 2019-05-10
Add babeld support to ubus-lime-metrics and last_internet_path

gmarcos87:
Currently lime-ubus tools are tied to bmx6, this pull request adds support for babeld detecting which of the two protocols is installed.

### DIFF
--- a/packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
+++ b/packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
@@ -1,4 +1,10 @@
 #!/bin/sh
-echo "adding cron to save last known path to the Internet"
-sed "\| \* \* \* .*internet_path*|d" -i /etc/crontabs/root
-echo "*/30 * * * * /usr/sbin/internet_path save" >>  /etc/crontabs/root
+
+unique_append()
+{
+	grep -qF "$1" "$2" || echo "$1" >> "$2"
+}
+
+unique_append \
+	"*/30 * * * * /usr/sbin/internet_path save"\
+	/etc/crontabs/root
--- a/packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
+++ b/packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
@@ -1,5 +1,4 @@
 #!/bin/sh
 echo "adding cron to save last known path to the Internet"
-sed "\| \* \* \* .*last_internet_path.*|d" -i /etc/crontabs/root
-echo "*/30 * * * * /usr/sbin/internet_path > /tmp/last_internet; if [ -s /tmp/last_internet ] && [\ "$(md5sum /tmp/last_internet | awk '{ print $1 }')\" != \"$(md5sum /etc/last_internet |  awk '{ print $1 }
-')\" ]; then cp /tmp/last_internet /etc/last_internet; fi" >>  /etc/crontabs/root
+sed "\| \* \* \* .*internet_path*|d" -i /etc/crontabs/root
+echo "*/30 * * * * /usr/sbin/internet_path save" >>  /etc/crontabs/root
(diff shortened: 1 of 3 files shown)


packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
```diff
@@ -1,5 +1,4 @@
 #!/bin/sh
 echo "adding cron to save last known path to the Internet"
-sed "\| \* \* \* .*last_internet_path.*|d" -i /etc/crontabs/root
-echo "*/30 * * * * /usr/sbin/internet_path > /tmp/last_internet; if [ -s /tmp/last_internet ] && [\ "$(md5sum /tmp/last_internet | awk '{ print $1 }')\" != \"$(md5sum /etc/last_internet |  awk '{ print $1 }
-')\" ]; then cp /tmp/last_internet /etc/last_internet; fi" >>  /etc/crontabs/root
+sed "\| \* \* \* .*internet_path*|d" -i /etc/crontabs/root
+echo "*/30 * * * * /usr/sbin/internet_path save" >>  /etc/crontabs/root
```
G10h4ck: Don't you prefere the unique_append idiom, lime in other packages?

packages/ubus-lime-metrics/files/usr/libexec/rpcd/lime-metrics
```diff
@@ -105,6 +126,21 @@ local function get_gateway(msg)
    printJson(_get_gateway())
 end
 
+local function get_last_internet_path(msg)
+    local path = list_from_file("/etc/last_internet_path")
```
G10h4ck: This information looks like quite volatile, and to store it in persistent flash memory cold cause flash breakage to many write cycles, is it necessary to store this information persistently?

packages/ubus-lime-metrics/files/usr/sbin/last_internet
```diff
@@ -2,15 +2,44 @@
...
-# if 8.8.8.8 is accessible, save internet path to file
-if ping -q -c5 -w10 8.8.8.8 &>/dev/null
-    then path=`mtr -6 -r -c 1 $gw.mesh  | grep "\.|" |  awk '{ print $2}' | cut -d'.' -f1`
-    printf "$path"
+if [ ! -z "$result" ] && [ "$1" = "save" ] ; then
+    newSum=$(echo $result | md5sum | cut -d' ' -f 1)
+    oldSum=$(md5sum /etc/last_internet_path | cut -d' ' -f 1)
+    if [ "$newSum" != "$oldSum" ]; then
+        echo -n -e "$result" > /etc/last_internet_path
+        echo 'Last path to internet updated'
```
G10h4ck: Probably we should monitor how often this happens, to determine how much flash write cycles are we burning with this

gmarcos87: In a stable network that path does not change regularly, we could avoid overwriting in another way: perform metrics and save only if they are better, or simply save only the shortest path if it is different from the previous one.

packages/ubus-lime-metrics/files/usr/libexec/rpcd/lime-metrics
gmarcos87: It is not really volatile information, it is information that must persist when the router restarts. Suppose that the user loses the Internet connection. The first thing they do is to restart the router, or if an electrical outage occurs in part of the network, etc. In order to be able to diagnose what happens with your usual exit to interent you need to be able to remember where you used to go.

packages/ubus-lime-metrics/files/etc/uci-defaults/99-save-last-known-path-to-internet
gmarcos87: Yeah, this package is older and I just fixed the script. I can adapt it to the unique_append style

G10h4ck: better use the unique_append (is this the exact name?) idiom then ;)

Issue #512, opened 2019-05-15
File collisions and dependency hell


dangowrt:
cooker failes complaining about both `dnsmasq-lease-share` and `shared-state-dnsmasq_leases` trying to install the same files:
```
 * check_data_file_clashes: Package dnsmasq-lease-share wants to install file /usr/src/lime-sdk/snapshot/ramips/mt7621/ib/build_dir/target-mipsel_24kc_musl/root-ramips/etc/uci-defaults/90_dnsmasq-lease-share
	But that file is already provided by package  * shared-state-dnsmasq_leases
 * check_data_file_clashes: Package dnsmasq-lease-share wants to install file /usr/src/lime-sdk/snapshot/ramips/mt7621/ib/build_dir/target-mipsel_24kc_musl/root-ramips/usr/bin/dnsmasq-lease-share.sh
	But that file is already provided by package  * shared-state-dnsmasq_leases
 * opkg_install_cmd: Cannot install package lime-full.
```

It wasn't that trivial for me to solve due to hard-dependencies introduced by the `lime-full` meta-package (which imho should be flattened-away and just use flavors instead, providing soft-dependencies instead).

ilario:
> [lime-full meta-package] should be flattened-away and just use flavors instead

Totally agree on that!!!

Just I have the doubt on what happens if we decide to support again the buildroot method (with `make menuconfig`) rather than lime-sdk: could we use flavors in that or we would need the meta-packages?

dangowrt:
@ilario we could easily provide a script to modify the buildroot's `.config` and select the packages of a flavor. The selection can then still be altered by the user using `menuconfig`.

ilario:
Like that!
In case people using buildroot wanted to keep the meta-packages, we could still remove them from the flavors (replacing lime-full with the complete list of the packages it should pull).

G10h4ck:
I use the build-root but never liked those glue-packages, I am all to have them removed, the laziness to not read a list of packages in the documentation and select them depending on one needs, is not a good reason to keep that maintaining hell cruft around IMHO

G10h4ck:
also `dnsmasq-lease-share` and `shared-state-dnsmasq_leases` should never be installed together, IMHO the latter is the one that works out of the two

ilario:
So, `lime-full` is gone. 
Do we want to eliminate also `lime-basic`?
For each package we eliminate, we will have to remember to fix what was referencing to it.

G10h4ck:
[...]

Issue #331, opened 2018-02-16
cascade of metapackages kills modularity


dangowrt:
I'm trying to use lime with something else than the usual combination of bmx6+batman-adv.
However, this is currently nearly impossible because of a cascade of meta-packages which pull-in a lot of packages via their dependencies. I tried crafting a flavor with bmx7-only and failed because the current approach of meta-packages and flavors was left behind in the past when we were still always building everything from source. When using the ImageBuilder, things like
```
DEPENDS:=+PACKAGE_lime-proto-bmx6:luci-app-bmx6
```
won't work because `CONFIG_PACKAGE_lime-proto-bmx6=m` no matter if the user actually intended to use bmx6 at all.

Hence I suggest to entirely drop all **optional** things from the lime-basic, lime-full, lime-mini, lime-webui, ... packages. Those packages can then be pulled in via the dependencies of a community-profile-package of a community which makes use of them. In that way, modularity can be restored and selecting lime-basic or lime-webui won't automatically pull-in potentially unwanted packages.
I'm not only talking about routing protocols and autogw stuff, but also things like lime-hwd-openwrt-wan which I went through difficult loops only in order to finally get rid of it.

What do you think?
@aparcar @p4u @nicoechaniz @altergui @G10h4ck 

If you agree, I'll start working on a PR to remove all protocol-specific or somehow optional dependencies from the meta-packages.

dangowrt:
Another example: In flavors.conf, many of the flavors remove dnsmasq. But lime-basic then pulls in dnsmasq-lease-share and dnsmasq-distributed-hosts which doesn't make sense without dnsmasq...

ilario:
On removing **unnecessary dependencies** from collections: +1
(in my opinion lime-sdk [flavors](/libremesh/lime-sdk/blob/master/flavors.conf) obsoletes lime-full-* collections)
As a related issue see #175 

---

The example you cited is a **conditional dependency**, do we want to remove them? They were useful when the recommended compilation method was [OpenWrt buildroot and menuconfig](/libremesh/lime-packages#using-lede-buildroot), now they're just annoying.

List of [conditional dependencies](https://wiki.openwrt.org/doc/devel/dependencies#dependency_types) in this repository:

```
lime-debug:		+PACKAGE_lime-proto-batadv:batctl
lime-map-agent:	+PACKAGE_lime-proto-bmx6:luci-lib-libremap-bmx6
lime-proto-anygw:	@(!PACKAGE_dnsmasq)
[...]

PR #513, opened 2019-05-18
remove lime-full meta package

spiccinini:
Remove lime-full as it is difficult to maintain (and to build consensus on what packages must be on lime-full). Some related issued: #512 #331


gmarcos87:
Shouldn't we remove the other metapackages too?

spiccinini:
> Shouldn't we remove the other metapackages too?

Probably but I don't know what will break nor what are the other meta-packages. I prefer that we discuss it as part of other PR

PR #515, opened 2019-05-30
[FBW] Add support for 5ghz multi-radio devices

gmarcos87:
Add FWB support for 5ghz multi-radio devices, like LibreRouter (with two 5ghz radios).
Changes include filtering according to device mac addresses and improved performance.

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
@@ -4,6 +4,7 @@ local ft = require('firstbootwizard.functools')
 local fs = require("nixio.fs")
 local iwinfo = require("iwinfo")
 local json = require("luci.json")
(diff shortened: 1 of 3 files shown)


packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard/utils.lua
```diff
@@ -132,6 +135,31 @@ function utils.filter_mesh(n)
...
+local find_network = function(network, data)
+    return ft.filter(function(net) return same_netwrok(net,network) end, data)
+end
+
+function utils.only_best(networks)
+    return ft.reduce(function(acc, network)
+        local existent = find_network(network, acc)
+        if #existent > 0 then
+            acc = ft.map(function(net)
+                if same_netwrok(net, network) and net.signal < network.signal then
```
nicopace: same_netwrok or same_network?

Issue #521, opened 2019-06-08
shared-state is not proposing a neighbour in one direction, but it is in the other


nicopace:
In QuintanaLibre, happens the following
```bash
nico@ingorian ~> ssh -A root@thisnode.info
root@ql-roxa:~# shared-state-get_candidates_neigh
fe80::96:66ff:fef1:5251%wlan1-adhoc_29
fe80::eade:27ff:fe91:606e%wlan1-adhoc_29
root@ql-roxa:~# ssh -A fe80::96:66ff:fef1:5251%wlan1-adhoc_29
root@ql-nicojesigioia:~# shared-state-get_candidates_neigh 
fe80::e1:d3ff:fe19:4e52%wlan2-adhoc_29
fe80::eade:27ff:fe91:606e%wlan1-adhoc_29
root@ql-nicojesigioia:~# ssh -A fe80::e1:d3ff:fe19:4e52%wlan2-adhoc_29
root@ql-oncelotes-bbone:~# exit
root@ql-nicojesigioia:~# ssh -A fe80::eade:27ff:fe91:606e%wlan1-adhoc_29
root@ql-gioiajesinico:~# exit
root@ql-nicojesigioia:~# exit
root@ql-roxa:~# shared-state-get_candidates_neigh 
fe80::96:66ff:fef1:5251%wlan1-adhoc_29
fe80::eade:27ff:fe91:606e%wlan1-adhoc_29
root@ql-roxa:~# ssh fe80::96:66ff:fef1:5251%wlan1-adhoc_29
root@ql-nicojesigioia:~# shared-state-get_candidates_neigh 
fe80::e1:d3ff:fe19:4e52%wlan2-adhoc_29
fe80::eade:27ff:fe91:606e%wlan1-adhoc_29
```

shared-state-get_candidates_neigh from roxa shows ql-nicojesigioia, but ql-nicojesigioia doesn't show roxa.

gmarcos87:
Whats are the fe80 addresses of ql-roxa?

nicopace:
don't know... but the mac addresses of it are:
```
root@ql-roxa:~# iw dev | grep addr
		addr 14:cc:20:ad:b0:6b
		addr 16:cc:20:ad:b0:6a
		addr 14:cc:20:ad:b0:6a
```

gmarcos87:
fe80::14cc:20ff:fead:b06a is filtred becose include ::1. Its a bug in shared-state-get_candidates_neigh

[...]

PR #523, opened 2019-06-09
Corrects own ip filtering

nicopace:
It was selecting too much, filtering out ip6 that partially matched with short ips.


nicopace:
Fixes #521

G10h4ck:
Seems ok, did you tested if all those special characters (^$*.) needs some escaping? Because as we need variable substitution there the are between "" not ''

G10h4ck:
Also next time please use a branch on a personal fork, instead of a branch on the main repository ;)

PR #524, opened 2019-06-09
Fix shared-state cgi-bin script /tmp/lua_* pollution

G10h4ck:
Due to the fact that uhttpd may interrupt the script for various reasons
and the incapability to handle signals in lua without requiring more
dependencies, the cgi-bin script could terminate witout cleaning up the
temporary files created.
The script should just check params and call shared-state.
So to reasonably solve this problem I just rewrote it in ASH that doesn't
require temporary files at all.

### DIFF
--- a/packages/shared-state/files/www/cgi-bin/shared-state
+++ b/packages/shared-state/files/www/cgi-bin/shared-state
@@ -1,65 +1,28 @@
-#!/usr/bin/lua
+#!/bin/sh
 
---! LibreMesh
---! Copyright (C) 2019  Marcos Gutierrez <gmarcos@altermundi.net>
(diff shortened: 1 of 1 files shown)


gmarcos87:
I thought lua_* files were generated by luci.httpclient (client side) and not uhttpd (server side)
If that were the case I think it would be best to get the json string via wget instead of the lua library. I don't think this pull request solves that.

G10h4ck:
>  If that were the case I think it would be best to get the json string via wget instead of the lua library. I don't think this pull request solves that.

The luci.httpclient bug is already workarounded in shared-state code, so this one was another source of tmp pollution, but not the only one ;)

PR #526, opened 2019-06-10
Fix some uci-defaults files

gmarcos87:
After updating QuintanaLibre we found some bugs in some uci-defaults, these are the patches for those bugs.


packages/ubus-lime-metrics/files/usr/sbin/last_internet
```diff
@@ -34,12 +34,13 @@ fi
...
+    echo -n -e "$result" > /tmp/last_internet_path
+    newSum=$(md5sum /tmp/last_internet_path | cut -d' ' -f 1)
     oldSum=$(md5sum /etc/last_internet_path | cut -d' ' -f 1)
     if [ "$newSum" != "$oldSum" ]; then
         echo -n -e "$result" > /etc/last_internet_path
-        echo 'Last path to internet updated'
+        logger 'Last path to internet updated'
     else 
-        echo 'Last path to internet has no changed'
+        logger 'Last path to internet has no changed'
```
G10h4ck: sed 's/has no/has not/'  ;)

gmarcos87: Im AFK right now. You can change it?

PR #535, opened 2019-06-17
New packages: auto-usb-wwan and soporteremoto

altergui:
quick development for Cheran network


altergui:
installing `soporteremoto` package will make the node install zerotier and join network `35c192ce9b5fa4cb` which will allow AlterMundi members to remotely access the node via SSH

altergui:
installing `soporteremoto` package will make the node install zerotier and join network `35c192ce9b5fa4cb` which will allow AlterMundi members to remotely access the node via SSH

altergui:
a minor note: currently, auto-usb-wwan will not properly configure a USB device that is connected on *first boot* (after flashing), because the hotplug.d runs before lime-config.
After reboot (and every boot afterwards), hotplug.d script will run again and properly set up the USB WWAN (without lime-config resetting wireless config)

gmarcos87:
If you do lime-config and wifi or lime-apply in the node that is connected the usb dongle you can be without network access? Will lime-config recognize it as a 2ghz interface and reconfigure it atuomatically according to the modes set in /etc/config/lime?

PR #537, opened 2019-06-20
Adds shared-state-nodes_and_links

gmarcos87:
This package allows to distribute via shared-state the location of the nodes, their macs and their wifi links with other macs. 
Modify ubus-lime-location to consume the information:
```
root@si-fridge:~# ubus call lime-location nodes_and_links
{
	"si-fridge": {
		"links: [
			"a0:f3:c1:6c:32:75",
			"c0:4a:00:fc:41:55"
			"14:cc:20:ad:b0:1d"
			"64:70:02:4e:cd:0b"
		],
		"coordinates: {
			"lon": "-64.39936101436615",
			"lat": "-31.799673850547368"
		},
		"macs": [
			"02:75:ba:dc:0d:51"
			"02:75:ba:dc:0d:52"
		],
		"hostname": "si-fridge"
	}
}
```

It is useful to show a map of the network in a decentralized way in lime-app

### DIFF
--- a/packages/shared-state-nodes_and_links/Makefile
+++ b/packages/shared-state-nodes_and_links/Makefile
@@ -1,0 +1,39 @@
+#
+# Copyright (C) 2019 Gioacchino Mazzurco <gio@altermundi.net>
+#
+# This is free software, licensed under the GNU Affero General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
+GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
+GIT_COMMIT_TSTAMP:=$(shell git log -n 1 --pretty=%at . )
+
+PKG_NAME:=shared-state-nodes_and_links
+PKG_VERSION=$(GIT_COMMIT_DATE)-$(GIT_COMMIT_TSTAMP)
+
+include $(INCLUDE_DIR)/package.mk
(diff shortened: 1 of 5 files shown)


gmarcos87:
When I published this, I realized a mistake @nicopace . There are macs that end with a point. Can you see that?

G10h4ck:
Why do we want this information of all nodes on all nodes? Should't we use libremap for it? Or there is some nice GUI in the way to show the topology of the network from each node?

gmarcos87:
Imagine that you lose connection to the Internet or that the network directly lacks it. LibreMap as a centralized solution and dependent on cloud services is not useful for such cases.
This is an example of how information is shown in lime-app. We also have the idea of making the path to the gateway different color.
photo_2019-06-18_11-45-36

gmarcos87:
One thing we must add is the ability to cache tiles with the totla extension of the network, so that leaflet can show the background of the map even when you go offline

G10h4ck:
Great job!

nicopace:
> When I published this, I realized a mistake @nicopace . There are macs that end with a point. Can you see that?

I can't reproduce.
Can you pssh this? `uci get lime.system.hostname; iw dev | grep addr | cut -d \  -f 2 | sort -u`.
with that we can detect who has the dot and refine the expression.

PR #541, opened 2019-06-22
Replaces .lan local domain with .thisnode.info fixes #540

### LINKED ISSUE #540: Local .lan domain not easy to use on mobile and on Chrome/Chromium
The currently in use local domain (FQDN) `.lan` works perfectly when using Firefox on desktop or on mobile.
As [reported by a user on lime-users mailing list](https://lists.libremesh.org/pipermail/lime-users/2019-May/001457.html), on most of the other browsers (I tested Chrome, Chromium, Opera, Samsung Internet Browser) either on desktop or on mobile, when typing `lime-ddeeff.lan` this string is searched on search engines rather than opened as an URL.

On desktop, Chrome/Chromium show this message, on mobile they don't and neither other browser do.

Screenshot from 2019-06-22 20-38-52

This is easy to workaround either adding a slash `lime-ddeeff.lan/` or specifying the protocol `http://lime-ddeeff.lan`, still this can be a problem for most of the users.

Two possible solutions have been proposed in the mailing list:

* use an accepted but abandoned top level domain (to be tested .bv .sj .iq .um .su .gb .gu .kp)
* use an existing and reliable domain (like `thisnode.info` which is owned by Altermundi community network and is already used for connecting to the closest network node) which would result in something like `lime-ddeeff.thisnode.info`

ilario:
See #540

### DIFF
--- a/packages/lime-system/files/etc/config/lime-defaults-factory
+++ b/packages/lime-system/files/etc/config/lime-defaults-factory
@@ -7,7 +7,7 @@
 
 config lime system
 	option hostname 'LiMe-%M4%M5%M6'
-	option domain 'lan'
+	option domain 'thisnode.info'
 	option keep_on_upgrade 'libremesh base-files-essential /etc/sysupgrade.conf'
 
 config lime network


nicopace:
Why replacing it?
would suggest to add instead of replace, both can be at the same time, .lan is short (but requires to write the whole http://site.lan/ url), and the other one is longer but proper without the http:// part, so we should leave both.

ilario:
> would suggest to add instead of replace, both can be at the same time

I didn't know it was possible. @nicopace how do you set it up in OpenWrt?

nicopace:
Don't know how... 

This is another approach: https://web.archive.org/web/20180722223228/https://iyware.com/dont-use-dev-for-development/
like the .localhost subdomain as it is saying is a local host (local for the network) , and is a valid domain!

ilario:
> .lan is short

Consider that even if we replace `.lan` with `.thisnode.info`, the nodes will be accessible BOTH with `http://lime-ddeeff.thisnode.info` and with `http://lime-ddeeff`
so, if someone prefers to write less, it can already avoid writing the domain name.

nicopace:
Well.. it looks it is not possible to define two domains (at least not in a fast search, because that "domain" key is also used to tell DHCP clients in which domain they are at.

> This is another approach: https://web.archive.org/web/20180722223228/https://iyware.com/dont-use-dev-for-development/

Based on that document that I shared earlier, to be standards compliant we can use the *.thisnode.info, or the *.localhost domains.

On the other hand, using *.thisnode.info allows us to capture when the users are using the domain outside their network (like through their mobile network when it exists), and help them through it...

PR #548, opened 2019-07-10
Have dnsmasq-lease-share.sh forward it's arguments to dnsmasq-lease-share.lua

jonesmz:
In my testing, I saw dnsmasq use a very large amount of CPU, and calling dnsmasq-leash-share.sh over and over again with the same collections of arguments.

I noticed that this script wasn't calling the lua file with any arguments.

Making this change resolves the dnsmasq process using lots of CPU.

Is there anything obvious that this change will break? I'm actually a little confused how DHCP was working properly in my mesh before making this change.


G10h4ck:
we are now used shared-state instead of alfred by default, so I suggest you to switch to shared-state-* to sync dnsmasq leases, anyway your contribution is welcome!

PR #550, opened 2019-07-13
Fix travis package build

spiccinini:
There is still a problem at the end of the build process when signing that I don't know how to fix. Here is a build example https://api.travis-ci.com/v3/job/215745040/log.txt

@aparcar what do you think about this? I've just adjusted some permissions and directories to the new images that you where developing


G10h4ck:
Seems a problem with extracting the private signing key. Maybe the password/key to do that is made available bu travisCI only if it is a branch of the main repository?

PR #551, opened 2019-07-25
WIP: Allow routing on top of AP-client links, keep general and specific wifi settings separate, fix #415

### LINKED ISSUE #415: Wifi station mode in backbone via routing protocols instead than br-lan

ilario:
LibreMesh by default creates 3 wireless interfaces on each radio and emploies the IEEE802.11s one for routing.
Up to now, access point-client wifi connections (AP-sta) was possible to use for routing with BMX6 as described in #426. 
For the next release, [BMX6 is not going to be included](https://github.com/LibreRouterOrg/openwrt/commit/46525b36892694d118447a42e945b4df7ad09eae) while [Babeld will be](https://github.com/LibreRouterOrg/openwrt/commit/d9be9446aa572efeb2514a30b29685aa9a3cf1d7).

The usage of AP-sta connections for the backbone can have the following advantages:
nearly every OpenWrt-supported router can be employed for a LibreMesh network; and the AP and client drivers are more tested than the IEEE802.11s drivers.

This PR allows the user to set up an AP interface (using interface specific configuration) out of the br-lan bridge removing the hardcoded setting `network` equal to `lan` from the [mode/ap.lua](https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/usr/lib/lua/lime/mode/ap.lua) and the [mode/apname.lua](https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/usr/lib/lua/lime/mode/apname.lua) scripts.
This implies that an `option ap_network 'lua'` and an `option apname_network 'lua'` will have to be explicitly set in the configuration for having a normal usage. This was added to `/etc/config/lime-defaults-factory`.

As implemented in eadcc14da3b5d72ae2446d9e62f044208074133f and in db1c3505ac39b178ed19cb0ecf939ce30e000032, the general wireless configuration is also used as a default for the specific interface configuration, making impossible (I think) to not set the `option network` for the specific interface configuration needed for the automatic setting of it by OpenWrt.
For this reason, this has been removed.
The use of specific configuration will not keep any setting from the generic configuration for this radio, so that some users could have to update their configurations.
This has been reflected in [lime-example](https://github.com/libremesh/lime-packages/blob/master/packages/lime-docs/files/lime-example) file.

In [lime-example](https://github.com/libremesh/lime-packages/blob/master/packages/lime-docs/files/lime-example), there is also an example of configuration for an AP with VLAN and Babeld + Batman-adv and also for a client with the same protocols.


ilario:
I thought that rather than using the `apname` I can create a new mode and proto `apbackbone`.

In that case I would not need to remove the hardcoded lan from mode/ap and mode/apname and neither I would need to remove tableMelt.

ilario:
Finally, clientbackbone is not useful and I rewrote the apbackbone part in #554, closing.

PR #552, opened 2019-07-25
Remove a repetition in lime-defaults-factory

ilario:


gmarcos87:
Hi @ilario, and how is mtu set by default in bmx6?

spiccinini:
@gmarcos87 It is repeated, it is in the lines 19 and 34 and ilario is removing the first ocurrence.

ilario:
@gmarcos87 as @spiccinini pointed out, the setting of `bmx6_mtu` is still present in the modified file, I just removed a line where it got set another time.

PR #557, opened 2019-08-11
bandwidth-test package for maximum available bandwidth towards the internet

ilario:
This Lua script downloads a list of files from the internet (it could also be used for measuring the local network speed, but for this case using netperf like in [lime-metrics](https://github.com/libremesh/lime-packages/blob/ce7a73038afb5de5b7709e9bbfc353c97c86913e/packages/ubus-lime-metrics/files/usr/libexec/rpcd/lime-metrics#L77) would be more proper).

In order to not take too long on slow networks, the downloads are interrupted after a specific time (the SINGLE_TEST_DURATION option). This has the drawback that on fast connections, the SINGLE_TEST_DURATION time has to be waited even if the file download has already finished.
In my opinion this is not a problem, but if you want I can try to change this.

The final speed output is taken as the median of a few measurements on different servers (by default 5, this is the NONZERO_TESTS option). The measurements that give zero speed are discarded. This happens quite often as Busybox wget cannot manage any kind of redirection.
It can also happen if a too short SINGLE_TEST_DURATION is employed (wget timeout is set to SINGLE_TEST_DURATION*0.75).

The list of files to download has been adapted from @p4u's [bmx7-auto-gw-bw-mode](https://github.com/libremesh/lime-packages/blob/master/packages/bmx7-auto-gw-bw-mode/files/usr/sbin/bmx7-auto-bw-test) package.

The parameters are documented in the package's help:

```
Usage: bandwidth-test [SINGLE_TEST_DURATION] [NONZERO_TESTS] [SERVERS_LIST]
Measures the maximum available bandwidth downloading a list of files.
The measurement will take approximately SINGLE_TEST_DURATION*NONZERO_TESTS seconds.
The download of each URL is attempted at most one time: multiple URLs should be provided.
The speed in B/s is printed to STDOUT.

SINGLE_TEST_DURATION  fixed duration of each download process,
                      if missing reads from UCI status-report (default 20)
NONZERO_TESTS         minimum number of successful downloads,
                      if missing reads from UCI status-report (default 5)
SERVERS_LIST          a space-separated list of files' URLs to download,
                      preferably large files.
                      When running with Busybox wget, has to include http://
                      and will likely fail with https://
                      if missing reads from UCI status-report
                      (defaults to a list of 10 MB files on various domains)
```

The arguments are optional but their ordering is fixed (the first argument is SINGLE_TEST_DURATION, the second is NONZERO_TESTS and from the third on they are the SERVERS_LIST).

The parameters can be provided as options from the command line.
If UCI is present in the system (e.g. when using this script in OpenWrt), it is used for accessing the parameters in `/etc/config/bandwidth-test`.
If no arguments are provided and UCI is missing or the configuration is empty, use the default hardcoded parameters.

### DIFF
--- a/libremesh.sdk.config
+++ b/libremesh.sdk.config
@@ -2,6 +2,7 @@
 # CONFIG_ALL_KMODS is not set
 # CONFIG_ALL is not set
 # CONFIG_SIGNED_PACKAGES is not set
+CONFIG_PACKAGE_bandwidth-test=m
 CONFIG_PACKAGE_dnsmasq=n
 CONFIG_PACKAGE_dnsmasq-dhcpv6=m
 CONFIG_PACKAGE_ALFRED_BATHOSTS=y
--- a/packages/bandwidth-test/Makefile
+++ b/packages/bandwidth-test/Makefile
@@ -16,7 +16,7 @@ include $(INCLUDE_DIR)/package.mk
 
 define Package/$(PKG_NAME)
   CATEGORY:=LiMe
-  TITLE:=Available internet bandwidth measurement
+  TITLE:=Internet download bandwidth measurement
   MAINTAINER:=Ilario Gelmetti <iochesonome@gmail.com>
   URL:=https://libremesh.org
   DEPENDS:=+pv
@@ -28,7 +28,8 @@ endef
 
 
 define Package/$(PKG_NAME)/description
-  	Measure the maximum available bandwidth towards the internet, downloading a list of files.
+  	Measure the maximum available bandwidth towards the internet, in the download direction.
+	Downloading a list of files via HTTP connections.
 endef
 
 define Build/Compile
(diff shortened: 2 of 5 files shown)


ilario:
An alternative approach would be to use netperf, I just realized that some public servers do exist:

> netperf.bufferbloat.net (also called netperf-east.bufferbloat.net)
> netperf-west.bufferbloat.net
> netperf-eu.bufferbloat.net

An advantage of bandwidth-test over netperf is that it used port 80, which is never filtered by shitty internet providers (e.g. see page 32 of eduroam policy [here](https://www.eduroam.org/wp-content/uploads/2016/05/GN3-12-192_eduroam-policy-service-definition_ver28_26072012.pdf)).

ilario:
> print a message more usable by other scritps without having to resort to cut and grep.

@gmarcos87 please note that the scripts parsable output is the only thing going to STDOUT while the human readable output is on STDERR. Is this enough to satisfy your comment?

gmarcos87:
Great

PR #563, opened 2019-08-23
[Unittesting P2] Add testing framework

spiccinini:
This adds the unit testing machinery and CI integration.

Should be applied after P1 (#562). 


spiccinini:
Ok, travis pipeline is failing because there is no test yet (because I splited the patch series). In part 3 #564 it should be working.

.travis.yml
```diff
@@ -3,52 +3,63 @@ sudo: true
...
+      stage: pkgbuild
+      name: "Build packages"
+      before_script:
+      - docker pull "$IMAGE_NAME" || true
+
+      script:
+      - |
+        cat <<EOF >> feeds.conf.default
+        src-git base https://git.openwrt.org/openwrt/openwrt.git;v18.06.1
+        src-git packages https://git.openwrt.org/feed/packages.git^35e0b737ab496f5b51e80079b0d8c9b442e223f5
```
nicopace: looks like this information should be bonded not to the test travisfile but to the branch you are working on somehow... right?

.travis.yml
```diff
@@ -3,52 +3,63 @@ sudo: true
...
+        src-git luci https://git.openwrt.org/project/luci.git^f64b1523447547032d5280fb0bcdde570f2ca913
+        src-git routing https://git.openwrt.org/feed/routing.git^1b9d1c419f0ecefda51922a7845ab2183d6acd76
+        src-git telephony https://git.openwrt.org/feed/telephony.git^b9d7b321d15a44c5abb9e5d43a4ec78abfd9031b
+        src-git libremesh https://github.com/libremesh/lime-packages.git;$TRAVIS_BRANCH
+        src-git libremap https://github.com/libremesh/libremap-agent.git
```
nicopace: no branch specified... shouldn't there be one?

.travis.yml
spiccinini: yes, but if you are testing other openwrt branch you can change that in the .travis.yml of that branch. Not that I think this is the best way to handle it...

Anyhow, I am not changing nothing of this, it's just indentation changes (because now it is inside a _stage_).

.travis.yml
spiccinini: Maybe, but I am not changing existing behaviour in this PR

gmarcos87: @spiccinini? This is the only question to answer before the merge. I suppose that if it is not explicit it takes the last commit of master.

spiccinini: Yes, it takes the last commit form master. But here the idea is not to change this, the only change in this part of the travis file is indentation

PR #564, opened 2019-08-24
[Unittesting P3] lime.config tests and utils

spiccinini:
Add testing utils and tests to lime.config. As like all the unittesting patch series this changes should not change behaviour.

Should be applied after P2 (#563).


tests/utils.lua
```diff
@@ -0,0 +1,43 @@
+local config = require 'lime.config'
+local libuci = require 'uci'
+
+local utils = {}
+
+utils.assert = assert
+
+UCI_CONFIG_FILES = {
```
nicopace: These whould be config files in a libremesh firmware, right?
must this list be comprenhensive? ilario added a config now for bandwidth test for example... should it be here? or it could be configurable so if you are testing a new module you could ask for that configfile?

spiccinini: Good question. The thing is that if uci does not find an empty config file then doing uci things related to that file does nothing (it does not add new configurations because as the empty file is not there it acts as the database does not exists). So if you want to test something that somehow uses that uci database then "you can't". 

So we need to create empty files for the parts we are testing. For now it is mostly `network`, `lime`, `lime-defaults`, `wireless`. But I think it is good to lower the problems of most testing uses to create all the files "we know". And yes, if a new uci database file is needed in the tests we must update it here, maybe I can add a comment above UCI_CONFIG_FILES to make it clear? Other way is to patch all the uci functions in the tests so it asserts with a message and explanation if uci could not find the file.... but it is very intrusive.

nicopace: It is ok how it is, just wanted to rise the concern to make you aware of it, not cause there was something concrete to do... thanks!

PR #565, opened 2019-08-24
[Unittesting P4] lime.network tests

spiccinini:
This adds lime.network testing and utilities/helpers. As like all the unittesting patch series this changes should not change actual behaviour.

Should be applied after P3 (#564 ).


spiccinini:
Ok, travis pipeline is failing because there is no test yet (because I splited the patch series)

PR #566, opened 2019-08-24
[Unittesting P5] fake iwinfo and wireless tests

spiccinini:
Should be applied after P4 (#565  ).

### DIFF
--- a/packages/lime-system/tests/test_lime_wireless.lua
+++ b/packages/lime-system/tests/test_lime_wireless.lua
@@ -1,0 +1,31 @@
+local wireless = require 'lime.wireless'
+local test_utils = require 'tests.utils'
+local iwinfo = require('iwinfo')
+
+local uci = nil
+
+describe('LiMe Wireless tests', function()
+
+    it('test get_mac for loopback', function()
+        iwinfo.fake.set_hwmodelist('phy0', { ["a"] = true, ["b"] = false, ["ac"] = false, ["g"] = false, ["n"] = true,})
+        assert.is_true(wireless.is5Ghz('phy0'))
+        iwinfo.fake.set_hwmodelist('phy0', { ["a"] = true, ["b"] = false, ["ac"] = false, ["g"] = false, ["n"] = false,})
+        assert.is_true(wireless.is5Ghz('phy0'))
+        iwinfo.fake.set_hwmodelist('phy0', { ["a"] = false, ["b"] = false, ["ac"] = true, ["g"] = false, ["n"] = true,})
+        assert.is_true(wireless.is5Ghz('phy0'))
+
+        iwinfo.fake.set_hwmodelist('phy0', { ["a"] = false, ["b"] = true, ["ac"] = false, ["g"] = false, ["n"] = true,})
+        assert.is_false(wireless.is5Ghz('phy0'))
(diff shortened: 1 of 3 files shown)


packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -111,8 +111,10 @@ function network.setup_rp_filter()
 	
 	sysctl_options = sysctl_options .. "net.ipv4.conf.default.rp_filter=2\nnet.ipv4.conf.all.rp_filter=2\n";
 	sysctl_file = io.open(sysctl_file_path, "w");
-	sysctl_file:write(sysctl_options);
-	sysctl_file:close();
+	if sysctl_file ~= nil then
```
nicopace: Shouldn't it do something if the file doesn't exist? like notify or something?

packages/lime-system/tests/test_lime_network.lua
```diff
@@ -76,6 +76,35 @@ describe('LiMe Network tests', function()
...
+
+        stub(network, "get_mac", function () return  {'00', '00', '00', '00', '00', '00'} end)
+        stub(network, "scandevices", function () return  {eth99={}} end)
+        stub(utils, "getBoardAsTable", function () return BOARD end)
+
+        test_utils.disable_asserts()
+        network.configure()
+        test_utils.enable_asserts()
+
+        assert.is.equal("1500", uci:get("network", "lan", "mtu"))
```
nicopace: Is this a default value? it is not self-documented, at least for me...

packages/lime-system/tests/test_lime_network.lua
```diff
@@ -76,6 +76,35 @@ describe('LiMe Network tests', function()
...
+        assert.is.equal("static", uci:get("network", "lan", "proto"))
```
nicopace: Is this a default value? it is not self-documented, at least for me...

packages/lime-system/files/usr/lib/lua/lime/network.lua
spiccinini: I wrote it this way because I think that this should only happen inside the unit testing environment (not having permissions to write to /etc) and we don't want to get a notification in this case. I can add a comment to clarify the intention as I think like you that it is unexpected to read that without an error report.

packages/lime-system/tests/test_lime_network.lua
spiccinini: It is hardcoded in `lime.proto.lan`: `22:	uci:set("network", "lan", "mtu", "1500")`.

My idea behind "testing this" was exactly to make it clear that we have a fixed value there, without judging if it is good or not. And to serve as example on how to test this things, not because I believe in `1500` ;)

packages/lime-system/files/usr/lib/lua/lime/network.lua
nicopace: The if statement looks a little strange, that's all.

packages/lime-system/tests/test_lime_network.lua
spiccinini: Yes, it is the default, also not documented.

PR #567, opened 2019-08-24
[Unittesting P6] Add fake device support and lime-config test 

spiccinini:
Adds fake devices support, starting with LibreRouter-v1. This devices is used to perform an integation test running the main part of lime-config with this fakes device

Should be applied after P5 (#566  ).


tests/devices/librerouter-v1/board.json
```diff
@@ -0,0 +1,53 @@
+{
```
nicopace: isn't this file available somewhere already? can find the same one in the /etc/board.json of the librerouters...

tests/devices/librerouter-v1/uci_config_network
```diff
@@ -0,0 +1,39 @@
+config interface 'loopback'
```
nicopace: I am curious if this is initially generated by whom...

tests/devices/librerouter-v1/board.json
nicopace: So we  can't get it from there for the tests then? (just thinking on deduplicating it because of the implications on maintenance...)

tests/test_iwinfo.lua
```diff
@@ -0,0 +1,144 @@
...
+        assert.is.equal(20, sta.rx_mhz)
+        assert.is.equal(40, sta.tx_mhz)
+        local assoclist = {['AA:BB:CC:DD:EE:FF'] = sta}
+        iwinfo.fake.set_assoclist('wlan1-apname', assoclist)
+        assert.are.same(assoclist, iwinfo.nl80211.assoclist('wlan1-apname'))
+
+    end)
+
+    it('test hwmodelist(radio_or_phy)', function()
+        local hwmodelist_n_2ghz = { ["a"] = false, ["b"] = true, ["ac"] = false, ["g"] = true, ["n"] = true,}
```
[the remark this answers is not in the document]
germanferrero: Great San. In respect to the rest of the PR series, I have read them all, but I don't feel like having the authority/knowledge to approve it.

tests/devices/librerouter-v1/board.json
spiccinini: I am not sure to be understanding your question. You mean from inside a LinbreRouter? Yes that is what I did (I copied the board.json file and the other two files, network and wireless) and that is what I think we should do for other devices too. I don't think there is an online database of this files so we don't have to do it. If we want to have this kind of integration test "with device emulation" then I don't think we have other options, or maybe I am not seeing it?

tests/test_iwinfo.lua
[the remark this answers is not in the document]
spiccinini: Thank you very much German! It's ok, let's wait other reviewers for the approve flag.

tests/devices/librerouter-v1/board.json
nicopace: I mean getting that information from where it comes instead of hardcoding it... 
don't worry, it is a lower priority thing. important to keep an eye on deduplication, but certainly not urgent

tests/test_iwinfo.lua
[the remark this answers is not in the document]
spiccinini: I've pushed a commit to #568 with improved names, thanks!

PR #568, opened 2019-08-26
[Unittesting P7] Add testing docs

spiccinini:


gmarcos87:
I think it is necessary to include more information in the Readme.MD.
Most users never get to CONTRIBUTING.md, they are often thinking that it is a matter of rules and norms of behavior rather than technical issues.

spiccinini:
> I think it is necessary to include more information in the Readme.MD.
> Most users never get to CONTRIBUTING.md, they are often thinking that it is a matter of rules and norms of behavior rather than technical issues.

Ok! I will add a paragraph to the Readme about the existence of the tests and the test docs

PR #570, opened 2019-09-02
Replaces the use of first_run with uci record.

### LINKED ISSUE #569: FirstBootWizard: restarts when you reflash
How to reproduce:
1. flash a router
2. do the firstbootwizard
3. flash again (without removing config)
4. access to http://thisnode.info

how should it be: you should be greeted with the lime-app

what happens: you are welcomed with the firstbootwizard

nicopace:
Fixes #569

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -5,8 +5,10 @@
 -- apply_file_config: Set lime-default and apply configurations
 -- apply_user_configs: Set a new mesh network
 -- check_scan_file: Return /tmp/scanning status
--- check_lock_file: Check /etc/first_run status
+-- check_lock: Check uci lime.first_run status
 -- read_configs: Return scan results
+
+local fbw = {}
 
 local json = require 'luci.json'
 local ft = require('firstbootwizard.functools')
@@ -203,7 +205,7 @@ function apply_file_config(file, hostname)
     uci_cursor:set("lime", "system","hostname", hostname)
     uci_cursor:commit("lime")
     -- Remove FBW lock file
-    remove_lock_file()
+    fbw.remove_lock()
     -- Apply new configuration
     os.execute("/usr/bin/lime-config")
     -- Start sharing lime-defaults and reboot
@@ -219,7 +221,7 @@ local function end_scan()
 end
 
 -- Read scan status
-function check_scan_file()
+function check_scan()
     local file = io.open("/tmp/scanning", "r")
     if(file == nil) then
         return nil
@@ -228,17 +230,15 @@ function check_scan_file()
 end
 
 -- Read scan lock file
-function check_lock_file()
-    local file = io.open("/etc/first_run", "r")
-    if(file == nil) then
-        return false
-    end
-    return true
+function fbw.check_lock()
+  uci_cursor = uci.cursor()
+  return uci_cursor:get("lime", "first_run") or true
 end
 
 -- Remove lock file
-function remove_lock_file()
-    utils.execute("rm /etc/first_run")
+function fbw.remove_lock()
+  uci_cursor = uci.cursor()
+  uci_cursor:set("lime", "first_run", true)
 end
 
 -- Get config from lime-default file
@@ -289,7 +289,7 @@ function apply_user_configs(configs, hostname)
     os.execute("/usr/bin/lime-config")
     -- Start sharing lime-defaults and reboot
     share_defualts()
-    remove_lock_file()
+    fbw.remove_lock()
     os.execute("reboot")
 end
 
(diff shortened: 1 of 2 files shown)


packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -228,17 +230,16 @@ function check_scan_file()
...
-function check_lock_file()
-    local file = io.open("/etc/first_run", "r")
-    if(file == nil) then
-        return false
-    end
-    return true
+function fbw.check_lock()
+  uci_cursor = uci.cursor()
+  first_run = uci_cursor:get("lime", "first_run") or true
+  print(first_run)
```
spiccinini: This should return `first_run`

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -228,17 +230,16 @@ function check_scan_file()
...
 end
 
 -- Remove lock file
-function remove_lock_file()
-    utils.execute("rm /etc/first_run")
+function fbw.remove_lock()
+  uci_cursor = uci.cursor()
+  uci_cursor:get("lime", "first_run", true)
```
spiccinini: This is weird, are you sure you sent your last changes?

spiccinini:
I have local changes of firstbootwizard conversion into a module that I did not post because I didn't want to make the unitteesting PR that big. But this can help you in this change as you will be able to do a unit test for this change ;) But then yes you will need that at least the first 3 unittesting PR get merged

nicopace:
the thing is that we need someone else to review your PRs as I have done it already :D
I will do the changes this time :)

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -289,7 +289,7 @@ function apply_user_configs(configs, hostname)
     os.execute("/usr/bin/lime-config")
     -- Start sharing lime-defaults and reboot
     share_defualts()
-    remove_lock_file()
+    remove_lock()
```
spiccinini: this should be fbw.remove_lock() ?

nicopace: Indeed!
So important are the UnitTests, right?

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -228,17 +230,15 @@ function check_scan_file()
...
+  uci_cursor = uci.cursor()
+  return uci_cursor:get("lime", "first_run") or true
 end
 
 -- Remove lock file
-function remove_lock_file()
-    utils.execute("rm /etc/first_run")
+function fbw.remove_lock()
+  uci_cursor = uci.cursor()
+  uci_cursor:set("lime", "first_run", true)
```
spiccinini: I am not so sure that a boolean value can be stored through uci. `lime.config` has a method to set and get boolean values.

nicopace: It can use boolean. check https://github.com/libremesh/lime-packages/blob/master/packages/lime-docs/files/lime-example

spiccinini: That is used through `lime.config` and uci reads it like a string not a boolean. 

The other thing is that I think there is something odd with the logic now. Removing the lock is seting "first_run" to true ? It is not very clear. I would use something like "fbw_configured=1" or something more closer to "not_first_run=1".

nicopace: when I do `uci set lime.first_run=true; uci commit` I get:
```
root@ql-roxa:~# grep first_run /etc/config/lime 
config true 'first_run'
```

When I read that value from lua:
```
# cat test.lua 
local uci = require("uci")
uci_cursor = uci.cursor()
first_run=uci_cursor:get("lime", "first_run")
print(first_run)
root@ql-roxa:~# lua test.lua 
true
```

gmarcos87:
Hi @nicopace, #571 was merged. Can you update your pull-request so there are no conflicts? You also have available the new LIbreMesh suite tests, in case you want to add any.

nicopace:
> Hi @nicopace, #571 was merged. Can you update your pull-request so there are no conflicts? You also have available the new LIbreMesh suite tests, in case you want to add any.

will do, thanks!

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
G10h4ck: that way you are creating a section of type "true", options must be something like lime.section_name.option_name  why aren't you using lime.config library? That is what should be use to edit /etc/config/lime

spiccinini:
Closing in favor of PR #639

PR #572, opened 2019-09-05
Add codcov env variable and badge

gmarcos87:
We recently enabled unitesting in LibreMesh. This small contribution incorporates a badge with the percentage of test coverage and the link to its analysis.
It also ignores the file luacov.stats.out that is generated when the tests run locally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 build
 *~
 *.kate-swp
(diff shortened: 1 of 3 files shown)


ilario:
This is greeeat! Congrats @spiccinini!!!

PR #574, opened 2019-09-06
[WIP]CI: switch to gitlab

aparcar:
To upload to repo.libremesh.org

Signing is still missing.


aparcar:
This does need some more work, but in theory this can be used now from all kinds of devices.

I'll setup something to allow package signing. If you're very much again the min.io approach please let me know.

aparcar:
All snapshot packages are now here http://downloads.libremesh.org/snapshots/packages/

spiccinini:
Hi! In which gitlab-ci instance  will this run? Our own gitlab-ci ? gitlab.com ?

spiccinini:
Also now that we have the unittesting in place a gci stage for this must be added.

aparcar:
gci? Sure lets add more stages but maybe not within this PR.

Runner is my own but coordinated via gitlab

spiccinini:
> gci? Sure lets add more stages but maybe not within this PR.

for gci i mean _gitlab ci_. Travis currently has two stages so you are removing one with this PR.

> 
> Runner is my own but coordinated via gitlab

So do you have a proposal on moving from travis to gitlab ci? Whats the rationale? Do we have issues with Travis that I am not aware? Moving away from github.com to gitlab.com? Runners will be maintained by us? I prefer gitlab-ci over travis but we had a working workflow so we should discuss this type of changes to achive consensus.

aparcar:
>  so you are removing one with this PR.

From what I understand my PR introduces a build and a deploy stage, how is that different from travis?

I currently work with GitLab in other projects as well therefore rather stick to a single workflow, instead of keeping all travis/circle/gitlab CI specialties in mind.

spiccinini:
> > so you are removing one with this PR.
> 
> From what I understand my PR introduces a build and a deploy stage, how is that different from travis?

In travis there are 3 steps: "unit tests", "build" and "deploy".

> I currently work with GitLab in other projects as well therefore rather stick to a single workflow, instead of keeping all travis/circle/gitlab CI specialties in mind.

I personally prefer gitlab/gitlab-ci over github/travis but what we had was working or at least from my point of view maybe you have another vision? If this help us doing a release good but plase share the big picture of what you have in mind. How relying in a runner of your own (or anyone else) is putting us in better position?

G10h4ck:
I prefer gitlab-ci too, isn't it possible to run the CI on a gitlab.com runner? Using gitlab.com infrastructure would unload maintenance from us. I have been using it with Doker in Docker runner setup for other things and I expect it to work also for libremesh CI

ilario:
Can we have an update on this? Is this still useful even if we merge #728?

spiccinini:
Closing as we are not moving in the very short term. Thanks @aparcar for the code. We can reopen or base the migration to gitlab on this branch.

PR #578, opened 2019-09-06
anygw push link MTU option also via RA

G10h4ck:
This should cause more devices to set the correct MTU on their interface, as it rely on RA options that are usually taken in account more rigorously by clients confronted to DHCPv4 options. So it should improve user experience in general.

### DIFF
--- a/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
+++ b/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
@@ -110,6 +110,7 @@ function anygw.configure(args)
 
 	local content = { }
 	table.insert(content, "enable-ra")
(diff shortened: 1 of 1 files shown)


gmarcos87:
Hi @G10h4ck  If you like, I can try it on the quintana network. What improvement should I expect and what can fail?

G10h4ck:
> Looks good but what do you think about using a constant for the MTU to deduplicate the use in the line 91?

nice suggestion

G10h4ck:
> Hi @G10h4ck If you like, I can try it on the quintana network. What improvement should I expect and what can fail?

So hopefully more device will apply the correct MTU on their interface. Sadly it didn't worked with my phone. What could go wrong is dnsmasq refusing to start because the configuration line is wrong, but it accepted it in my tests.

Cheers!

PR #584, opened 2019-09-09
Removes cotonete

nicopace:
Cotonete is in beta state and needs to be tested more in order to be in the repository.
this PR removes it for now, as I have moved it to https://github.com/nicopace/cotonete


gmarcos87:
Okay, I hope I can see cotonete again in this repo someday.

PR #586, opened 2019-09-09
Fix device tests

spiccinini:
Allow disabling or customizing the validation of the existance of the interfaces from the tests. Now mocking this in the tests instead of disabling the assertions globaly.

Tests are failing if run in a machine without `eth0` (not the case of travis nor my computer...)


gmarcos87:
Works!

PR #595, opened 2019-09-18
hwd-openwrt-wan check table existence before using it

ilario:
On devices in which WAN interface does not exist on the board, the indexing of a non-existing table makes lime-hwd-openwrt-wan to fail.

```
# lime-config 
Clearing wireless config...
Clearing network config...
Disabling odhcpd
sh: /etc/init.d/odhcpd: not found
Cleaning dnsmasq
Disabling 6relayd...
/usr/lib/lua/lime/hwd/openwrt_wan.lua:20: attempt to index field 'wan' (a nil value)
stack traceback:
	/usr/lib/lua/lime/config.lua:144: in function </usr/lib/lua/lime/config.lua:144>
	/usr/lib/lua/lime/hwd/openwrt_wan.lua:20: in function 'detect_hardware'
	/usr/lib/lua/lime/hardware_detection.lua:32: in function </usr/lib/lua/lime/hardware_detection.lua:27>
	[C]: in function 'xpcall'
	/usr/lib/lua/lime/config.lua:144: in function 'main'
	/usr/bin/lime-config:59: in main chunk
	[C]: ?
```

This could hypotetically happen also when looking for the LAN subtable in network.lua [here](https://github.com/libremesh/lime-packages/blob/3209f8b53da7401ed41e57fdb3f4b226a4c5273d/packages/lime-system/files/usr/lib/lua/lime/network.lua#L33) but I suppose that the LAN table will always exist in every router?
Or are there routers with only a WAN port or no ethetnet ports at all?


ilario:
I pushed also a commit for fixing `network.lua` for devices whose only ethernet interface was recognized as WAN.
I have no idea if such devices exist.

For the case of devices without any ethernet interfaces, I have no idea what to do with primary_interface.

ilario:
I removed the commit about `network.lua` and I'll create an issue for that.

G10h4ck:
also you should check for `board` and `board['network']` being not null

G10h4ck:
some devices have no ethernet ports at al so network could not exist

ilario:
Even `board`??

Is the approach (assign, if, assign, if, assign) I used good or do you suggest to use another one?

For devices with no ethernet ports, please comment also on #597

ilario:
I added the check for the `network` subtable.

@spiccinini do you think that `board` can also be `nil` and we should check for it?

spiccinini:
> Ok for `board['network']`, makes sense, but even for `board`??

In a good device there is always board. And I also think that network too, but it is not bad to check it

ilario:
> Good! Do you want to try adding a test something like this:

Many thanks @spiccinini !!!
I promise I'll try to understand this test :D
But I'll have time to work on LibreMesh maybe next week, likely at mid-October.

ilario:
Ah hem... I copied and pasted @spiccinini tests just adding the most trivial one for a device with no network table (I suppose this can happen if no ethernet interface is present).
The tests fail on my computer so I couldn't confirm it works.

PR #599, opened 2019-09-20
adding kmod-ebtables-ipv6 to anygw dependencies

ilario:
lime-proto-anygw is using ebtables functions (`--ip6-protocol`) which require a kernel module currently not included in its Makefile:

https://github.com/libremesh/lime-packages/blob/f5d680cf269f8ff109c39c639206e5283be9b52d/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua#L69-L74


G10h4ck:
good catch!

PR #600, opened 2019-09-20
Support Babeld without VLAN on ethernet interfaces inside br-lan

### LINKED ISSUE #596: Babeld without VLAN does not run on LAN ethernet ports
Blocks #581 

tl;dr without VLAN, Babeld currently cannot run on ethernet

When instructing Babeld to run on the plain interfaces, without any VLAN, it fails to run on LAN ethernet ports.
The reason is that they lack an IPv6 Link Local address, and they cannot have it as they are included in `br-lan` bridge.

So, just in the case of Babeld witout VLAN, we should run it directly over the br-lan bridge.

The problem is that Batman-adv interface is also into the bridge and to have hello packets of Babeld running into Batman-adv would completely break Babeld topology estimation.

So some firewall rules like the ones in lime-proto-bmx6 [here](https://github.com/libremesh/lime-packages/blob/master/packages/lime-proto-bmx6/src/bmx6.lua#L130-L134) will have to be used.

ilario:
Fix #596 
Part of the code has been copied and pasted [from BMX6](https://github.com/libremesh/lime-packages/blob/master/packages/lime-proto-bmx6/src/bmx6.lua#L117-L138) (which run on br-lan by default).

We want to support running Babeld without VLAN (`list protocols babeld:0`) because:

* it is not actually needed
* seems that [switch on some Mediatek routers does not like it](https://forum.openwrt.org/t/mediatek-and-vlan-802-1ad-on-ethernet/42346)

But Babeld cannot run directly on an interface (e.g. `eth0` or `eth0.1`) which is included into a bridge (e.g. `br-lan`) because it cannot have an IPv6 Link-Local.
What have that is the bridge itself, so Babeld has to run on the bridge.
But the bridge includes `bat0` (Batman-adv stuff) which assures that all the packets get through, distorting the way that Babeld has for determining the metric of a route (sending UDP packets and checking how many get lost).

Maybe the configurable option `babeld_over_batman` is not useful and we can consider that it is always false?

I am still testing this PR, please review but wait for merging :)

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -51,6 +51,7 @@ config lime network
 	option bmx7_over_batman false
 	option bmx7_pref_gw none			# Force bmx7 to use a specific gateway to Internet (hostname must be used as identifier)
 	option bmx7_wifi_rate_max 'auto'
+	option babeld_over_batman false			# When Babeld is run without VLAN (babeld:0), it runs on the bridge which includes Batman-adv's bat0, keeping this false avoids to have Babeld metrics distorted by Batman-adv
 	option anygw_mac 'aa:aa:aa:%N1:%N2:aa'		# Parametrizable with %Nn. Keep in mind that the ebtables rule will use a mask of ff:ff:ff:00:00:00 so br-lan will not forward anything coming in that matches the first 3 bytes of it's own anygw_mac (aa:aa:aa: by default)
 #	option autoap_enabled 0				# Requires lime-ap-watchping installed. If enabled AP SSID is changed to ERROR when network issues
 #	option autoap_hosts "8.8.8.8 141.1.1.1"		# Requires lime-ap-watchping installed. Hosts used to check if the network is working fine
--- a/packages/lime-proto-anygw/Makefile
+++ b/packages/lime-proto-anygw/Makefile
@@ -23,7 +23,7 @@ define Package/$(PKG_NAME)
 	URL:=http://libremesh.org
 	DEPENDS:=+dnsmasq-dhcpv6 @(!PACKAGE_dnsmasq) +ebtables +libuci-lua \
 		+lime-system +lua +kmod-ebtables +kmod-macvlan \
-		+shared-state +shared-state-dnsmasq_leases
+		+shared-state +shared-state-dnsmasq_leases +kmod-ebtables-ipv6
   PKGARCH:=all
 endef
 
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -119,18 +119,36 @@ function babeld.setup_interface(ifname, args)
 	--! ethernet interfaces, but the ones which are inside of the LAN bridge
 	--! (e.g. eth0 or eth0.1) cannot have an IPv6 Link-Local and Babeld needs it.
 	--! So Babeld has to run on the bridge interface br-lan
-	local addIPtoIf = true
 	local isIntoLAN = false
+	local addIPtoIf = true
 	for _,v in pairs(args["deviceProtos"]) do
 		if v == "lan" then
 			isIntoLAN = true
-			break
+		--! would be weird to add a static IP to the WAN interface
+		elseif v == "wan" then
+			addIPtoIf = false
+		end
+	end
+
+	if ifname:match("^wlan") then
+		--! currently (2019-10-12) mode-ap and mode-apname have an hardcoded
+		--! "option network lan" so they are always in the br-lan bridge
+		if ifname:match("^wlan.*ap$") or ifname:match("^wlan.*apname$") then
+			isIntoLAN = true
+
+		--! all the WLAN interfaces are ignored by proto-lan
+		--! so they are not in the bridge even if proto-lan is present
+		--! (except mode-ap and mode-apname as mentioned above)
+		else
+			isIntoLAN = false
 		end
 	end
 
 	if tonumber(vlanId) == 0 and isIntoLAN then
-		utils.log("Rather than "..ifname..", adding br-lan into Babeld interfaces")
+		utils.log("Rather than "..ifname..
+		  ", adding br-lan into Babeld interfaces")
 		ifname = "br-lan"
+		--! br-lan has already an IPv4, no need to add it
 		addIPtoIf = false
 	end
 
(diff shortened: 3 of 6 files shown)


packages/lime-proto-anygw/Makefile
```diff
@@ -23,7 +23,7 @@ define Package/$(PKG_NAME)
 	URL:=http://libremesh.org
 	DEPENDS:=+dnsmasq-dhcpv6 @(!PACKAGE_dnsmasq) +ebtables +libuci-lua \
 		+lime-system +lua +kmod-ebtables +kmod-macvlan \
-		+shared-state +shared-state-dnsmasq_leases
+		+shared-state +shared-state-dnsmasq_leases +kmod-ebtables-ipv6
```
G10h4ck: Didn't you fixed this in another PR? here it seems unrelated

packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
```diff
@@ -103,17 +103,54 @@ function babeld.setup_interface(ifname, args)
...
 
+	local addIPtoIf = true
+
+	--! If Babeld is without VLAN (vlanId is 0) it cannot run directly
+	--! on ethernet interfaces which are inside of a bridge (e.g. eth0 or eth0.1)
+	--! because they cannot have an IPv6 Link-Local, so Babeld has to run on 
+	--! the bridge interface br-lan
+	--! If Babeld's Hello packets run over Batman-adv (whose bat0 is also 
+	--! included in br-lan), the links will have a wrong quality metric,
+	--! so these hello on bat0 have to be filtered
```
G10h4ck: It is not that they have a wrong quality metric, it's that babeld will see all nodes as direct neighbours through batman

packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
```diff
@@ -103,17 +103,54 @@ function babeld.setup_interface(ifname, args)
...
+		local hasBatman = false
+		local babeldOverBatman = config.get_bool("network", "babeld_over_batman")
+		local hasLan = false
+		for _,protoArgs in pairs(config.get("network", "protocols")) do
+			local proto =  utils.split(protoArgs, network.protoParamsSeparator)[1]
+			if(proto == "lan") then hasLan = true
+			elseif(proto == "batadv") then hasBatman = true end
+		end
+	
+		if hasLan and ifname:match("^eth%d") then
```
G10h4ck: What happens with wan port? Or port manually configured with ground routing?

G10h4ck: Also consider to move this to configure method as it don't seems to need to be run for each interface, run it once seems enough

packages/lime-proto-anygw/Makefile
ilario: That kernel module is needed both by anygw (see #599) and Babeld in case of no VLAN and presence of Batman.

packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
ilario: mh... Isn't this the same? But ok, I'll change the text :)

G10h4ck: no it is not the same

packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
ilario: > What happens with wan port?

Woops... With this PR, WAN port is not anymore used by Babeld.
This was unintentional, sorry.
Anyway WAN port is used for connecting to non-LibreMesh networks, so Babeld is not needed there.
I am ok with fixing it if we want Babeld on WAN.
Rather than checking for lan protocol, I can check for the presence of the interface into the 

```
config interface 'lan'
	list ifname 'bat0'
	list ifname 'eth0.1'
```
table.

> Or port manually configured with ground routing?

Ports configured with specific configuration are not treated by the whole section.
Do you mean that even if ground routing ports have a specific configuration, they should trigger this fix anyway? Are them included into br-lan bridge?

> Also consider to move this to configure method as it don't seems to need to be run for each interface, run it once seems enough

Sounds good!
Currently it runs more than once even if running once would be ok.
Running more than once is not detrimental but I agree that it is a waste.
If I put this into the configure, it would be executed even if no ethernet lan port is present, but I suppose this is ok, right?

ilario: >>   Also consider to move this to configure method as it don't seems to need to be run for each interface, run it once seems enough

> If I put this into the configure, it would be executed even if no ethernet lan port is present, but I suppose this is ok, right?

Wait, the ebtables rule can be in the configuration part, but the `ifname = "br-lan"` stuff has to stay into the setup interface one...
This is a mess.

G10h4ck: >Ports configured with specific configuration are not treated by the whole section.
> Do you mean that even if ground routing ports have a specific configuration, they should trigger this fix anyway? Are them included into br-lan bridge?

It all depends on what kind of configuration they got, if they have babled specified as proto they get there too

G10h4ck: > This is a mess.

ilario: > > Ports configured with specific configuration are not treated by the whole section.

I was wrong, interfaces with specific configuration are treated by this section.

> > Do you mean that even if ground routing ports have a specific configuration, they should trigger this fix anyway? Are them included into br-lan bridge?

The answer to my question was: no, interfaces configured via ground routing should not be touched by this fix and they will never be included in br-lan.

> It all depends on what kind of configuration they got, if they have babled specified as proto they get there too

Interfaces managed by ground routing will never have VLAN ID 0, right? If this is correct, they will never trigger the fix and the problem is solved. Otherwise, how can I check if the interface is also configured for ground routing?

ilario:
I tested these modifications and it works. I think I addressed all the comments.
Now the complete list of protocols of an interface gets passed to each of the protocols scripts, so that the Babeld script can check the presence of the `lan` proto on the same interface.
This way, the WAN interface will not get touched and neither will the interfaces managed by ground routing.
The code grew to an exaggerated length, I am sorry.
One awful thing is the check for the kind of interface in the last part of the code, where I add a new IPv4 to the interface. Is there a reason and a better way to do it or is it just an OpenWrt bug? 
Please review again @G10h4ck

ilario:
@spiccinini could you advise about where to insert the MTU fix from #601?
I mean, if Babeld is running on br-lan and inside of it there are some ethernet interfaces, the br-lan MTU needs to be set or is it already ok?

spiccinini:
> @spiccinini could you advise about where to insert the MTU fix from #601?
> I mean, if Babeld is running on br-lan and inside of it there are some ethernet interfaces, the br-lan MTU needs to be set or is it already ok?

The fix from #601 should be enough, it lowers the MTU  only when using vlans (vid !=0)

packages/lime-docs/files/lime-example
```diff
@@ -51,6 +51,7 @@ config lime network
 	option bmx7_over_batman false
 	option bmx7_pref_gw none			# Force bmx7 to use a specific gateway to Internet (hostname must be used as identifier)
 	option bmx7_wifi_rate_max 'auto'
+	option babeld_over_batman false			# When Babeld is run without VLAN (babeld:0), it runs on the bridge which includes Batman-adv's bat0, keeping this false avoids to have Babeld metrics distorted by Batman-adv
```
spiccinini: @ilario When someone would want to change from `false` to `true` ? If there is not known useful use case I prefer not to add more options.

ilario: I agree with you, and I'll remove this.
I added this originally just because the same was present for BMX6 over Batman-adv

packages/lime-proto-anygw/Makefile
ilario: Gosh, now I noticed that this file was actually not pertaining to this PR, sorry.

ilario:
Replaced by #631

PR #601, opened 2019-09-20
proto.babel: fix eth MTU when using vlan

### LINKED ISSUE #580: droped packets from babel vlan iface
When using netperf from a node through the babel interface the throughput is almost 0. Ping does not work when using a size bigger than 1470. Using `mtu = 1496` seems to fix this issue.
To reproduce the problem:
With two LibreRouters. LR-A connected through WAN to "internet". LR-B connected to LR-A with LAN port to LAN port of LR-A. In this scenario pinging from LR-B to internet with ping -s SIZE 192.168.1.1 only works when SIZE <= 1470.

spiccinini:
This fixes #580 

Some of the other protocols have mtu configurations but not all for example olsr, bgp. So this is only fixing babel I don't know if we want to create a more general patch that always use MTU = 1496 when using ETH + VLAN.

I have tested this with two LibreRouters. LR-A connected through WAN to "internet". LR-B connected to LR-A with LAN port to LAN port of LR-A. In this scenario pinging from LR-B to internet with `ping -s SIZE 192.168.1.1` only works when `SIZE <= 1470`. With this fix size could be greater than 1470 without problem.

### DIFF
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -110,6 +110,10 @@ function babeld.setup_interface(ifname, args)
 
 	local uci = config.get_uci_cursor()
 
+	if(vlanId ~= 0 and ifname:match("^eth")) then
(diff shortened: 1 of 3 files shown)


aparcar:
Sorry can't judge this one...

gmarcos87:
Tested in quintana.libre and in labolsa.libre and works

PR #603, opened 2019-09-29
Simpler WLAN channel config

### LINKED ISSUE #602: List of 5 GHz channels and one or two radio devices
The current [implementation of the list of channels](https://github.com/libremesh/lime-packages/blob/69ed489fc771cbd3bd4e0047ba2c926e98304ce8/packages/lime-system/files/usr/lib/lua/lime/wireless.lua#L119-L122) (you can specify a list of wifi channels instead of a single channel, then each radio on the same band should receive a different channel, [currently this is active by default](https://github.com/libremesh/lime-packages/blob/69ed489fc771cbd3bd4e0047ba2c926e98304ce8/packages/lime-system/files/etc/config/lime-defaults-factory#L47-L48)) is troublesome when considering networks with mixed one and two radios devices.

Let's take an example using the current default configuration and two commonly employed routers:

* LibreMesh on TP-Link WDR3600: `wlan0` has channel 11 on 2.4 GHz band, `wlan1` has channel 48 on 5 GHz band
* LibreMesh on Ubiquiti NanoStation M5: `wlan0` has channel 157 (if country code is ok, see #594) on 5 GHz band

one would expect the two routers to connect each other using the 5 GHz radios, but this does not happen because of the channels.
Is it intentional? To me it looks like a bug.

spiccinini:
Fixes #602 . Also related to #503 
This is not backward compatible as the channel list must be reordered prior to upgrade. I am not very happy with this but the current config is odd so....  One way to do it backward compatible is to add a new option like simple_channel_sorting = 1 and use this code only when the option is present (or something like that) but I don't know as it will add more complexity.

What do you think?


ilario:
> What do you think?

I think that it looks amazing!!!
Thanks @spiccinini!!!

> This is not backward compatible 

In my opinion it's ok and we don't need to set the option.
This feature is up to now present just in the code under development (not in any release yet), so users should not be surprised if things change for good :)

Next week I'll have time to test this on my routers :)

gmarcos87:
Approved! I think we should make a documentation or recommendation on how to perform the migration between releases. They are no longer compatible by multiple aspects, not only the order of the channels.

PR #615, opened 2019-10-08
LimeApp updated to v0.2.0

gmarcos87:
Improvements in relation to the previous version:

- Eliminated duplication of hosts in the configuration screen
- Navigate between nodes more clearly
- Validation in the FBW form

### DIFF
--- a/packages/lime-app/Makefile
+++ b/packages/lime-app/Makefile
@@ -6,11 +6,11 @@
 include $(TOPDIR)/rules.mk
 
 PKG_NAME:=lime-app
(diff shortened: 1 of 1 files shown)


altergui:
tested this, thanks a million @gmarcos!!
now in FBW the "Choose a name for this node" when doing validation says "Only letters, numbers and underscores are allowed"
this is wrong, it should allow only letters, numbers and *dashes* (-) , not underscores (_)

altergui:
cannot comment on "Navigate between nodes more clearly" since i haven't noticed any difference. But to me this looks good enough to merge

PR #617, opened 2019-10-09
Update Pirania with major fixes

luandro:
- fixes not showing user's own phone
- shows device first in list
- other fixes


altergui:
this will include these two commits:
https://github.com/libremesh/pirania/commit/2ae5239b0a72ad9b3716355c4d517c5ed9a331e0
https://github.com/libremesh/pirania/commit/5fff8c3e64bdd3a25a612763fd9bde7fd0e2e6a4
they both look way too big to be reviewed by reading them, i.e. they mix up different unrelated changes. next time i suggest to split into individual commits (just by reading the commit message, i guess it should be at least 5 different commits), this would really help cherry-picking fixes.
anyway i'll simply compile the whole updated code, test the user experience, and review/accept this pull request based on that

nicopace:
The actual review should happen in the piranhia code, not here.
This should only be a version bump.
Will wait for @altergui to test, let's review the review process @luandro ... not to block you as a maintainer, more to help you get the code better fitted if it is of any use for you.

altergui:
tested with 3 devices (laptop, android phone, iphone), the user's own phone was shown at the bottom *most* of the times (and preselected in the box), notably some times didn't (particularly on the iphone, tried a few times, it was never shown). but already is better than before, so accepting this merge

nicopace:
> tested with 3 devices (laptop, android phone, iphone), the user's own phone was shown at the bottom _most_ of the times (and preselected in the box), notably some times didn't (particularly on the iphone, tried a few times, it was never shown). but already is better than before, so accepting this merge

The iphones have a thing with mac randomization... it could be related to it (also, quite difficult to use in mac based captive portals... another issue (or rather non-issue, as in rural communities iphones are extremely rare)).

PR #622, opened 2019-10-10
shared-state: get_candidates_neigh reduce ping interval to 0.2s

spiccinini:
This will speed up the process, even if it will be cached if there are too many neighbour it takes a lot of time, at least 4s per neighbour. Now it will take x5 less time

### DIFF
--- a/packages/shared-state/files/usr/bin/shared-state-get_candidates_neigh
+++ b/packages/shared-state/files/usr/bin/shared-state-get_candidates_neigh
@@ -30,22 +30,41 @@ lastRunFile="/tmp/shared-state-get_candidates_neigh.lastrun"
 lastRun=$(if [ -s $lastRunFile ]; then echo $(cat $lastRunFile); else  echo -9999; fi)
 currUptime="$(get_uptime)"
 
-[ "$(($currUptime - $lastRun))" -lt "90" ] &&
+[ "$(($currUptime - $lastRun))" -lt "30" ] &&
 {
 	cat "$cacheFile"
 	exit 0
 }
 
-candidateAddresses="$(
+batmanNonMeshCandidateAddresses="$(
 for iface in $(ls /sys/class/net/) ; do
 	ls /sys/class/net/${iface}/upper_bat? &> /dev/null || continue
 
-	ping6 -c 2 ff02::1%${iface} 2> /dev/null | \
+	echo "${iface}" | grep -E "mesh|adhoc" &> /dev/null && continue
+
+	ping6 -i 0.1 -c 2 ff02::1%${iface} 2> /dev/null | \
 		awk '{if ($3 == "from") print substr($4, 1, length($4)-1)'"\"%${iface}\""'}'
 done | sort -u -r)"
 
-candidateAddresses="$candidateAddresses
-$(ping6 -c 2 ff02::1%br-lan | \
+wifi_get_macs="
+wireless = require('lime.wireless')
+iwinfo = require('iwinfo')
+ip = require('luci.ip')
+
+for _, ifc in ipairs(wireless.mesh_ifaces()) do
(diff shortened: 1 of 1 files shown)


gmarcos87:
I ran the script on a pretty core node of quintanalibre (si-radio). The result varied in each run between 2, 5 and 6 nodes detected. Perhaps it is too short a time to achieve a consistent result.

spiccinini:
> I ran the script on a pretty core node of quintanalibre (si-radio). The result varied in each run between 2, 5 and 6 nodes detected. Perhaps it is too short a time to achieve a consistent result.

Maybe we can increse the time but reduce from 2 pings to 1?

gmarcos87:
I tried different combinations of values and I can never reproduce the same state. Maybe what we should do is remove the cache or reduce its duration, however shared-state-get_candidates_neigh now responds more quickly

spiccinini:
> I tried different combinations of values and I can never reproduce the same state. Maybe what we should do is remove the cache or reduce its duration, however shared-state-get_candidates_neigh now responds more quickly

This may be because hidden node node problems. As we are pinging to broadcast responses may be colliding (and lost) if not all the receiver nodes hear at each other.

Through the wireless interfaces it may be better to not ping but to use the `iwinfo.nl80211.assoclist` filtering only stations with less than N seconds of inactive time, and then convert the mac address to the fe80 ipv6 address.   What do you think @gmarcos87 @G10h4ck @nicopace ?

nicopace:
I like the strategy @spiccinini .
Also, having a clear strategy that is different for different types of links makes sense.
We need to by mindful though, that not always an antenna that is connected in mesh/adhoc mode is another libremesh, but can be a device that is bridging with a libremesh node through a cable... the ping is the most effective one (or a broadcast package like the mikrotik does).
Still, maybe we can assume there is someone, and then still ping to make sure it is there.

spiccinini:
Just as a reminder, luci.ip has the ability to convert from MAC to link local ipv6 and back:

```
> luci.ip.new("fe80::6666:b3ff:fe47:e1b9")
> print(addr:tomac())
64:66:B3:47:E1:B9
```
```
> ip = luci.ip.new("64:66:B3:47:E1:B9")          
> print(ip:tolinklocal())
fe80:51::31:6666:b3ff:fe47:e1b9
```

gmarcos87:
> Just as a reminder, luci.ip has the ability to convert from MAC to link local ipv6 and back:

source

Initially we used the associated list, but that leaves out the links via eth that are necessary to capture as well.

spiccinini:
> Initially we used the associated list, but that leaves out the links via eth that are necessary to capture as well.

I propose to use ping for ETH and the assoclist for wireless, what do you think?

gmarcos87:
@spiccinini its ok for me

gmarcos87:
Okay, this is gonna be impossible.
I'd better pull now and resume this task in the near future. It was worth a try

G10h4ck:
> > Initially we used the associated list, but that leaves out the links via eth that are necessary to capture as well.
> 
> I propose to use ping for ETH and the assoclist for wireless, what do you think?

I remember we considered this option when talking about shared-state in la quintana, but don't remember why we didn't implemented it in the end, probably it didn't covered some corner case, BTW as I can't remember I'd approve this and see what happens :p

G10h4ck:
Rereading the discussion I remembered that one of the cases not covered by using assoclist on wifi interfaces, is when on ONLY on one side a dumb wifi device is used to connect and bridge to another LiMe device via ethernet, I am not sure we have any of those cases in reality.

spiccinini:
Here is an attempt at implementing what we have been discusing in this thread.

G10h4ck (approved):
A ver como se porta en la quintana con esta cambio ;)

PR #624, opened 2019-10-11
Logging intead of print

spiccinini:
utils.log is prefered because we can control its output for example
disabling writing to stdout when testing.

I removed all the prints that polluted the testing output, so I am confident that this is not breaking anything.

### DIFF
--- a/packages/lime-hwd-openwrt-wan/files/usr/lib/lua/lime/hwd/openwrt_wan.lua
+++ b/packages/lime-hwd-openwrt-wan/files/usr/lib/lua/lime/hwd/openwrt_wan.lua
@@ -48,7 +48,7 @@ function openwrt_wan.detect_hardware()
 			config.set(openwrt_wan.sectionName, "linux_name", ifname)
 			config.end_batch()
 		else
-			print("No wan interface detected")
+			utils.log("No wan interface detected")
 		end
 	end
 end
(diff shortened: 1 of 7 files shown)


spiccinini:
> Looks good!
> As an additional note: we should discuss whether to replace `print` by `logger -s` in utils.log

Yes, I am against calling a process each time we want to log something. Sadly there is no exact replacement for `logger`.  `nixio.syslog` does not allow to set DAEMON mode (to set the name of the process and the pid number) but I prefer to use it as it is only a syscall (100x cheaper than calling a process)

ilario:
Sorry for the OT, moved this discussion on #626.

nicopace:
> > Looks good!
> > As an additional note: we should discuss whether to replace `print` by `logger -s` in utils.log
> 
> Yes, I am against calling a process each time we want to log something. Sadly there is no exact replacement for `logger`. `nixio.syslog` does not allow to set the facillity to DAEMON mode (to set the name of the process and the pid number) but I prefer to use it as it is only a syscall (100x cheaper than calling a process). We may write our own logging to syslog function as it is only a UNIX socket write to /dev/log but we dont have access to sockets from lua AFAIK

we should add luasocket for that ... we were thinking on adding it still for other things.
curious that we can't do that setting, but the tradeoff makes sens... let's see down the road if we need it or not.

Issue #620, opened 2019-10-09
Review network profiles and workarrounds specially from quintanalibre


spiccinini:
All the uci-defaults list is:
```
/rom/etc/uci-defaults/00_uhttpd_ubus
/rom/etc/uci-defaults/03_network-switchX-migration
/rom/etc/uci-defaults/03_network-vlan-migration
/rom/etc/uci-defaults/04_led_migration
/rom/etc/uci-defaults/09_fix-checksum
/rom/etc/uci-defaults/09_fix-seama-header
/rom/etc/uci-defaults/10-fstab
/rom/etc/uci-defaults/10_migrate-shadow
/rom/etc/uci-defaults/12_network-generate-ula
/rom/etc/uci-defaults/13_fix_group_user
/rom/etc/uci-defaults/20_migrate-feeds
/rom/etc/uci-defaults/30_uboot-envtools
/rom/etc/uci-defaults/50-set-automount
/rom/etc/uci-defaults/80_libremap-agent
/rom/etc/uci-defaults/88_lime-merge-factory-defaults
/rom/etc/uci-defaults/89_lime-release
/rom/etc/uci-defaults/90-pirania-cron
/rom/etc/uci-defaults/90_dnsmasq-lease-share
/rom/etc/uci-defaults/90_lime-app
/rom/etc/uci-defaults/90_lime-banner
/rom/etc/uci-defaults/90_lime-customize-profile
/rom/etc/uci-defaults/90_watchping
/rom/etc/uci-defaults/91-disable_passwd_login
/rom/etc/uci-defaults/91-set_libremap
/rom/etc/uci-defaults/91-set_system_opts
/rom/etc/uci-defaults/91_lime-config
/rom/etc/uci-defaults/92-cron-workaround-tasks
/rom/etc/uci-defaults/92_add-lime-repos
/rom/etc/uci-defaults/94-dnsmasq_startup_workaround
/rom/etc/uci-defaults/95-lime-app-rpc-acl
/rom/etc/uci-defaults/96-minido-info
/rom/etc/uci-defaults/96-thisnode-info-translation
/rom/etc/uci-defaults/97-dropbear
/rom/etc/uci-defaults/98-disable-fbw-and-pirania
/rom/etc/uci-defaults/99-restart-babeld-workaround
/rom/etc/uci-defaults/99-save-last-known-path-to-internet
```

spiccinini:
[...]

Issue #610, opened 2019-10-04
if the kernel needs to kill a process, trigger a reboot


suarezwilliamf:
it may happen that because the device is running out of memory the kernel, using omkiller, may want to kill some processes.
the libremesh experience depends on many processes, so we should monitor their uptime or reboot them directly.
(this is @nicopace by the way :) )

nicopace:
The reason why it is a reboot and not a service reboot, is that we don't know which service is the one that is eating the ram... and at that stage, we are late already for fixing it.
Also, we should think what happens if we reboot too many times... maybe the router could enter to a failsafe mode automatically... like at uboot... @spiccinini, what do you think ?

ilario:
We could implement something like: 
if (uptime > 24h and free RAM < 2 MB) then reboot

But if it is a process filling the RAM, and not the tmpfs (/tmp), the OOM killer should work, I saw it acting just once and it killed LuCI which was the correct process to kill...

ilario:
No solid solution and no interest in the last 4 years, closing

PR #625, opened 2019-10-12
Add deferable reboot package

spiccinini:
Provides a deferable reboot mechanism. The system will be rebooted when the uptime of the system reach a specified amount of seconds and if /tmp/deferable-reboot.defer file does not contain a number that is greater than the current uptime.

This way if anyone wants to postpone the reboot, the way is to store the next checkpoint in seconds. For example if the system must not reboot from now and up to 120 seconds in the future then something like this can be donde with `awk '{print $1 + 120}' /proc/uptime > /tmp/deferable-reboot.defer`

Fixes part of #620 

Also adds `lime.utils.uptime_s()` and refactors ubus-lime-utils to use it.


packages/deferable-reboot/Makefile
```diff
@@ -0,0 +1,54 @@
...
+#
+# This is free software, licensed under the GNU General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
+GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
+GIT_COMMIT_TSTAMP:=$(shell git log -n 1 --pretty=%at . )
+
+PKG_NAME:=safe-upgrade
```
nicopace: change package name to deferable-reboot

packages/deferable-reboot/tests/test_deferable_reboot.lua
```diff
@@ -0,0 +1,69 @@
...
+
+describe('Deferable Reboot tests #defreboot', function()
+
+    it('test load config from lime.config default value', function()
+        defreboot.config()
+        assert.is.equal(defreboot.DEFAULT_REBOOT_UPTIME, defreboot.min_uptime)
+    end)
+
+    it('test load config from lime.config', function()
+        config.set('system', 'lime')
```
nicopace: maybe use its own config file?

spiccinini: I thought about that, the convention is to use it's own config file?

packages/deferable-reboot/Makefile
spiccinini: Thanks!

spiccinini:
> not clear from the readme what are we going to do with it, but seems fine. :)
> an example of for what using it would be useful.

The idea is to have something that performs a reboot at each node, depending only on the uptime. This way nodes that hung for some reason will be rebooted automaticaly. 
This reboot can be deferred if the node has knowledage that "everything is fine", this is not provided by this script, just the ability to postpone the reboot.

packages/deferable-reboot/tests/test_deferable_reboot.lua
ilario: I also think it's better to keep the configuration file separated, so that the package will have a chance to be merged into OpenWrt repository becoming independent from LibreMesh.

ilario:
This is good!
And could also mitigate the filling of the RAM described in #610. 

In the alert message, could you add some instructions about how to defer or stop the planned reboot?

During my GSoC I started developing `lime-scheduler` for running stuff at the peak time or at the night time (based on the number of the hosts connected to the network) and it would be good if we could use it for ensuring that the deferable-reboot is activated during the night time.

lime-scheduler is not ready yet (sorry for this) but the interaction I think would make sense is:

* remove the cron line or disable the execution via cron using a configuration line and `lime-scheduler` which plans the execution of `deferable-reboot` (using the `at` command) at the night time
* or, alternatively: `lime-scheduler` defers the reboot by as many seconds as it takes to be in the night time

The second option seems easier and does not require any change in your code :)

packages/deferable-reboot/tests/test_deferable_reboot.lua
nicopace: > I thought about that, the convention is to use it's own config file?

it looks like this package is generic enough to be upstreamed at any point in the future... 
There is no written convention AFAIK, but my rule of finger would be to put in lime things tied to the lime ecosystem, and outside of it those that are more generic.

ilario:
I just realized that rebooting more than once per day the router would make impossible for `tests-scheduler` to accumulate enough data for deciding when is the peak time and when is the night time (it currently stores its data in /tmp)...
@gmarcos87 do you think that I should copy the `tests-scheduler` load data to the flash memory once per day so that it can be kept over reboots? Originally I didn't do it for avoiding wear of the flash memory.
If you want to have a look, the very dirty code of tests-scheduler is here: https://github.com/ilario/lime-packages/tree/lime-tests-scheduler2/packages/tests-scheduler and I am testing it in these days.

gmarcos87:
@ilario I think that writing once a day flash is not inconvenient. If you calculate the useful life of the others router components writing the flash once a day is not a real problem.

packages/deferable-reboot/tests/test_deferable_reboot.lua
gmarcos87: I think if a configuration is not relevant to communities it may be in its own configuration file. On the other hand, if it's a configuration that makes sense for customizing a libremesh network it should be included within the lime configuration.

gmarcos87:
When we make the diagnostic scripts we must avoid that they put the network in a reboot loop. For example, in Quintana we check the connectivity to trigger a reboot, and after the network was restarted for some reason, some time it didn't connect to the internet before the script ran and everything was restarted again.

packages/deferable-reboot/tests/test_deferable_reboot.lua
nicopace: This is a tricky one. Though I agree, I wonder if it should be in both places (as we do with dhcp, network and others). This might be the case so we can have the chance to update it to an openwrt package...

ilario:
> @ilario I think that writing once a day flash is not inconvenient. If you calculate the useful life of the others router components writing the flash once a day is not a real problem.

Ok, problem solved using an PROCD service: at the shutdown the script copies the data from `/tmp` to `/etc/tests-scheduler/` and on boot it copies the data back to `/tmp` for being used and updated.

packages/deferable-reboot/tests/test_deferable_reboot.lua
spiccinini: Yes, probably for upstreaming it would be best to use it's own config file and also to read from lime.config (but as lime.config will not exists then we will need to add conditional code if lime is not installed)  I don't know how relevant would be for a community to change this. 
Anyhow I am adding this support

nicopace: Maybe we can do the same as we are doing with the rest, for lime-config to write the /etc/config/deferreboot file based on the config on /etc/config/lime if it is needed...

spiccinini: Hm. I don't see how doing this helps upstreaming. If we add code to the same package that has to know about lime.config to do that then I don't know it has to be like in other package lime lime-deferable-reboot.... To much hassle for this simple code? 

Maaaybe we can have an automatic way of configuring other config files from lime config based on the name of the option? Probably too obscure...

nicopace: The way I was proposing is the same way as it is being done with the rest of the config: for /etc/config/lime to be used by lime-config to generate the config files of /etc/config/deferable-reboot , and for deferable-reboot not to have to know the existence of lime or not.

would that work?

spiccinini: Yes, for this we should be doing another package or where do you think that the code that writes this config should be? Not in lime-system as this will tie lime with deferable-reboot. I think something like lime-deferable-reboot and that somehow by a plugin system this config is read from lime and writen to that file. That is what I tried to say in my last comment. And probably too complex system for a tiny peace of code. 

I think I will revert my last commit and just use the lime.config way of configuring it and decide how to untie it when, if ever, we try to upstream it. what do you think?

PR #628, opened 2019-10-17
wifi-unstuck-wa: add new package with radio workarrounds

### LINKED ISSUE #620: Review network profiles and workarrounds specially from quintanalibre

spiccinini:
A collection of workarrounds to radio and/or radio driver bugs.  At this moment it is very simple and only performs periodic scans at two frequencies of radios that have mesh or adhoc interfaces.

Partially fixes #620


spiccinini:
> If I got it right, this does a scan every 10 minutes, right?

Yes

> Wouldn't it be better to use [cotonete](https://github.com/nicopace/cotonete) for detection of the problem and make it run this script as a fix?

Yes! But lets have something that works until we have cotonete back working (and then we can remove the cronjob).

ilario:
> lets have something that works until we have cotonete back working (and then we can remove the cronjob)

Ok.
Anyway, @nicopace what was missing from Cotonete? Just the fix, right?

ilario:
I have a router I use as a client, and from times to times it seems to get deaf also.
Do you think it would make sense to add `sta` additionally to `adhoc` and `mesh` to the kind of interfaces which can get stuck?
As soon as that interface will get deaf I'll be able to test if it gets detected by Cotonete and fixed by this :D

spiccinini:
> I have a router I use as a client, and from times to times it seems to get deaf also.
> Do you think it would make sense to add `sta` additionally to `adhoc` and `mesh` to the kind of interfaces which can get stuck?
> As soon as that interface will get deaf I'll be able to test if it gets detected by Cotonete and fixed by this :D

Yes, I didn't know about deaf in ap/sta. Which driver is it?

ilario:
> Yes, I didn't know about deaf in ap/sta. Which driver is it?

Well, I still have to confirm that the problem is the deaf interface (looks so: it is associated but it does not ping, but I want to check using Cotonete).
The device is a Ubiquiti Nanostation LoCo M2, with ath9k driver.

spiccinini:
> Could need to be expanded to station interfaces and integrated with Cotonete, but as a temporary fix seems very good and very fast (the wireless interface stays down for less than half of a second)!

Yes it takes 0.17s per interface so it is almost transparent. I will add station support

gmarcos87:
Aproved. In the paravachasca network we are suffering from a new set of bugs incorporated in these latest releases of openwrt. For some of them we don't have workaround or ways to replicate them. Among the list of detected bugs are:
- Deaf radios that don't respond to the scan workarround
- Completely wrong MCS negotiations
- Some 5ghz channels that do not work

spiccinini:
> Aproved. In the paravachasca network we are suffering from a new set of bugs incorporated in these latest releases of openwrt. For some of them we don't have workaround or ways to replicate them. Among the list of detected bugs are:
> 
>     * Deaf radios that don't respond to the scan workarround
> 
>     * Completely wrong MCS negotiations
> 
>     * Some 5ghz channels that do not work

Thanks! for the deaf radios I hope I can access a device in that state to debug and write a proper fix.

PR #632, opened 2019-10-20
Adds tests-scheduler for planning tests at network usage peak and minimum

ilario:
This was the object of [my GSoC](https://blog.freifunk.net/2019/08/18/load-correlated-distributed-bandwidth-analysis-for-libremesh-networks-4-conclusions-and-further-work/), I am currently testing it but if you want to review, it already works.

It uses the number of hosts connected to the network (the broadcast domain) for estimating the network load. When I designed it I didn't think that modern people keep the smartphone always on and connected also during the night, so this could generate a false "peak" of usage?

I used shared-state for avoiding running the night tests (e.g. bandwidth-test) at the same time. I increased the TTL to 48 h, but seems that shared-state does not propagate data in the expected way, could anyone check if I'm using it the correct way?

(sorry for leaving all the WIP commits, I'll squash them at some point)


ilario:
Things that are still missing:

* add a cron line for copying the probe data to the flash memory once per day (the current method does not consider the case in which electricity cuts happen)
* shared-state does not work as expected, so find a list of the best hours (rather than just the first element of this list) and randomly take one of these for the night test. Do the same with the 5 minutes blocks.

ilario:
Closing this, as there was no interest on this in the last years. If anyone thinks it can be useful, just comment here and I will try to get this to a decent working state.

PR #633, opened 2019-10-21
protos: force vlanId to be numeric

ilario:
Between the arguments which can be provided to the routing protocols, the first one is usually the VLAN ID.
Force it to be numeric, for all the protocols except Batman-adv one which can be parametrized with %N1 (and except Babeld which is being treated in #631).


spiccinini:
As I understand this is fixing bugs as the vlanID *is always* a string the way it is passed now. I am +1 in the fix, only hoping that we can add tests.

Issue #618, opened 2019-10-09
Use `config/lime` for node config only and `config/lime-defaults` for the network config


spiccinini:
Having all the configuration in `/etc/config/lime` dificults knowing which configs are changed only for a node vs the configs that are for the network.

This may impact other parts of the code that is not using `lime.config`  (not sure but probably fbw or lime-app)

spiccinini:
Fixed in #635

PR #635, opened 2019-10-26
Node config clearly separated from network config

spiccinini:
This is a work in progress towards fixing #618 . I wanted to publish this as a RFC and to make it public so we are not doing the same work as this was assigned initialy to @G10h4ck 

Changes
* For some reason `config.get` and `config.get_all` were modifying `config/lime` storing the value found in `config/lime-defaults`. This is not only very bad practice as it is very unexpected but also this was populating config/lime with all the config/lime-defaults.
* ~Renamed config.foreach into config.node_foreach to make it clear that this is only retrieving the node specific configurations.~
* Added tests
*  Rename config/lime and config/lime-defaults into something that reflects more clear that we are splitting it into node specific configuration and a "shared network" configuration. Some ideas: `lime-node` fand `lime-network` or `lime-shared` (please propose better names).
*  Add lime-defaults-factory as a third and last level of config. So lime.config.get should first look into `lime-node`, if not found there look into `lime-network` and last seach into `lime-defaults-factory`. To properly support a working hierarchy we have to dinamicaly merge options from the different config levels. One idea is that we merge the three files into a `/config/lime-autogen` and use simple uci. This file should never be edited by hand and is automatically updated each time the other files are modified (for example using the modification time, it must be always newer than `lime-node`, `lime-network` and `lime-default-firmware`.
*  Review every place that it is using `uci` to get the lime configurations and change it to use `lime.config` if posible.
*  Review every use of `config/lime` and `config/lime-defaults`

TODO
- [ ] Write more docs / Update the docs in the webpage (after merging)


spiccinini:
I've just pushed most of the missing code. What do you think? I sill have to test it with hardware.

G10h4ck:
I am not sure about the naming of the files, in particular network is very broad term and i am not sure if it will be confusing, I have read the code and it seems quite robust, much better then previous approach!

spiccinini:
> I am not sure about the naming of the files, in particular network is very broad term and i am not sure if it will be confusing, I have read the code and it seems quite robust, much better then previous approach!

Thanks! I am olso not a fan of the name `lime-network` !  

We did a small brainstorming here at brasil and this came up as the better idea: 
use `config/lime-this-node` and `config/lime-all-nodes`. 

Other ideas:
`lime-community`
`lime-community-shared`
`lime-shared`

What do you think?

G10h4ck:
all-nodes is even more confusing because different communities nodes interoperate at L3 but usually have different configuration...

spiccinini:
> all-nodes is even more confusing because different communities nodes interoperate at L3 but usually have different configuration...

All the nodes of the community should share this config, so for them they are all the nodes that matter?

spiccinini:
Maybe lime-this-network? ...

G10h4ck:
we used to refer to L2 communities as clouds or nubes

spiccinini:
> we used to refer to L2 communities as clouds or nubes

Yes. But I don't think it is very clear. Also the division of L2/L3 may change... Other option is to call it "banana" or "cacahuate" or whatever that has no meaning....I also don't like this idea.

G10h4ck:
what about the even more confusin lime-neighboroud :p ? I agree that ATM lime-network seems the most sane

nicoechaniz:
regarding naming for the three levels, I like:
lime-node
lime-community
lime-defaults

spiccinini:
Tested up to this commit with a librerouter including firstbootwizard. Remaining change: rename the lime-network into lime-community

wfleurant:
looking forward to test on hardwares with 1 and with 2 radios. the lime configs are usually setting WiFi as a WAN and some are used in LAN/Mesh. hashing out when/why ULA (Unique Local Addr) is and is not advertised (depending on number of radios and/or number of ethernet ports (?)) is something i'd like to document better. in this regard, these anticipated changes to /etc/config/lime-* begs how to maintain the current documentation. Is this examples file simply divided up?  
    

https://github.com/libremesh/lime-packages/blob/master/packages/lime-docs/files/lime-example

spiccinini:
> looking forward to test on hardwares with 1 and with 2 radios. the lime configs are usually setting WiFi as a WAN and some are used in LAN/Mesh. hashing out when/why ULA (Unique Local Addr) is and is not advertised (depending on number of radios and/or number of ethernet ports (?)) is something i'd like to document better. in this regard, these anticipated changes to /etc/config/lime-* begs how to maintain the current documentation. Is this examples file simply divided up?
> 
> https://github.com/libremesh/lime-packages/blob/master/packages/lime-docs/files/lime-example

Hi! No, we are not splitting the configs per se, we are creating a hierarchy of configurations. By default lime-node and lime-community will be empty. If you want to change one of the defaults you must think if you want that config to apply only to this node then you use lime-node, if you want to config something that is shared by all the modes of a community then config it in lime-community. As lime-node has higher priority it will override the community configs. 
And yes we should update all the docs ... I am not very sure where as I didn't read LibreMesh docs yet.

spiccinini:
Now it should be ready to be merged, I also tested FBW.

spiccinini:
As this PR is changing the name of configs files specially lime-defaulta it will require an update to network-profiles

spiccinini:
I have rebased because there was a conflict with master

ilario:
This is an important change which is likely to create confusion within the users, please consider documenting it clearly on the website or on the lime-users mailing list or on the Matrix/Riot community chatroom, either now or when next release will be out!

spiccinini:
> This is an important change which is likely to create confusion within the users, please consider documenting it clearly on the website or on the lime-users mailing list or on the Matrix/Riot community chatroom, either now or when next release will be out!

Yes! Please help if you can.

PR #636, opened 2019-11-06
Option to ignore devices in shared-state-presist

gmarcos87:
We are currently thinking about other automatic behaviors with storage devices in routers. With this pull-request the behavior of shared-state-persist is changed so that it ignores the device if it has a file with the name .shared_state_persist_ignore_device in the root.

### DIFF
--- a/packages/shared-state-persist/files/usr/bin/shared-state-persist
+++ b/packages/shared-state-persist/files/usr/bin/shared-state-persist
@@ -74,7 +74,7 @@ local function getDevices()
 end
 
 local function isWritable(devicePath)
-	return fs.stat(devicePath..'/.read-only') == nil
+	return fs.stat(devicePath..'/.shared_state_persist_ignore_device') == nil
 end
 
 local function runPersist()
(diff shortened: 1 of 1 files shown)


G10h4ck:
If the point is to check the device is read-only or not we should be able to check the permissions, otherwise I would use a more specific file name like .shared_state_persist_ignore_device, or a configuration in uci to exclude a specific device

gmarcos87:
@G10h4ck, done. I change it to .shared_state_persist_ignore_device

germanferrero:
@gmarcos87 it is still looking for .read-only file

gmarcos87:
> @gmarcos87 it is still looking for .read-only file

I forgot to do "git push".

gmarcos87:
> Can you add at least a dummy unittest so at least we are checking that there are no syntax errors in the file? If you have more time you can test the new code :)

I have minimal connectivity at the moment, I can not download docker and everything that means the test environment.

nicopace:
I would like to propose for the file not to be hidden in linux.
Using a dotfile will mean that users that are not experienced will not be able to know if that file is there or not.
On the other hand, making it visible will mean that it is easy to debug if something is going wrong.

spiccinini:
@gmarcos87  I dared to modify the name of the file. Do you want to merge it ?

PR #638, opened 2019-11-08
updated pirania to latest version

luandro:
Important updates with:

- fixed client_ip script
- better authentication interface


luandro:
Still needs some work and testing before pushing.

Issue #433, opened 2018-11-28
lime-map-agent: Remove luci dependency


nicopace:
This package shouldn't depend on luci:
https://github.com/libremesh/lime-packages/blob/d2ba7a7b85ac6f4bb26741fd365016dccb39545d/packages/lime-map-agent/Makefile#L24

ilario:
@nicopace do you mean all the luci* dependencies?

ilario:
Maybe I got what you mean (investigating on #609): `luci-basic` is a dependency of `luci-lib-libremap` which is selected by `libremap-agent` which is selected by `lime-map-agent`.
Which point of this chain has to be broken?

ilario:
TL;DR: current implementation of libremap agent simply needs LuCI.

I thought that for breaking this dependency chain which causes #609 we could remove `luci-base` dependency in `luci-lib-libremap` (spoiler: no way)

in order to do so, we need to remove the usage of things from `luci-base`: 

* `luci.sys` gets used in `luci-lib-libremap`'s libremap.lua for few things:
  - sys.exec is a redirect to luci.utils library (also included in luci-base):
https://github.com/openwrt/luci/blob/1b835a50aabfbfe7d8d16a6d15c853315dc7ba79/modules/luci-base/luasrc/sys.lua#L25
which anyway just executes the command with io.popen:
https://github.com/openwrt/luci/blob/f73da9ca82ff8e0352c3942ea78ab7d1d0317668/modules/luci-base/luasrc/util.lua#L597-L603
  - sys.hostname just gets the hostname, can be done directly with nixio:
https://github.com/openwrt/luci/blob/1b835a50aabfbfe7d8d16a6d15c853315dc7ba79/modules/luci-base/luasrc/sys.lua#L36-L37
* `luci.httpclient` also depends on luci.util provided by luci-base and many other luci-base stuff

So I would say there's no way to disentangle luci-lib-libremap from luci-base.

Checking out [libremap-agent executable](https://github.com/libremap/libremap-agent-openwrt/blob/master/libremap-agent/files/usr/sbin/libremap-agent) turns out that it is also strictly bound to luci libraries and to luci-lib-libremap.

So seems that this issue cannot be solved.
If we want the libremap agent (do we?), we will have all this dependency cascade including the LuCI bloat.

spiccinini:
Great work @ilario documenting the dependencies!

Issue #327, opened 2018-02-13
replace luci with lime-app


aparcar:
Can we collect here what's needed to replace luci in the default installation with the (significatly?) smaller lime-app?

Missing features:

- [ ] set admin password
- [ ] add ssh key
- [ ] install sysupgrade https://github.com/libremesh/lime-packages/issues/325
- [ ] change basic wifi settings

p4u:
For each network interface, define the protocols to use (WAN, LAN, BMX, BAT-adv, etc.)

Raw edit of /etc/config/lime via web interface (just for advanced users but might be useful)

aparcar:
@gmarcos87 if you like to work on that I'm happy to help with the ubus backend

nicopace:
@aparcar it is important to notice that lime-app was not meant to replace luci but to include the features that non-technical people require.

In that regard, I guess it is featureful as it is :)
My take would be for it to replace luci in the default image, and allow the hackers to add luci if desired.

ilario:
@nicopace @gmarcos87 can you update the list with what's ready and what is not wanted?

ilario:
I just tested lime-app (see also #609), it is very cool and looks powerful but I cannot yet use it at 100%.

From the list in the first comment, what is more important and still missing is (in my opinion):

* set admin password (in my case, extremely dangerous as the WAN port of my router is directly accessible from the internet, repeating the [issue we had more than 2 years ago](https://lists.libremesh.org/pipermail/lime-dev/2017-January/000827.html))
* add ssh key
* 2.4 GHz distance
* 5 GHz distance
* Country code
* raw edit of /etc/config/lime for advanced users

More things that are needed are:

* a way to have lime-app on HTTPS
* the notes set in lime-app should be printed also in the banner shown when connecting via SSH (see #611)

Things that should be working but seem not working to me:

* ubus-lime-batman-adv: should it add some information about Batman-adv in lime-app? I cannot see where
* ubus-lime-groundrouting: should it add a way to configure ground routing? I cannot see where
* ubus-lime-openairview: depends on fft-eval which is not present in 18.06 releases so I suppose will work once we base a release on OpenWrt 19.something
* location: I cannot move my node from the middle of the ocean, neither with Firefox nor with Chromium
* metrics: makes me wait forever with "Measuring links" (my node is directly connected to the internet with a public IPv4 on its WAN port)
[...]

PR #644, opened 2019-11-23
Make ubus-lime-location optional

ilario:
Here I removed the hard dependency of ubus-lime-location from lime-app.
In parallel, I already documented that it should be manually selected when compiling the firmware, see https://github.com/libremesh/lime-web/pull/101

Considering that ubus-lime-location causes a cascade of dependencies which includes luci-base (see #433, ubus-lime-location selects libremap-agent which selects luci-lib-libremap which selects luci-base), I thought that users trying to compile a firmware with lime-app for routers with limited flash memory could want to avoid including LuCI.

See also discussion on  #471

### DIFF
--- a/packages/lime-app/Makefile
+++ b/packages/lime-app/Makefile
@@ -21,7 +21,7 @@ define Package/$(PKG_NAME)
 	MAINTAINER:=Marcos Gutierrez <gmarcos@altermundi.net>
 	URL:=http://github.com/libremesh/lime-app
 	DEPENDS:=+uhttpd +uhttpd-mod-ubus \
-		+ubus-lime-location +ubus-lime-metrics +ubus-lime-utils \
+		+ubus-lime-metrics +ubus-lime-utils \
 		+ubus-lime-openairview +ubus-lime-grondrouting
 	PKGARCH:=all
 endef


nicopace:
The proposal breaks the experience of the lime-app, and should be avoided.


El 23 de noviembre de 2019 20:02:08 GMT+02:00, Ilario Gelmetti <notifications@github.com> escribió:
>Here I removed the hard dependency of ubus-lime-location from lime-app.
>In parallel, I already documented that it should be manually selected
>when compiling the firmware, see
>https://github.com/libremesh/lime-web/pull/101
>
>Considering that ubus-lime-location causes a cascade of dependencies
>which includes luci-base (see #433, ubus-lime-location selects
>libremap-agent which selects luci-lib-libremap which selects
>luci-base), I thought that users trying to compile a firmware with
>lime-app for routers with limited flash memory could want to avoid
>including LuCI.
>
>See also discussion on  #471
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/libremesh/lime-packages/pull/644
>
>-- Commit Summary --
>
>  * remove ubus-lime-location from lime-app dependencies
>
>-- File Changes --
>
>    M packages/lime-app/Makefile (2)
>
>-- Patch Links --
>
>https://github.com/libremesh/lime-packages/pull/644.patch
>https://github.com/libremesh/lime-packages/pull/644.diff
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/libremesh/lime-packages/pull/644

-- 

ilario:
> The proposal breaks the experience of the lime-app

What's the difference between having ubus-lime-location as a hard dependency of lime-app or having it selected as a package via menuconfig?

ilario:
Do you mean that without that package lime-app does not work at all? I actually did not test that as I assumed the rest of the functions should work.

ilario:
I just tested a complete image compiled with lime-app but without ubus-lime-location.
The location entry still appears in the menu and even the map appears.
I cannot change the position of the node but this was also the case when ubus-lime-location was installed (see #327, maybe this is the reason why there are more than ten thousands nodes in the middle of the ocean [in LibreMap here](http://libremap.net/#bbox/-9.384032109601689,-13.33740234375,7.013667927566642,13.02978515625)).

My original point (making the dependency optional in order to allow users to make an image without `luci-base` package) is strongly hindered by the fact that also `check-http-date` and `shared-state` require `luci-base` (through the dependency on `luci-httpclient`).
I am not sure if any user will ever use LibreMesh without shared-state (?).

ilario:
Having a closer look to ubus-lime-location I think I managed to understand what @nicopace meant: this module is used also for discovering which are the other nodes in the network and this information is used by lime-app at least for "changing the base station".

Weird thing is that for doing this, ubus-lime-location uses shared-state-nodes_and_links but there is no dependency from it in its Makefile.
https://github.com/libremesh/lime-packages/blob/152e1442cfae1ef4e8997afa56039f199bf303a4/packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location#L95-L96

Anyway, closing this PR.

PR #646, opened 2019-11-24
No-js Pirania

luandro:
Update with a released a major update on Pirania that works without Javascript, making it much more reliable.


nicopace:
The revision of this commit is complex, as has nothing to do with the code that it is here, but on the other repo :)
any recommendation on improving the mechanism? (the same happens with the lime-app).

luandro:
You're right @nicopace. I should have made PRs in Pirania repo... thought nobody was paying much attention, that's why I've just been merging stuff to master on my own. Will be more careful next time.

luandro:
Anyways, there seems to be problems with this PR. I'll close this for now...

PR #651, opened 2019-11-25
lime-app installs the redirect to /app

### LINKED ISSUE #609: Redirect from LuCI to lime-app not working
I am not experienced with lime-app, but when I open thisnode.info what I get is just this:

Screenshot from 2019-10-04 16-44-31

and I have to explicitly include `/app` in the URL in order to get the FirstBootWizard or the app.
What am I doing wrong?
I compiled following the instructions [on the website](https://libremesh.org/development.html).
Thanks!

ilario:
Fixes #609 
The new index file used for redirection has been copied from LibreRouter's OpenWrt repository [here](https://github.com/LibreRouterOrg/openwrt/blob/d2282e29b0888dfa272149d845f48b202ec61b6f/files/www/index.html).

### DIFF
--- a/packages/lime-app/files/96-lime-app-index_page
+++ b/packages/lime-app/files/96-lime-app-index_page
@@ -1,5 +1,3 @@
 #!/bin/sh
 
 uci set uhttpd.main.index_page=lime_app_index.html
-uci commit uhttpd
-/etc/init.d/uhttpd restart
--- a/packages/lime-app/files/96-lime-app-index_page
+++ b/packages/lime-app/files/96-lime-app-index_page
@@ -1,0 +1,5 @@
+#!/bin/sh
+
+uci set uhttpd.main.index_page=lime_app_index.html
+uci commit uhttpd
+/etc/init.d/uhttpd restart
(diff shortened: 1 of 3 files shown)


packages/lime-app/files/96-lime-app-index_page
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+uci set uhttpd.main.index_page=lime_app_index.html
+uci commit uhttpd
```
spiccinini: uci commit should not be necesary as the boot script runs uci commit after each uci-defaults file (https://github.com/openwrt/openwrt/blob/master/package/base-files/files/etc/init.d/boot#L17)

packages/lime-app/files/96-lime-app-index_page
```diff
@@ -0,0 +1,5 @@
...
+/etc/init.d/uhttpd restart
```
spiccinini: restarting a service in the uci-defaults is not a good idea because at boot time this runs before uhttpd has even started to start (because boot has `START=10` and for example uhttpd runs at has `START=50`)

packages/lime-app/files/96-lime-app-index_page
ilario: thanks! I didn't know :)

packages/lime-app/files/96-lime-app-index_page
ilario: Rrrrright!

Issue #262, opened 2017-12-11
Advices for an optimal setup of the bands for Dual Band Routers


modante:
Hello.
We are a little neighborhood with 3 libremesh Dual Band Routers (TP-Link wdr3500, wdr3600 and wdr4100) that can work simultaneously in 2.4 and 5 GHz.
One of the routers have an internet connection to a WISP through the 2,4 GHz band using a NanoStaion Loco device and connected to the WAN port and this connection is shared through the rest of the routers of the mesh.
I saw that the default config of the LibreMesh clones the 3 WiFi Stations (LibreMesh.org, LibreMesh.org/LiMe-xxxxxx and LiMe) in all the radios availables on the device and I suppose that for the mesh (LiMe AdHoc) uses the best connection of them or both (please, confirm me the behavior).
I thought that in order to increase the performance (decreasing the interferences and dividing the bandwidth between the 2 radios) would be better to use the 5 GHz radios to run the mesh (LiMe), removing the APs from this band and exclusively the 2,4 Ghz to run the connection with the devices (mobiles, PCs, etc.), removing the mesh connection (LiMe) of this band.
What do you think is a better setup in this case?
Thank you and regards :-)

p4u:
> I saw that the default config of the LibreMesh clones the 3 WiFi Stations (LibreMesh.org, LibreMesh.org/LiMe-xxxxxx and LiMe) in all the radios availables on the device and I suppose that for the mesh (LiMe AdHoc) uses the best connection of them or both (please, confirm me the behavior).

It's a bit more complex than this but yes, the routing protocols decide for each packet which is the best next hope to a destination.

> I thought that in order to increase the performance (decreasing the interferences and dividing the bandwidth between the 2 radios) would be better to use the 5 GHz radios to run the mesh (LiMe), removing the APs from this band and exclusively the 2,4 Ghz to run the connection with the devices (mobiles, PCs, etc.), removing the mesh connection (LiMe) of this band.
> What do you think is a better setup in this case?

I would say it does not have a big impact because in general the 5G AP is not used and Mesh at 2.4GHz is usually not used either (5GHz is usually the best path). However disable it won't hurt.

What might have a big impact is to change the 2.4GHz AP channels, by default it's always 11 but you can set it to different channels such as 1 or 6. 5GHz radios must be always the same freq (to not breack the mesh cloud) but 2.4GHz can be different.

[...]

Issue #109, opened 2017-03-31
Measure performance impact of having multiple virtual AP


nicopace:
The current LiMe implementation uses many virtual AP (like multiple SSIDs). We want to know if this is a performance hit in order to take better decisions on how we use this functionality.

aparcar:
@dangowrt you know something about that?

ilario:
To know if the presence of multiple interfaces (and nodes or clients connected to it) affects the performances would be very useful also for discussing the optimal setup in #262.

nicopace:
I don't feel this is a release-blocker @ilario. could you expand or remove the tag? thanks!

ilario:
Sorry, I originally added the tag (and I removed it now) just because is a thing that so easy to test and potentially so important for the performances that I thought we absolutely have to do it before the next release.

spiccinini:
@ilario do you have an idea of how to measure performance?

ilario:
Meh, the idea was to set up a test network, maybe 3 routers connected in a line like A-B-C (I have only omnidirectional routers so maybe I would have to decrease the output power in order to avoid a direct A-C connection), with a client connected via Wifi to B (doing a little of background traffic) and then run an iperf3 between one PC connected by cable to A and another connected by cable to C.
The situations I would compare are:

A(2.4GHz: ap, apname; 5GHz: ap, apname, mesh)
B(2.4GHz: ap; 5GHz: ap, apname (client connects here), mesh)
C(2.4GHz: ap, apname; 5GHz: ap, apname, mesh)

A(2.4GHz: ap, apname; 5GHz: ap, apname, mesh)
B(2.4GHz: ap, apname (client connects here); 5GHz: ap, mesh)
C(2.4GHz: ap, apname; 5GHz: ap, apname, mesh)

A(2.4GHz: ap, apname; 5GHz: mesh)
B(2.4GHz: ap, apname (client connects here); 5GHz: mesh)
C(2.4GHz: ap, apname; 5GHz: mesh)

A(2.4GHz: ap, apname; 5GHz: mesh)
B(2.4GHz: apname (client connects here); 5GHz: mesh)
C(2.4GHz: ap, apname; 5GHz: mesh)

ilario:
Some testing also on this has been reported here:
https://lists.libremesh.org/pipermail/lime-users/2020-November/001817.html

ilario:
Closing, until when there will be hints of this being important.

PR #653, opened 2019-11-26
Remove AP and APname from 5 GHz radios

ilario:
Following the discussion on #262, related also with #109.
The idea is that for communities using dual band routers, the 5 GHz radios should be used just for mesh backbone (node to node connections) while the 2.4 GHz radios should be used for clients connections (AP and APname).
On LibreRouter (a dual band router), the default is already like this, as you can see [here in their repository](https://github.com/LibreRouterOrg/openwrt/blob/863d464a4741a52494b197e6ca50abb2ac60f198/files/etc/config/lime-community#L17-L19).
I expect this to provide a performances improvement but I did not test it. Can anyone comment on this?

We cannot set this as a default because single-band 2.4 GHz routers (e.g. Ubiquiti NanoStation M2) would be cut out of the mesh backbone, so in this pull request the mesh IEEE802.11s interface is active both on 2.4 and 5 GHz radios.

Routers with just 5 GHz radio (e.g. Ubiquiti NanoStation M5) will be used just for mesh backbone, I hope this is ok?

Please comment!!


G10h4ck:
I don't agree, what happens when flashing a 5Ghz only device?
I think it is acceptable if the default is not performant, but we shouldn't lock out from their device people because they ain't got 2.4Ghz radio...

ilario:
> I don't agree, what happens when flashing a 5Ghz only device?

The wireless will be configured just for mesh.

> we shouldn't lock out from their device people because they ain't got 2.4Ghz radio...

The 5 GHz only devices are usually directional or sectorial antennas, people usually connects to that via ethernet cable, not via wifi. I think that nobody will be "locked out", but if we think that this could happen it is ok to close this for me.
But we should document this possibility and suggest to the community to use it.

ilario:
Did not receive any positive feedback, closing.
The same issue can be addressed with [lime-smart-wifi](https://github.com/libremesh/lime-packages/tree/c4fb7118e8393b65264626cb60340f0bb0228793/packages/lime-smart-wifi) package but it looks like it's not going to be included in any release.

PR #655, opened 2019-11-27
lime.network: fix /etc/dnsmasq.d/*.conf not being loaded

spiccinini:
`/etc/dnsmasq.d/*.conf` weren't loaded so for example `/etc/dnsmasq.d/lime-proto-anygw-20-ipv6.conf` was not loaded...

So this fixes ipv6 slaac dns resolution among other things. Options now aplied from `/etc/dnsmasq.d/lime-proto-anygw-20-ipv6.conf`:
```
enable-ra
ra-param=anygw,mtu:1350,120
dhcp-range=tag:anygw,2a00:1508:a85:3a00::,ra-names,24h
dhcp-option=tag:anygw,option6:domain-search,thisnode.info
```

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -138,7 +138,15 @@ function network.setup_dns()
 	)
 	uci:save("dhcp")
 
-	fs.writefile("/etc/dnsmasq.conf", "conf-dir=/etc/dnsmasq.d\n")
+	--! some dynamic scripts prefer to use confdir=/tmp/dnsmasq.d (which is the openwrt default)
+	--! so we add all the /etc/dnsmasq.d/*.conf files as individual conf-file options instead
+	--! of using uci set confdir=/etc/dnsmasq.d
+	local dnsmasq_conf_file = ""
+	for conf_file in fs.glob("/etc/dnsmasq.d/*.conf") do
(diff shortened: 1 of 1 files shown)


G10h4ck:
dnsmasq doesn't support multiple conf-dir ?

spiccinini:
> dnsmasq doesn't support multiple conf-dir ?

No the docs doesn't say anything about multiple confdirs and it didn't work when I tested it yesterday.

ilario:
How did you verify that those options were not loaded?
To me it looks that multiple conf-dir are working.
To check, I added new DNS servers in a file in `/etc/dnsmasq.d` and another in `/tmp/dnsmasq.d`, then I launched `dnsmasq --conf-file=/var/etc/dnsmasq.conf.cfg01411c` and from `logread` I can see that both the new DNS servers are being configured.

spiccinini:
> How did you verify that those options were not loaded?
> To me it looks that multiple conf-dir are working.
> To check, I added new DNS servers in a file in `/etc/dnsmasq.d` and another in `/tmp/dnsmasq.d`, then I launched `dnsmasq --conf-file=/var/etc/dnsmasq.conf.cfg01411c` and from `logread` I can see that both the new DNS servers are being configured.

So you say that right now dnsmasq should be loading `/etc/dnsmasq.d/lime-proto-anygw-20-ipv6.conf` but it wasn't  .... I will test again!.

spiccinini:
My bad, somehow I thought that this wasn't working but it is working

Issue #236, opened 2017-09-22
IPv6 default route announced even when there's no IPv6 gateway


ilario:
Connecting to a LibreMesh node I received a default route for IPv6, even if there was no IPv6 gateway.
```
# ip -6 r
2a00:1508:a2e:1300::/64 dev enp0s25 proto kernel metric 202 mtu 1500 pref medium
fe80::/64 dev enp0s25 proto kernel metric 256 pref medium
default via fe80::a8aa:aaff:fe2e:13aa dev enp0s25 metric 202 mtu 1500 pref medium
```

This wouldn't be a problem if we used ISP DNS (related to #95) as they don't answer for AAAA queries, but the custom ones set in lime-defaults (4.2.2.2, 141.1.1.1) does.
Most of modern applications fallback on IPv4 after realizing that IPv6 doesn't work, but many still don't.
This result in some applications not being able to reach the internet, for example ping.

In my opinion we should disable IPv6 default route announcement by default and let the user or some automatic stuff like watchping6 from #71 activate it.

ilario:
I think @altergui also observed this.

ilario:
Can anyone point me to the line in the code where this is set?

G10h4ck:
> Most of modern applications fallback on IPv4 after realizing that IPv6 doesn't work, but many still don't.

ilario:
> reasonable linux distributions handle ICMP messages properly

My experience of this problem was while running Linux

ilario:
> reasonable linux distributions handle ICMP messages properly

* Arch Linux does not
* LEDE 17.01.4 does not (I was trying to "opkg update" on a plain LEDE connected to a LibreMesh gateway and failed using IPv6)

Maybe we should consider this issue more seriously? @G10h4ck

ilario:
Related to #71

altergui:
On 22/09/17 11:28, Ilario Gelmetti wrote:
> I think @altergui <https://github.com/altergui> also observed this.

NAK, the default gw should always be announced, regardless of v6
internet connectivity, the same way a default ipv4 gw is announced,
regardless of v4 internet connectivity

(else, there would be no ipv6 way of crossing over to another cloud, in
a set of libremesh clouds connected between them , but not to the internet)

dangowrt:
Hi!

On Mon, Jan 22, 2018 at 02:03:03PM +0000, Gui wrote:
> On 22/09/17 11:28, Ilario Gelmetti wrote:
[...]

PR #657, opened 2019-11-28
Change lime-defaults to use Ipv6 private ULA addresses

### LINKED ISSUE #656: Use private instead of public ipv6 addresses
The proposal is to use RFC4193 ULA addresses  aka `fd00::/8`

Quoting from this comment https://github.com/libremesh/lime-packages/issues/236#issuecomment-359461126

> Our currenty way of having an publicly routed default IPv6 prefix
> (literally **everybody** uses 2a00:1508:0a42:2400::/64) as the default
> prefix for LiMe nodes is wrong in a couple of ways:
> - this prefix can impossibly be routed back to all nodes using it
> - using addresses within that prefix as a source address will hence
>   only work in a single-gateway scenario **having exactly this prefix
>   delegated from upstream**
> - RAs with that public prefix are sent out and hence clients pick
>   (unreachable form outside) addresses, and as the default gateway
>   is also within the 2000::/4 range it will make clients believe
>   that we got access to other public IPv6 destinations which isn't
>   true in most cases -- this will slow down surfing as browsers
>   sometimes try reaching IPv6 addresses and fail.

> I suggest to move the default main_ipv6_address into an
> network-name-hash-based ULA prefix, e.g. fd%N1:%N2%N3:%N4%N5::/48 or
> something like that to decrease the overall confusion...

spiccinini:
Fixes #656


ilario:
Works!
Could you adapt the documentation in lime-example here
https://github.com/libremesh/lime-packages/blob/3c76578e5c51dd6ffdadb0e2d3b61f016f52deab/packages/lime-docs/files/lime-example#L8

and here?
https://github.com/libremesh/lime-packages/blob/3c76578e5c51dd6ffdadb0e2d3b61f016f52deab/packages/lime-docs/files/lime-example#L76

ilario:
Is there a difference between using /64 or /48 here? I mean, we're fixing just the first 3 fields of the IPv6, so /48 would be an option, right?

ilario:
Also fixes #236

G10h4ck:
Shouldn't we change this in lime-default-factory too?

ilario:
@G10h4ck lime-default-factory introduced in #434 has been annihilated in #635, in two days it would have been its birthday.

G10h4ck:
> @G10h4ck lime-default-factory introduced in #434 has been annihilated in #635, in two days it would have been its birthday.

`sed 's/lime-default-factory/lime-default/g`

ilario:
> > @G10h4ck lime-default-factory introduced in #434 has been annihilated in #635, in two days it would have been its birthday.
> 
> `sed 's/lime-default-factory/lime-default/g`

Ah, it is already edited also in lime-defaults, can you check if it is ok?

PR #663, opened 2019-12-08
Set default country code to US

### LINKED ISSUE #594: Channel 157 and non-specified country code
In the [lime-defaults-factory](https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/etc/config/lime-defaults-factory) we specify a list of 5 GHz channels as:

```
	list channel_5ghz '157'
	list channel_5ghz '48'
```

which (for most of the double band routers where phy0 is 2.4 GHz and phy1 is 5 GHz) means that we use just channel 48 (see #542).

The channel 48 works with any country code but the channel 157 does not.

And we do not specify a default country code in our configuration file, so by default the country code depends on the router.

For example in one of my routers (TP-Link WDR3600) it is US for which 157 works fine, in another (YouHua WR1200JS) it is 00:

```
# iw reg get
global
country 00: DFS-UNSET
	(2402 - 2472 @ 40), (N/A, 20), (N/A)
	(2457 - 2482 @ 20), (N/A, 20), (N/A), AUTO-BW, PASSIVE-SCAN
	(2474 - 2494 @ 20), (N/A, 20), (N/A), NO-OFDM, PASSIVE-SCAN
	(5170 - 5250 @ 80), (N/A, 20), (N/A), AUTO-BW
	(5250 - 5330 @ 80), (N/A, 20), (0 ms), DFS, AUTO-BW, PASSIVE-SCAN
	(5490 - 5730 @ 160), (N/A, 20), (0 ms), DFS, PASSIVE-SCAN
	(5735 - 5835 @ 80), (N/A, 20), (N/A), PASSIVE-SCAN
	(57240 - 63720 @ 2160), (N/A, 0), (N/A)
```

and with this country code the channel 157 behaves very badly:

```
# iwinfo
wlan0-ap  ESSID: "es.hackmeeting.org"
          Access Point: D6:5F:25:EB:7F:64
          Mode: Master  Channel: 11 (2.462 GHz)
          Tx-Power: 22 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: unknown
          Bit Rate: unknown
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: 14C3:7603 14C3:7603 [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

wlan0-apname ESSID: "LibreMesh.org/LiMeHM-eb7f64"
          Access Point: D2:5F:25:EB:7F:64
          Mode: Master  Channel: 11 (2.462 GHz)
          Tx-Power: 22 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: unknown
          Bit Rate: unknown
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: 14C3:7603 14C3:7603 [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

wlan0-mesh ESSID: "LiMe"
          Access Point: D4:5F:25:EB:7F:64
          Mode: Mesh Point  Channel: 11 (2.462 GHz)
          Tx-Power: 22 dBm  Link Quality: 70/70
          Signal: -22 dBm  Noise: unknown
          Bit Rate: 40.0 MBit/s
          Encryption: unknown
          Type: nl80211  HW Mode(s): 802.11bgn
          Hardware: 14C3:7603 14C3:7603 [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy0

wlan1-ap  ESSID: "es.hackmeeting.org"
          Access Point: D6:5F:25:EB:7F:66
          Mode: Master  Channel: 157 (5.785 GHz)
          Tx-Power: 16 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: unknown
          Bit Rate: unknown
          Encryption: none
          Type: nl80211  HW Mode(s): 802.11nac
          Hardware: 14C3:7662 14C3:7662 [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy1

wlan1-mesh ESSID: unknown
          Access Point: D4:5F:25:EB:7F:66
          Mode: Mesh Point  Channel: unknown (unknown)
          Tx-Power: 16 dBm  Link Quality: unknown/70
          Signal: unknown  Noise: unknown
          Bit Rate: unknown
          Encryption: unknown
          Type: nl80211  HW Mode(s): 802.11nac
          Hardware: 14C3:7662 14C3:7662 [Generic MAC80211]
          TX power offset: unknown
          Frequency offset: unknown
          Supports VAPs: yes  PHY name: phy1
```

As you can see, wlan1-apname does not get created and wlan1-mesh is broken.

While both works fine with channel 48 or with channel 157 + explicit `option country US` in `/etc/config/lime`.

So there can be two solutions:

* replace channel 157 with one channel that works with every countrycode (e.g. 36 which is too close to 48 @spiccinini said)
* or specify a countrycode in the default configuration (e.g. US)

Reference table of allowed channels on Wikipedia [here](https://en.wikipedia.org/wiki/List_of_WLAN_channels#regulatory_tables5.0ghz).

ilario:
A default country code has to be set for routers not having a factory default one, see #594.
`US` is a safe default, as most of the commercial routers have it as a factory default.

Fixes #594

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -59,7 +59,7 @@ config lime network
 ### WiFi general options
 
 config lime wifi
-#	option country 'ES'				# set this to your location country code, for example in Spain, setting ES allows you to use channel 13
+	option country 'US'				# set this to your location country code, for example in Spain, setting ES allows you to use channel 13
 	option channel_2ghz '11'			# May be either a list or a single option, in case of a list channel will be selected according to radio index
 #	option channel_5ghz '48'			# Check for allowed channels on https://en.wikipedia.org/wiki/List_of_WLAN_channels#regulatory_tables5.0ghz
 	list channel_5ghz '48'				# If channel is a list, each entry of the list is used for a different radio
--- a/packages/lime-system/files/etc/config/lime-defaults
+++ b/packages/lime-system/files/etc/config/lime-defaults
@@ -45,6 +45,7 @@ config lime network
 	option use_odhcpd false
 
 config lime wifi
+	option country 'US'
 	option channel_2ghz '11'
 	list channel_5ghz '48'
 	list channel_5ghz '157'


dangowrt:
The only 'safe default' would be using `00` (aka. WORLD) country code.
The fact that many commercial routers have the US (or CN) country code hard-coded in their ath9k EEPROM data is simply that those vendors did not implement regulatory compliancy in the way Atheros imagined. They ignore the value from EEPROM and rather got their own proprietary solution to make devices come up in a compliant way according to the place they were sold, ie. hard-coded in the driver of a country-specific firmware (TP-LINK) or selectable by the user on first-boot before any radio comes up (ubiquiti).
Setting `US` as country code in OpenWrt hence makes the device non-compliant with EU radio regulations (too much TX power in 2.4GHz, use of prohibited channels in 5GHz) and really isn't a good idea (users face legal risks).

ilario:
Yesss, I agree that 00 would be the most correct default.
00 is also the country code I found when flashing OpenWrt on YouHua WR1200JS routers, and this originated the problem reported in #594. I thought that this was an acceptable solution to that issue...
The alternative solution is to use 
```
	list channel_5ghz '48'
	list channel_5ghz '36'
```
(which are too close for using HT40) instead of the currently employed 
```
	list channel_5ghz '48'
	list channel_5ghz '157'
```
where the 157 is not possible to use with country code 00 (is this right?).

dangowrt:
Hm, I understand and didn't think of having 2 radios in 5 GHz band...
Maybe the default should be HT20 with channels available in WORLD regdomain then? Anyway, those defaults are for demonstration purpose and will practially always be overwritten by the community profile which also sets a country code.

ilario:
> Maybe the default should be HT20 with channels available in WORLD regdomain then?

Makes sense.
@spiccinini @G10h4ck @gmarcos87 @nicopace opinions?

ilario:
Thinking it again, the (only?) router affected by #594 will be LibreRouter, and its developers already use a [custom /etc/config/lime-community configuration](https://github.com/LibreRouterOrg/openwrt/blob/863d464a4741a52494b197e6ca50abb2ac60f198/files/etc/config/lime-community) where they can set the 00 country code if they decide so.
For the general LibreMesh user pool maybe is better to leave this undefined.
Do you agree that we can close this and migrate #594 on LibreRouter repositories?

G10h4ck:
I agree on closing this, I have played a bit with it in some routers and in some cases specifying a country code meant the radio won't come up even if the channel used was legal in that country... while the radio came removing the country code line from configuration.

PR #664, opened 2019-12-08
Add txpower option for wireless radios

ilario:
I some rare cases can make sense to decrease the output power of wireless radios.
This PR adds the possibility to set the txpower both globally (makes sense only for single-radio routers) or specifically for each wireless radio.


pedro-nonfree:
> I some rare cases can make sense to decrease the output power of wireless radios.

this is important in bmx6 networks, where a bad link could make bad behavior on routing.

to address this issue, if you allow override of variables it is feasible: you have a global variable configuration, and from local variable that could be overridden

ilario:
> to address this issue, if you allow override of variables it is feasible: you have a global variable configuration, and from local variable that could be overridden

Currently, the default configurations are located in /etc/config/lime-defaults.
Then the communities can add their options in /etc/config/lime-community which has priority over the lime-defaults file.
Finally, the node owner can customize /etc/config/lime-node which has priority over both of the previous files (see #635). This last file is where it makes sense to include customization of txpower.

G10h4ck:
to play with txpower is not that rare actually ;)

PR #665, opened 2019-12-08
Specific wireless interface: if no mode is configured, use default ones

ilario:
Currently, if a specific wireless interface configuration is used but no mode is specified, `lime-config` fails due to indexing of a nil value in `if modes[1] ~= "manual" then` introduced in https://github.com/libremesh/lime-packages/commit/d4f7a8c2bf20a373854c1f796dec9f9916642bbb.
Even if it was not failing, I think that a specific interface configuration without any mode does not make much sense. So now, if no mode is specified, the general ones are used.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -116,7 +116,7 @@ function wireless.configure()
 		local options = config.get_all("wifi")
 
 		if specRadio then
-			modes = specRadio["modes"]
+			modes = specRadio["modes"] or modes
 			options = utils.tableMelt(options, specRadio)
 		end
 


G10h4ck:
the problem with this is that it becomes troublesome to keep the behaviour of specific stuff coherent, so for example why doing this for wifi modes, and not for network protocols, or wifi channels?
IMHO if a configuration is broken we should not do black magic to make it works, what I suggest here is to solve the crash, and at maximum, taking in account that we are on an embedded platform and we should keep the things as little as possible.print a warning message saing that modes are missing

ilario:
> the problem with this is that it becomes troublesome to keep the behaviour of specific stuff coherent, so for example why doing this for wifi modes, and not for network protocols, or wifi channels?

Are you referring to the usage of tableMelt to use default options in the specific interface configuration? That is not being modified in this PR, instead the same approach is expanded to the modes. I am not a fan of tableMelt usage but if we decide to keep it, also the default modes should be used as a fallback, for coherence, and I agree that we should do the same for the `net` sections.

> IMHO if a configuration is broken we should not do black magic to make it works, what I suggest here is to solve the crash, and at maximum, taking in account that we are on an embedded platform and we should keep the things as little as possible.print a warning message saing that modes are missing

Ok, I'll open an alternative PR. But in case the alternative approach is chosen, consider completely removing also tableMelt usage.

G10h4ck:
This PR is ok and we should merge it as is, If you want to investigate also the specific net section you are welcome, but I warn you that it will need broad and deep testing in many different setups

ilario:
The alternative PR, in which the situation is unchanged but the documentation is fixed, is #668.

> If you want to investigate also the specific net section you are welcome, but I warn you that it will need broad and deep testing in many different setups

No thanks

G10h4ck:
I think we should merge this and close the other PR

PR #668, opened 2019-12-16
Specific wireless interface: document that modes list is always needed

ilario:
Alternative to #665, see the discussion there.


G10h4ck:
I think the right and coherent way to fix this is https://github.com/libremesh/lime-packages/pull/665

Issue #613, opened 2019-10-04
Consider including wpad-mesh-wolfssl and update lime-example


ilario:
Currently we ship images with just wpad-mini.
This means that one willing to have encryption over the ieee802.11s mesh connections has to install additional packages.
This is incorrectly documented in lime-example: [wpad-mesh](https://github.com/libremesh/lime-packages/blob/18c8aa59ec14bf8a2e82f72b0b761194d525213c/packages/lime-docs/files/lime-example#L88) should be replaced by wpad-mesh-wolfssl or wpad-mesh-openssl.
In my opinion we could just add wpad-mesh-wolfssl in the default packages list.
This also have the advantage of adding the support for WPA3-SAE.

I tried to compile an image (with all the LibreMesh packages, for a TP-Link WDR3600) with just wpad-mini and one with just wpad-mesh-wolfssl (strange thing, the package does not appear at first in the menuconfig, it appears after selecting and deselecting wpad):
With only wpad-mini the sysupgrade.bin image is 4915204 bytes.
With only wpad-mesh-wolfssl the sysupgrade.bin image is 5308420 bytes.
So there's 384 kB difference.

Both if we agree on this or not, we should document this also on the website and update lime-example.

nicopace:
would this also allow the use of ssl on uhttpd?

ilario:
I don't think so... Maybe that would be more overlapping if we use `wpad-mesh-openssl`?

nicopace:
Yep, they share packages:
* wpad-mesh-wolfssl depends on libwolfssl
* libuhttpd-wolfssl depends on libustream-wolfssl that depends on libwolfssl
This adition would add ~30kb extra.
I think it is worth exploring... also in relation with https://hackmd.io/@nicopace/PWA-LibreRouter
CC: @gmarcos87

dangowrt:
> would this also allow the use of ssl on uhttpd?

Yes, you'd need to install `libustream-wolfssl` and `px5g-standalone` to make uhttpd use WolfSSL for HTTPS.

ilario:
This sounds great!
As I suppose we're not going to support any 4 MB flash memory router anymore, there's no excuse for not supporting mesh encryption by default :)
@germanferrero @gmarcos87 @spiccinini do you plan to allow the access to lime-app using HTTPS? Do you know if these packages would allow it?

dangowrt:
[...]

PR #669, opened 2019-12-16
Update lime-example

ilario:
Many modifications to lime-defaults were not reflected to lime-example.

Solves a small parts of #613 and #262

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -16,8 +16,10 @@
 
 config lime system
 	option hostname 'LiMe-%M4%M5%M6'		# Parametrizable with %Mn
-	option domain 'lan'				# DNS domain for the L2 cloud it could be something like mycloud.mynetwork.tld, hosts that get their IP by DHCP will also get a fully qualified domain name like hostname.mycloud.mynetwork.tld
+	option domain 'thisnode.info'				# DNS domain for the L2 cloud it could be something like mycloud.mynetwork.tld, hosts that get their IP by DHCP will also get a fully qualified domain name like hostname.mycloud.mynetwork.tld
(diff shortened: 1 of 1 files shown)


packages/lime-docs/files/lime-example
```diff
@@ -26,7 +28,7 @@ config lime network
...
 	option anygw_dhcp_limit '0'			# Number of IPs available for DHCP. Use zero for having the DHCP pool ranging from anygw_dhcp_start up to the end of the subnet. For example, if the subnet is 10.x.0.0/16, the start of the DHCP pool is at 10.x.100.2 and you want the DHCP pool to finish at 10.x.127.254, the limit parameter will have to be (127 - 100) * 256 + (254 - 2) + 1 = 7165. Instead, if you want the DHCP pool to go from 10.x.100.2 up to 10.x.255.254 (last valid IPv4 in the /16 subnet) you can just set the limit to zero.
-	option main_ipv6_address '2a00:1508:0a%N1:%N200::/64'	# Parametrizable in the same way as main_ipv4_address. If used, the IP autocompletion will fill maximum the last 24 bits, so specifying an IP autocompletion range bigger than /104 is not useful.
+	option main_ipv6_address 'fd%N1:%N2%N3:%N4%N5::/64'	# Parametrizable in the same way as main_ipv4_address. If used, the IP autocompletion will fill maximum the last 24 bits, so specifying an IP autocompletion range bigger than /104 is not useful.
```
G10h4ck: As this is a documentation file an address in the `2001:db8::/32` subnet as specified by RFC3849 should be used instead

ilario: I didn't know that that RFC existed. 
Anyway I do not agree with the requested change: this is an example of documentation that tries to contain valid commented examples, rather than generic documentation.

PR #671, opened 2020-01-03
Moving pirania master into repo

luandro:


packages/pirania-app/files/www/cgi-bin/pirania/handle_voucher
```diff
@@ -0,0 +1,56 @@
...
+local noJs = params['nojs']
+
+local url_authenticated = uci_cursor:get("pirania", "base_config", "url_authenticated")
+local url_fail = uci_cursor:get("pirania", "base_config", "url_fail")
+local url_info = uci_cursor:get("pirania", "base_config", "url_info")
+
+local url
+
+if (valid < 1) then
+  local db = dba.load(config.db)
```
spiccinini: Here the `config` variable is not loaded from anywhere. Is this working? If it is working then it may be being populated as a global variable by mistake somewhere (but I can't find the place)

luandro: Good catch @spiccinini! None of the `cgi-bin` handles are importing `config`.

Can we merge #672 before fixing this?

spiccinini: Yes, I aproved the PR (needs one more aprove though)

luandro: :+1:

luandro:
Added `require('voucher.config')` to handles. Should we investigate further why it was on global? Also can't find anything...

spiccinini:
> Added `require('voucher.config')` to handles. Should we investigate further why it was on global? Also can't find anything...

Ok I found it, in config.lua it is created a global variable "config" and it is created when the code performs a requre to `voucher.logic`.  I think we should expose a simpler "API to the user" that doesn't need to handle the `dba` object nor the config

packages/pirania/files/usr/lib/lua/voucher/debugtools.lua
```diff
@@ -0,0 +1,39 @@
+#!/bin/lua
+
+debugtools = {}
```
spiccinini: I think we should remove this

luandro: Makes sense, never even noticed it.

luandro:
Ready for merge @spiccinini?

PR #672, opened 2020-01-03
Pirania lime app integration

luandro:
Integration with LiMe-App. Check reviews from [pirania PR](https://github.com/libremesh/pirania/pull/60).


packages/pirania/tests/test_voucher.lua
```diff
@@ -0,0 +1,36 @@
+-- local config = require 'lime.config'
+local dba = require('voucher.db')
+local logic = require('voucher.logic')
+local utils = require('voucher.utils')
```
luandro: @spiccinini am I calling this right?

packages/pirania/tests/test_voucher.lua
```diff
@@ -0,0 +1,36 @@
...
+local key = 'm-belaelu-marcos-android'
+local voucher = '2u8hyfve'
+local epoc = 1579647256049
+local upload = 10
+local download = 10
+local amountofmacsallowed = 1
+
+describe('Pirania tests #voucher', function()
+    it ('create voucher', function ()
+        local db = dba.read_db_from_csv(tempFolder)
```
luandro: Throws error `attempt to call field 'read_db_from_csv'`

packages/pirania-app/files/etc/pirania/governance.json
```diff
@@ -0,0 +1,18 @@
+{
+  "provider": {
```
nicopace: can you describe the structure of this file? it is not clear what provider, community and member are, and how this affects the usage.
would be good to have a general documentaiton

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
```diff
@@ -8,11 +8,11 @@ You may obtain a copy of the License at
 ]]--
 
 require "ubus"
```
nicopace: could this file be called pirania instead of pirania-app?
in that way it would be pirania ubus interface, and it would make sense for others to use it for their own interfaces.

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
```diff
@@ -27,6 +27,35 @@ local function shell(command)
     return result
 end
 
+local function read_governance(msg)
+    local result = {}
+    local path = '/etc/pirania/governance.json'
+    result = utils.readJsonFile(path)
```
nicopace: you don't need to read it as json, you can just print it out, as it is already a json :)

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
```diff
@@ -43,27 +72,15 @@ local function get_clients(msg)
...
-    local file = io.open( path, "r" )
-
-    if file then
-        -- read all contents of file into a string
-        local contents = file:read( "*a" )
-        myTable = json.decode(contents)
-        io.close( file )
-    end
-    printJson(myTable)
+    local result = {}
```
nicopace: sounds like there is a generalization of reading a printing a JSON file (here and in governance you do the same)

packages/pirania/files/usr/libexec/rpcd/pirania
```diff
@@ -19,71 +19,105 @@ local uci = require('uci')
...
-    local result = handle:read("*a")
-    handle:close()
-    return result
+local function status(msg)
+    local output = logic.iptables_status()
+    printJson(output)
+end
+
+local function enable(msg)
+    uci_cursor:set("pirania", "base_config", "enabled", "1")
```
nicopace: is it needed to both set enabled=1 and run captive-portal start?
@spiccinini

luandro: Setting `enabled` makes pirania start on startup. To make it run you need to `captive-portal start`.

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
luandro: There are two different apis. `pirania` deals with all the logic of the `voucher` system, while `pirania-app` deals more specifically with stuff related to the front-end app, such as content. Should it be merged into `pirania`?

packages/pirania-app/files/etc/pirania/governance.json
luandro: Will do.

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
luandro: it needs to be parsed though.

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
luandro: :+1:

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
nicopace: I feel that the '-app' doesn't represent what it is actually doing.
Adding a piece of documentation to it would help figure out why they are grouped the way they are... and may be merging or renaming them. Thanks!

packages/pirania/files/usr/libexec/rpcd/pirania
nicopace: Maybe we can use the /etc/init.d/ scripts for the startup?

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
nicopace: If it is already in the filesystem, it should have been parsed before... so not sure if it is needed

packages/pirania-app/files/usr/libexec/rpcd/pirania-app
luandro: Docs have been added. Lemme know what u think. This decision was made by @altergui and me when we first got pirania working, and I built the vanilla JS app. The distinction between pirania and pirania-app may not make much sense now anymore.

nicopace: it might not be needed anymore indeed

nicopace: thanks for the docs!
It might be good to describe the whole governance.json.
it seems it has some clear structure with specific values that are used in different parts... would be good to know how them affect the use of the app.

nicopace: thanks for the docs!
It might be good to describe the whole governance.json.
it seems it has some clear structure with specific values that are used in different parts... would be good to know how them affect the use of the app.

Issue #558, opened 2019-08-12
How to use network-profiles with OpenWrt buildroot compilation method?


ilario:
The suggested compilation method has changed over the time (buildroot, lime-build, lime-sdk, buildroot).

With the LibreMesh 17.06 release, communities accepted to use [flavors](https://github.com/libremesh/lime-sdk/blob/develop/flavors.conf) (e.g. lime_default, lime_mini, lime_zero...) and [network-profiles](https://github.com/libremesh/network-profiles) which nicely work with [lime-sdk](https://github.com/libremesh/lime-sdk) and [Chef](https://chef.libremesh.org/).

While Chef is still supported, [searching into the mailing list](https://lists.libremesh.org/pipermail/lime-users/2019-May/001467.html) seems that the currently supported local compilation method, changed from lime-sdk to the OpenWrt buildroot. 

The documentation about LibreMesh compilation with OpenWrt buildroot ([here](https://github.com/LibreRouterOrg/openwrt/blob/librerouter-18.06/README.md) and [here](https://github.com/libremesh/lime-packages/blob/master/README.md)) does not mention neither flavors nor network-profiles.

The buildroot method documentation has to be expanded to help the lime-sdk => buildroot transition for the communities.

ilario:
The mechanism employed by https://github.com/aparcar/openwrt-metabuilder is to install the network-profiles as packages downloading them from here http://chef.libremesh.org/network-profiles/ or here http://repo.libremesh.org/network-profiles/ I think we should chose one of these and suggest to the users to use it.
The Chef's one?
@aparcar @p4u

spiccinini:
> While Chef is still supported, [searching into the mailing list](https://lists.libremesh.org/pipermail/lime-users/2019-May/001467.html) seems that the currently supported local compilation method, changed from lime-sdk to the OpenWrt buildroot.
> 
> The documentation about LibreMesh compilation with OpenWrt buildroot ([here](https://github.com/LibreRouterOrg/openwrt/blob/librerouter-18.06/README.md) and [here](https://github.com/libremesh/lime-packages/blob/master/README.md)) does not mention neither flavors nor network-profiles.
> 
> The buildroot method documentation has to be expanded to help the lime-sdk => buildroot transition for the communities.

Thanks Ilario for gathering all this information. I didn't know about network_profiles_builder. 

To support network-profiles in _the buildroot way_ I am inclined to do it with a symbolic link to a local network-profiles repository:
```
$ cd openwrt
[...]

PR #673, opened 2020-01-04
Remove all conditional dependencies

### LINKED ISSUE #331: cascade of metapackages kills modularity

ilario:
As reported by @dangowrt in #331, the conditional dependencies work only when compiling using the buildroot method (which is currently the recommended one, see https://libremesh.org/development.html) but not with methods employing the imagebuilder, like the [Chef](https://github.com/libremesh/chef) or the [metabuilder](https://github.com/aparcar/openwrt-metabuilder) (which are much more convenient and will have to be supported by the next release). 

Fixes #331


G10h4ck:
I do not agree with this changes, chef and image builder should eventually solve their limitation instead.

Also something like chef could be programmed on top of the buildroot instead, which have lot mroe advantages and much less bugs that imagebuilder and similar...

ilario:
> Also something like chef could be programmed on top of the buildroot instead

History is cyclic, they say. In august I commented https://github.com/libremesh/lime-packages/issues/558#issue-479568278

> compilation method has changed over the time (buildroot, lime-build, lime-sdk, buildroot)

and your comment means that we'll have the lime-build again!!!!!

ilario:
In this PR I removed also batctl from lime-debug:
in OpenWrt 18.06 could be annoying but if we add a hard dependency on batctl, also kmod-batman-adv will be selected by lime-debug (as [kmod-batman-adv is a dependency of batctl in 18.06](https://openwrt.org/packages/pkgdata_owrt18_6/batctl))
the opposite happens in OpenWrt 19.07, so there's no need for lime-debug to depend on batctl as it [is a dependency of kmod-batman-adv](https://openwrt.org/packages/pkgdata/kmod-batman-adv).

ilario:
I removed the commit about lime-proto-bmx6 which is better addressed in #716.

ilario:
The following conditional dependencies are addressed in this PR.

- [ ] lime-proto-wan:	+!PACKAGE_firewall:iptables
- [ ] lime-proto-bmx6:	+!PACKAGE_firewall:iptables
- [x] lime-webui:	+PACKAGE_lime-proto-bmx6:luci-app-bmx6
- [x] lime-webui:	+PACKAGE_lime-proto-batadv:luci-app-batman-adv
- [x] lime-debug:	+PACKAGE_lime-proto-batadv:batctl
- [ ] shared-state-dnsmasq_leases:	@(!PACKAGE_dnsmasq-lease-share) (what is this syntax??)  this package does not even exist anymore!!!
- [x] lime-map-agent:	+PACKAGE_lime-proto-bmx6:luci-lib-libremap-bmx6

spiccinini (approved):

aparcar (approved):

Issue #195, opened 2017-08-01
Webui: add link to offline documentation in /www/docs


ilario:
[lime-docs](/libremesh/lime-packages/tree/master/packages/lime-docs) package includes documentation for offline use in LibreMesh firmware images.
The documentation is installed in /www/docs directory (with a symbolic link from /docs), so that it can be accessed by the http server.
A link has to be added in the [lime-webui](/libremesh/lime-packages/tree/master/packages/lime-webui) for accessing this documentation directory.

pedro-nonfree:
I don't know what's the syntax being used (asciidoc?) but I think that in the website should render that format in something more web-based

ilario:
AsciiDoc renderer is written in Python, AsciiDoctor renderer is written in Ruby. I believe that for obtaining HTML pages from AsciiDoc files, one of these two is needed but none can fit in a router.

If we really really want this, the only option I see is to download HTML pages from the website during lime-docs compilation.

nicopace:
I would go for a progressive enhancement approach... a link for now would suffice.
Also, asciidoc was ment to be read in ascii, so we shouldn't bother much about that.
@ilario is that something u can do? (A menu option in luci for accessing the docs).

ilario:
@nicopace I'm not knowledgeable with the web interface, maybe @gmarcos87?

ilario:
@gmarcos87 do you think it's a good or a bad idea to have a documentation tab in the lime-app?

nicopace:
I wouldn't add it like it is in the lime-app as it would create noise to the lime-app users without adding value.
The technical capacity traning part of lime-app needs to get thought through as a project in itself.
I wouldn't add it on the web, would just add a comment on the /etc/banner saying that there is documentation on the router, as most of the doc are more for a technical audience than they are for the average user.

ilario:
> I wouldn't add it on the web, would just add a comment on the /etc/banner saying that there is documentation on the router

Ok, makes sense. New issue for this here: #616 
So now it does not make any sense that the docs are placed in /www/docs and linked in /docs, we can change also that.

ilario:
> I wouldn't add it on the web

Ok, so in #675 I removed it from /www

PR #675, opened 2020-01-09
lime-docs to /docs instead than /www/docs

ilario:
As discussed in #195, there's no interest in adding the docs to lime-app, so the docs location does not need to be `/www/docs` anymore and can be `/docs`.


spiccinini (approved):

nicopace (approved):

ilario:
Thanks for approving but actually I changed my mind in #782 (10 months more recent than this PR) and forgot to close this PR.

PR #676, opened 2020-01-10
lime-proto-batadv VLAN ID start from 50 rather than from 16

### LINKED ISSUE #670: Batman-adv VLAN ID can collide with other protocols' ID
The ID for Batman-adv's VLAN is 16 + a number from 0 to 255, so it can be between 16 and 271 as described here:

https://github.com/libremesh/lime-packages/blob/8a942a174f8f10b0acc6e1c8eb6ad47db0218088/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua#L69-L72

The problem is that olsr2 is set on 16, babeld is on 17 and bmx7 is on 18 so collisions can happen.

ilario:
Read bug report on #670 
Currently, the default Babeld VLAN ID is 17 but Batman-adv VLAN ID is randomized between 16 and 271. This means that when changing the ap_name option (which gets used for generating Batman-adv VLAN ID) there's a probability of 1 over 256 that the configuration will be really broken.
Both this solution and the alternative one (change Babeld VLAN ID) will break retrocompatiblity with routers flashed with recent development code.

Fix #670

### DIFF
--- a/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
+++ b/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
@@ -67,9 +67,9 @@ function batadv.setup_interface(ifname, args)
 	local mtu = 1532
 
 	--! Unless a specific integer is passed, parse network_id (%N1) template
-	--! and use that number + 16 to get a vlanId between 16 and 271 for batadv
+	--! and use that number + 50 to get a vlanId between 50 and 305 for batadv
 	--! (to avoid overlapping with other protocols)
-	if not tonumber(vlanId) then vlanId = 16 + utils.applyNetTemplate10(vlanId) end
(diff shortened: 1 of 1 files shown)


spiccinini:
Great! 
I would use 20 as a start point instead of 50 so we are less likely to break things (many networks deployed with current code). Also we should document that the VLAN ID available to use (for the other protocols) are only up to 19, and from M to 4095.

ilario:
> I would use 20 as a start point instead of 50 so we are less likely to break things

Wait, also 20 would break retrocompatibility in the same way.
I mean, now default Batman-adv VLAN is 16+13=29, which would go either to 50+13 or 20+13.

What would not break it would be something like 
`vlanId = 20 + (utils.applyNetTemplate10(vlanId) - 4) % 256`
I'm going to make this change to the PR.

ilario:
See the alternative PR on #678

Issue #581, opened 2019-09-08
Lets remove unneded vlans from protocol interfaces


spiccinini:
The scenarios we are using in the networks doesn't need vlans for the protocols. Let's remove them to reduce complexity and bugs (like #580 )

ilario:
+1
See also discussion on https://lists.libremesh.org/pipermail/lime-dev/2019-August/001144.html and unconfirmed report on https://forum.openwrt.org/t/mediatek-and-vlan-802-1ad-on-ethernet/42346

G10h4ck:
I agree in removing not neded vlans!

G10h4ck:
It also would solve the problem with mediatek hardware switches

ilario:
So, not needed VLANs are Babeld, BMX*, OLSR* ones, right?
But we would keep the BATMAN-adv one (if I remember correctly, @G10h4ck said that Batman needs to be on an interface by its own, and we use the VLAN=%N1 for breaking the Batman domains in different areas of the networks), right?

spiccinini:
> So, not needed VLANs are Babeld, BMX*, OLSR* ones, right?
Yes

> But we would keep the BATMAN-adv one (if I remember correctly, @G10h4ck said that Batman needs to be on an interface by its own, and we use the VLAN=%N1 for breaking the Batman domains in different areas of the networks), right?

Yes. that's right

spiccinini:
vlan=0 can be used to disable the vlan. I added some fixes for this use in #593

ilario:
Just noticed that Batman-adv has implemented (since long time) some VLAN mechanism on top of bat0 interface, see [here](https://www.open-mesh.org/projects/batman-adv/wiki/Tweaking#VLAN-handling) and [here](https://github.com/openwrt-routing/packages/commit/54af5a209e0a0a75b5eb712c0ca8056e66de02c0).
Does anyone understand if the usage of Batman-adv's VLAN can replace the usage of classical VLANs?
If this is correct, the hardware switches would not recognize these packets as tagged ever, correct?

ilario:
From here https://www.open-mesh.org/projects/batman-adv/wiki/Tweaking#VLAN-handling :

> The batX mesh interface created by batman-adv also supports VLANs which enables the administrator to configure virtual networks with independent settings on top of a single mesh cloud.

which sounds like the broadcast packets from a client will be limited on a single VLAN zone but the Batman-adv hello packages would go everywhere, does anyone know if this is correct?

[...]

PR #678, opened 2020-01-13
lime-proto-batadv VLAN ID start from 20 rather than from 16 and keep old default VLAN ID

ilario:
See the discussion on #676
This PR is an alternative to #676 
Retrocompatibility with the previous release is not going to happen anyway, so this keeps the retrocompatibility just for the users using development code.
At the price of a bit more weird way to calculate the VLAN ID.

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -30,10 +30,10 @@ config lime network
 	list protocols ieee80211s			# List of protocols configured by LiMe, some of these require the relative package "lime-proto-...". Note that if you set here some protocols, you overwrite the *whole* list of protocols set in /etc/config/lime-defaults
 	list protocols lan
 	list protocols anygw
-	list protocols batadv:%N1			# Parametrizable with %Nn (which depends from ap_ssid), note that this will range between 16 and 272
+	list protocols batadv:%N1			# Parametrizable with %Nn (which depends from ap_ssid), note that this will range between 29 and 284
 #	list protocols batadv:0				# If 0 VLAN tags are disabled and the routing is done on the raw interface
 	list protocols bmx6:13				# The VLAN type can be provided as a third argument, for example bmx6:13:8021q for using VLAN 802.1q instead of the default 802.1ad
-	list protocols olsr:14
+	list protocols olsr:14				# Do not use a VLAN ID between 29 and 284 as this range is reserved for batadv:%N1 parameterization, maximum ID is 4095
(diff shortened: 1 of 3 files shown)


G10h4ck:
I'd start from 30, we are already really really near 20...
Have you tested the new code?

spiccinini:
> I'd start from 30, we are already really really near 20...
> Have you tested the new code?

VLAN ids from 0 to 19 are not available? Also above 256 + 20 can be used

G10h4ck:
AFAIR some hardware switch behave badly with big VLAN ids, so the 256 limitation probably comes from there

ilario:
> I'd start from 30, 

If we want to start from 30 we could as well use the easy approach from #676 as the "retrocompatibility" would be lost anyway (remember that currently the default BATMAN-adv VLAN ID is 29).

> we are already really really near 20...

What's the problem? We can start using the VLAN ID between 1 and 12 (using 802.1ad, while I suppose that if we wanted to use 802.1q we should stay away from small numbers, correct?) or between 276 and 4095.

> Have you tested the new code?

I didn't but now I did :)
Using the default ESSID "LibreMesh.org" still results in the ID 29 which is what differentiates this PR from #676.
Using AP_SSID "y" the %N1 is zero and the ID is 272.
Using AP_SSID "ooo1" the %N1 is 3 and the ID is 275.
Using AP_SSID "xxxx1" the %N1 is 4 and the ID is 20.
Using AP_SSID "p1" the %N1 is 255 and the ID is 271.
So, it works as expected.

PS: I just noticed a little stupid thing: the MD5 we use for calculatin %N1 in lime-system/utils.lua takes also a newline after the AP_SSID, so we're using the MD5 of "LibreMesh.org\n" instead of that of "LibreMesh.org". But this has no impact at all and I would not change this.

PPS: please read and comment if possible on my comment about BATMAN-adv's own VLAN mechanism here: https://github.com/libremesh/lime-packages/issues/581#issuecomment-573362953

G10h4ck:
> > I'd start from 30,
> 
> If we want to start from 30 we could as well use the easy approach from #676 as the "retrocompatibility" would be lost anyway (remember that currently the default BATMAN-adv VLAN ID is 29).

ok so lets start from 29

ilario:
> ok so lets start from 29

Done

ilario:
@spiccinini I added that line :) thanks

spiccinini:
Thanks @ilario sorry for the delay

PR #680, opened 2020-01-22
Pirania tests improvements and other minnor changes

spiccinini:
Now each pirania test is independent of the other tests. 
Also this fixes the tests as they were not passing in Travis due to some hostname or something dependent on the host


spiccinini:
I have more improvements that will add in a new PR

packages/pirania/files/usr/lib/lua/voucher/logic.lua
```diff
@@ -191,9 +191,9 @@ function logic.check_voucher_validity(voucherid, db)
     return res
 end
 
-local function setIpset(mac, expiretime)
-    -- ipset only supports timeout up to 4294967
-    if tonumber(expiretime) > 4294967 then expiretime = 4294967 end
+function logic._setIpset(mac, expiretime)
```
luandro: Is there any special reason for the underscore? Breaks the overall naming style.

spiccinini: The rationale is that as that function was a `local` only function and that now it is exposed at the module level but the idea is that it is not part of the AP.  The leading underscore is a Python convention to say "this is not part of the API", I don't know any special convention for this in Lua so I used python's.

PR #681, opened 2020-01-25
Workaround subtle MTU bugs due to vlan usage

G10h4ck:
As of today many ethernet drivers on the routers we use doesn't seem
  capable to properly handle mini-jumbo frames, most of them seems capable
  to handle ethernet frames up to 1504 bytes (plain ethernet 1500 + vlan 4).
  In most router those 4 bytes extra benefit get used by the hardware
  switch 802.1q vlan configurations, so when LiMe add it's own (usually
  802.1ad) vlan the required MTU to be totally trasparent would be 1508
  which is not supported on many routers in our deployments.
  To deal with that limitation we reduce the proto vlan interface MTU to 1496,
  but as PMTU discovery, on which TCP rely, is usually broken in IPv4 and as a
  consecuence TCP end up wrongly assuming that the PMTU is 1500, we need to
  mangle some TCP packets to workaround that issue, and avoid a part of the
  packets (the ones that end up being 1497-1500 big) being silently lost and
  TCP to hang.


G10h4ck:
I have tested it on a tl-wdr4300 with LibreRouterOS v1.1 and got the expected result, if someone using OpenWrt firewall fw3 would test and report it would be welcome too.

spiccinini:
How can I reproduce the issue and test if this works? With a client connected through ETH directly to the Node?

G10h4ck:
I found out the bug conneting via SSH from atatopo to czuk, and apparently the packets were being droppend in lapraviana,  so it was probably like this  atatopo ethX mtu 1500 <- ethernet cable -> switch <- ethernet cable -> Wifi device <- Wifi link -> Wifi device <- Ethernet cable -> ethX.Y-Z mtu 1492 lapraviana ethH.K-Y mtu 1492 <- ethernet cable -> Wifi device <- Wifi Link -> anygw mtu 1500 Czuk-bbone

but I am not sure of what would be the minimal setup to reproduce the issue

Issue #93, opened 2017-03-31
Babeld check internet access before sharing [was: Babeld gateways prioritization [was: ForceWAN Web Interface]]


nicopace:
A user wants to configure what happens when he has WAN access. The options would be to force share, force not to share, or let the algorithm choose what to do.

In spanish: wan autodetecte que tiene internet, o que lo fuerces a que no o a que si.

nicopace:
@p4u @altergui is the isue well defined? I got this from a postit on the board today and have failed to get a clear idea of what it ment.

ilario:
I think it is clear enough. An example: you plug internet into WAN port of your LiMe router and use WiFi from LiMe router for accessing internet in your house, but you don't want your internet to be shared with the rest of the LiMe network. Or even you plug in but by default you don't want to use it for accessing the internet (e.g. because or erratic connection performances due to ISP).

altergui:
This was actually misunderstood, sorry.
The original idea when we wrote the postit is this:
a user connects a LiMe router WAN port to a ADSL, but the ADSL provider filters out ICMP pings. (I have personally encountered several cases like this). Everything works fine, besides ping. Thus, watchping doesn't realize there's "Internet" connectivity, and doesn't trigger the bmx6 tunIn announcement.
Resulting behaviour reported by the user: The LiMe router gateway provides Internet access (if you directly connect over wifi to it) but it doesn't "share" it with the rest of the mesh (i.e. when you connect to any other node)

This would be solved by a "simple checkbox" in the web interface which basically tells a router "Please announce that you have Internet connectivity to the rest of the mesh, even if you can't detect it with your poor watchping method"
Behind the scenes, it would basically force the tunIn announcement in bmx6 (just like watchping does)

ilario:
Ok, so the problem would be mitigated also with a solution like replacing watchping as discussed on #66.
In my opinion both what you wrote and what I wrote are worth to implement, they would fit in two different radio buttons, something like this, in the webui:

Use directly connected gateway:

- always, without checks
- depending internet availability
- never

Share direct internet access with other nodes:

- yes, allow other nodes to use this as gateway
- no, just clients of this router can use it as gateway

ilario:
Coming back to the original scope of the issue: announce internet access without checking, some ways to do this:

[...]

PR #684, opened 2020-01-31
ubus-lime-metrics: use http instead of icmp to detect internet

spiccinini:
Some ISP drop some ICMP packets (like ping), so using http is better to detect internet (and also http/tcp is more reliable than icmp).

### DIFF
--- a/packages/ubus-lime-metrics/files/usr/sbin/last_internet
+++ b/packages/ubus-lime-metrics/files/usr/sbin/last_internet
@@ -8,7 +8,7 @@
 
 result=''
 # if 8.8.8.8 is accessible, print internet path
-if ping -q -c5 -w10 8.8.8.8 &>/dev/null; then
+if wget http://detectportal.firefox.com/success.txt -q -O - &>/dev/null; then
     #BMX6 Path to INternet
     if [[ -f '/usr/lib/opkg/info/lime-proto-bmx6.control' ]]; then 
         default_dev=`ip r | grep "default dev" | cut -d' ' -f3`;


ilario:
What if we implement an utility doing this?
So that we can use that both here and in the watchping (see #93).
For the util, it would be useful to also checking if the content of the discharged file is the correct one, so that a captive portal does not make a false positive, do you agree?

gmarcos87:
It's great, especially it will improve the experience of mesh networks near airports hahaha

nicopace:
I think @ilario comment has some merit.
There are some ISPs that sell metered connections (mainly mobile-based), so when your credit runs off a captive portal-like interface appears (basically, http traffic gets redirected).
I think it is an important addition to be able to detect these situations so the communities can react to it by adding more credit for example.

nicopace:
> What if we implement an utility doing this?
> So that we can use that both here and in the watchping (see #93).
> For the util, it would be useful to also checking if the content of the discharged file is the correct one, so that a captive portal does not make a false positive, do you agree?

The only thing is that the detection is not related to the content of the request, but on the HTTP return code.

You can check out the captive portal strategies here:
* https://en.wikipedia.org/wiki/Captive_portal
* https://rootsh3ll.com/captive-portal-guide/

packages/ubus-lime-metrics/files/usr/sbin/last_internet
```diff
@@ -8,7 +8,7 @@
 
 result=''
 # if 8.8.8.8 is accessible, print internet path
-if ping -q -c5 -w10 8.8.8.8 &>/dev/null; then
+if wget http://detectportal.firefox.com/success.txt -q -O - &>/dev/null; then
```
G10h4ck: It is better we check also the content of the file being "success" because otherwise in case fo having a captive portal in the middle the node will still think internet is available

spiccinini: Yes :) I will push a new version, probable as a new script "check_internet` or something like that.

gmarcos87: If there is a captive portal it must still be considered a gateway. Many times these captive portals are a notification that must be seen once (lack of payment for example) or login screens to the service. If we don't consider them as gateways they can end up being more conflictive than we think.

nicopace: But if you have many gateways, and this one is being captive-portaled, it shouldn't be chosen as a gateway so the others could be enabled.
It should definitely be a third state of "gateway unavailable", where we notify the user that though it seems there could be a gateway on the other side, it is not possible to access it right now.

spiccinini:
Closing in favor of #712

PR #686, opened 2020-02-05
utils: adds unsafe_shell() and shell_quote() functions

spiccinini:
We have multiple shell() functions all over the code and in some places the use is not safe. Let's face that it is a very handy practice but let's do it "safe". So I propose to add a new `utils.unsafe_shell()` function. The idea of having the unsafe word in its name is to encourage the developer of taking it carefully. 
Also adding `utils.shell_quote()` that is based on Python's shlex.quote(). Using unsafe_shell + shell_quote may be considered "safe".

After this PR is merged I will push changes to replace the actual shell calls by this new functions.


ilario:
Wow! Looks great! Thanks

PR #688, opened 2020-02-11
LimeApp updated to v0.2.3

gmarcos87:


gmarcos87:
This fix thisnode.info as default api. @germanferrero

Issue #689, opened 2020-02-12
luci dispatcher exception when entering thisnode.info 


germanferrero:
When entering thisnode.info, I'm first redirected to http://thisnode.info/cgi-bin/luci/ and this error is shown:
`/usr/lib/lua/luci/dispatcher.lua:319: No valid theme found
stack traceback:
	[C]: in function 'assert'
	/usr/lib/lua/luci/dispatcher.lua:319: in function 'dispatch'
	/usr/lib/lua/luci/dispatcher.lua:127: in function </usr/lib/lua/luci/dispatcher.lua:126>`

To reproduce this behavior I modified feeds.conf.default of https://github.com/LibreRouterOrg/openwrt/commit/7a739bf1de88cbc0e1febf659c693e2917e2c1ff
to point libremesh feed first to
https://github.com/libremesh/lime-packages/commit/306df423c7a653a5b4ae1899db19ac527c66e627
and then to
https://github.com/libremesh/lime-packages/commit/cca04284b5418d9865634c549befba250f32b542
Getting the same error on both.
While it's working correctly when pointing the feed to
https://github.com/libremesh/lime-packages/commit/3dd7c9780ebf96c2cd94b596cc4ef64e1654482b

I suspect that removing these lines could have introduced the error:
https://github.com/libremesh/lime-packages/commit/cca04284b5418d9865634c549befba250f32b542#diff-9be046bb897dc4ddd137de4ef4e65a04L42-L44

ilario:
> I suspect that removing these lines could have introduced the error:
> [cca0428#diff-9be046bb897dc4ddd137de4ef4e65a04L42-L44](https://github.com/libremesh/lime-packages/commit/cca04284b5418d9865634c549befba250f32b542#diff-9be046bb897dc4ddd137de4ef4e65a04L42-L44)

Yesss, that looks correct @germanferrero, those two lines were added in #651 which solved this exact problem.

PR #690, opened 2020-02-12
recover limeapp index page

germanferrero:
closes https://github.com/libremesh/lime-packages/issues/689


ilario:
Fixes #609

germanferrero:
I have noticed that the regression was caused by an automatic version upgrade at travis CI job:
https://github.com/libremesh/lime-packages/commit/7c1dc6358d184c4328a63d4ea6f6ce29886b7660#diff-6025d3631b4e5bf5aa0020a675ed9003
How ever I couldn't understand why travis ignored debf02fca3982d4668bed69aed5a7d19329ee121 and aebd8570c43fc559acc68d6dc6264c36bda77af0

I just let this as a note in case it happens again.

PR #691, opened 2020-02-22
lime-utils: fix get_node_status hostname bug recently introduced

spiccinini:
So I introduced this bug. Let's get rid of using uci to get the hostname.
(Without this the lime-app doesnt display the status page as the hostname is missing)


packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils
```diff
@@ -51,8 +51,7 @@ end
 
 local function get_node_status(msg)
     local result = {}
-    local uci = config.get_uci_cursor()
-    result.hostname = uci:get("system", "hostname")
+    result.hostname = io.input("/proc/sys/kernel/hostname"):read("*line")
```
germanferrero: I understand that the bug is introduced because system is an unnamed section, and we should get the hostname with something like this:
uci.cursor():foreach("system", "system", function(s) hostname = s.hostname end)
which is awful.

Maybe a good lesson could be to avoid unnamed sections in uci configuration files?

germanferrero: Moreover it seems we have been already fighting with this config at https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/usr/lib/lua/lime/system.lua

Should we use this util here?

spiccinini: @germanferrero I think that getting the hostname from the uci system instead of the current hostname should be only done "while we are reconfiguring the system" (as it is performed in lime.system.get_hostname, etc). But probably is a good idea to refactor this  `get_hostname` into the main lime.utils module.

Yes you are right that using unnamed sections for sections that are unique is a bad idea... I don't know why openwrt does that in this config (this is the openwrt system config file).

PR #692, opened 2020-03-08
fixed missing config from pirania logic

luandro:
Wasn't importing the config for Pirania logic. That should be fixed now.


nicopace:
It doesn't seem to be required in that file... why would you be importing it if you are not using it?

spiccinini:
> It doesn't seem to be required in that file... why would you be importing it if you are not using it?

Yes it is used in multiple places (as config.db).

nicopace:
> > It doesn't seem to be required in that file... why would you be importing it if you are not using it?
> 
> Yes it is used in multiple places (as config.db).

wow... my bad... clicked a button expecting github to give me the whole, when it actually showed me just 10 more lines heheh..

Approving.

PR #697, opened 2020-03-24
lime-report fix dmesg command and add info on firewalls

ilario:
There were two errors in the capture of the dmesg output in lime-report utility: tail was using -c option instead of -n and the fact that the command included a pipe made necessary to enclose the command in quotes.

### DIFF
--- a/packages/lime-report/files/lime-report.sh
+++ b/packages/lime-report/files/lime-report.sh
@@ -58,9 +58,9 @@ generate_status() {
     paste_cmd iwinfo
     paste_cmd wifi status
     paste_cmd swconfig dev switch0 show
-    paste_cmd iptables -vL -t filter
-    paste_cmd iptables -vL -t nat
-    paste_cmd iptables -vL -t mangle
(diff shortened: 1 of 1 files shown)


ilario:
I added also the output of the rules of iptables and ebtables.

ilario:
> Please add also -n option to iptables commends so we can get some statistics on how much packets/bytes matched rules

The `-n` option of `iptables` does not help with that. Did you mean the `-n` option for having "numeric output of addresses and ports". What you mention, "get some statistics on how much packets/bytes matched rules", is already achieved using the `-v` option. For `ebtables`, what would make sense is to add `--Lc`, what do you think?

G10h4ck:
Anyway I'd add the -n option to iptables, it avoid the reverse lookup which take lot of time and is unuseful for the report. +1 to add the ectables option to show statistics too

ilario:
@G10h4ck I applied the requested changes, please review

G10h4ck (approved):

PR #700, opened 2020-04-14
Add root password management

spiccinini:
This PR allows to configure how to manage a root password, adding a mechanism to generate  a random password and also one to use a fixed password. This fixed password is handy to use as a community shared password for administrative purposes. This mechanism is executed when running `lime-config`, so also at least once at first boot. 
In another PR I will add support to FirstBootWizard for the fixed shared password.

From the docs of the options

* `option root_password_policy 'DO_NOTHING'`
    When configured to SET_SECRET, the root password secret will be configured as specified in root_password_secret. When set to RANDOM a strong random password will be set if root has no password, use this is your firmware is built with the ssh keys inside. DO_NOTHING does nothing, leaving the root password empty (you will have to set it manually or through FirstBootWizard).
* `option root_password_secret '$1$K0.R51EI$bkjIoF5feHcpqCuEOrymB.'`
    This is the password hash as stored in /etc/shadow, it is only used when root_password_policy=SET_SECRET. You can generate the secret with 'openssl passwd -1' to be compatible with most openwrt firmwares, use a strong password with at least 10 numbers and letters, the longer the better!. For improved security use "openssl passwd -6" for SHA512 (or -5 for SHA256) but be aware that not all firmwares support this.


spiccinini:
Thanks @germanferrero for the review. I have also already tested it now in a device, so I am removing the WIP.

packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils-admin
```diff
@@ -0,0 +1,43 @@
...
+local ubus = require "ubus"
+local json = require 'luci.jsonc'
+local utils = require 'lime.utils'
+
+local conn = ubus.connect()
+if not conn then
+    error("Failed to connect to ubus")
+end
+
+local function set_root_password(msg)
```
germanferrero, line 18: @spiccinini this is the ubus method we want to expose in order to let a limeapp user to change the shared-password?
In that case we should be writting the root_secret option at lime-community right?

spiccinini: Great catch :)

spiccinini: I've addressed this in #701

packages/lime-system/files/usr/lib/lua/lime/system.lua
```diff
@@ -15,12 +15,36 @@ function system.get_hostname()
...
+		local secret = config.get("system", "root_password_secret")
+		local current_secret = utils.get_root_secret()
+		if current_secret == nil then
+			error("Can't get root password")
+		end
+		if current_secret ~= secret then
+			utils.set_root_secret(secret)
+		end
+	elseif policy == "RANDOM" then
+		if #utils.get_root_secret() <= 1 then
```
G10h4ck, line 40: `<= 1` it's because first line might be an empty line on `\n` if not set? In that case I think this should be handled directly inside `utils.get_root_secret()` returning `nil` if the first line of `/etc/shadow` is empty/invalid

spiccinini: Some possible options are `*` (password never set, and empty) or `!` account locked. For example:
```
root::0:0:99999:7:::
daemon:*:0:0:99999:7:::
ftp:*:0:0:99999:7:::
```
I can add a comment here to clarify this.

G10h4ck: ok good to comment it ;)

Issue #612, opened 2019-10-04
Homogenize adhoc_ssid between FirstBootWizard and lime-defaults-factory


ilario:
adhoc_ssid in FirstBootWizard code and in lime-defaults-factory should be identical and they are not.
See
https://github.com/libremesh/lime-packages/blob/e9efbea1003c8e333f5d34c220274d13d0ce5550/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua#L283
and 
https://github.com/libremesh/lime-packages/blob/18c8aa59ec14bf8a2e82f72b0b761194d525213c/packages/lime-system/files/etc/config/lime-defaults-factory#L57

I can see no reason for having that line in FirstBootWizard code (and the next about ieee80211s_mesh_id), deleting it may fix.
Additionally, those two lines could break communities setting a custom adhoc_ssid and ieee80211s_mesh_id in their lime-defaults or network-profiles.

ilario:
The ssid has been homogenized sometimes in the past.
Anyway there are other settings that look different:

https://github.com/libremesh/lime-packages/blob/606ae4c6b02118571a3d312b6426d1c57e4e796c/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua#L155

https://github.com/libremesh/lime-packages/blob/9d9837d2809c5fae1de3c3b804161a77d70fc785/packages/lime-system/files/etc/config/lime-defaults#L63-L66

https://github.com/libremesh/lime-packages/blob/606ae4c6b02118571a3d312b6426d1c57e4e796c/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua#L149

https://github.com/libremesh/lime-packages/blob/9d9837d2809c5fae1de3c3b804161a77d70fc785/packages/lime-system/files/etc/config/lime-defaults#L54-L55

Can't FBW just take the values from lime-defaults config file instead of having hardcoded ones?

spiccinini:
Related to this issue https://github.com/libremesh/lime-packages/pull/859

PR #701, opened 2020-04-15
FBW community shared password network creation

spiccinini:
Allow  the creation of a network with a shared community admin password
Also `fbw.apply_user_configs(configs. hostname)` is changed by a clearer `fbw.create_network(ssid, hostname, password)`

This is to be applied on top of PR #700 , so only the last few commits are new to review.

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -255,22 +255,23 @@ function fbw.read_configs()
 end
 
 -- Apply configuration for a new network ( used in ubus daemon)
-function fbw.apply_user_configs(configs, hostname)
-    fbw.log('apply_file_config(ssid=' .. configs.ssid .. ', hostname=' .. hostname .. ')')
+function fbw.create_network(ssid, hostname, password)
+    fbw.log('apply_file_config(ssid=' .. ssid .. ', hostname=' .. hostname .. ')')
     local uci_cursor = config.get_uci_cursor()
-    -- Mesh network name
-    local name = configs.ssid
-    -- Format hostname
-    hostname = hostname or config.get("system", "hostname")
+
     -- Save changes in lime-community
-    uci_cursor:set("lime-community", 'wifi', 'ap_ssid', name)
-    uci_cursor:set("lime-community", 'wifi', 'apname_ssid', name..'/%H')
-    uci_cursor:set("lime-community", 'wifi', 'adhoc_ssid', 'LiMe.%H')
-    uci_cursor:set("lime-community", 'wifi', 'ieee80211s_mesh_id', 'LiMe')
+    if password ~= nil and password ~= '' then
+        lutils.set_password('root', password) -- this takes 1 second, it may be replaced with nixio.crypt(password, '$1$vv44cu1H')
+        uci_cursor:set("lime-community", 'system', 'root_password_policy', 'SET_SECRET')
+        uci_cursor:set("lime-community", 'system', 'root_password_secret', lutils.get_root_secret())
+    end
+    uci_cursor:set("lime-community", 'wifi', 'ap_ssid', ssid)
+    uci_cursor:set("lime-community", 'wifi', 'apname_ssid', ssid..'/%H')
     uci_cursor:commit("lime-community")
+
     -- Apply new configuration and setup hostname
     config.reset_node_config()
-    uci_cursor:set("lime-node", 'system', 'hostname', hostname)
+    uci_cursor:set("lime-node", 'system', 'hostname', hostname or config.get("system", "hostname"))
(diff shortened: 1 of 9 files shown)


ilario:
This looks great! Thanks @spiccinini!
I still have to test, but this is a big step for solving #327

packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -296,4 +297,35 @@ function utils.set_password(username, password)
 									pass, pass, user))
 end
 
+function utils.get_root_secret()
+	local f = io.open(utils.SHADOW_FILENAME, "r")
+	if f ~= nil then
+		local root_line = f:read("*l") --! root user is always in the first line
```
nicopace, line 303: And if it is not in the first line?
Seems an unnecessary simplification

nicopace: From https://linuxize.com/post/etc-shadow-file/

> Typically, the first line describes the root user, followed by the system and normal user accounts.

so it is not forced to be in the first line.

spiccinini: Thanks for the review!!!
Yes!  Albeit it is not forced, that is the way it is generated in OpenWrt (and in most linux systems, as this is part of the "first boot/bootstrap" of the system). Some user may change the order manually but it is very rare to edit this file by hand.
At the time of writing these lines of code, I have chosen to write a way simpler but not always correct code (specially the `set_root_secret`). I am open to write the longer but correct version. Thoughts?

nicopace: ok with me :)

spiccinini:
As we changed the API in fbw to match the new api in lime-app, let's try to land both changes close in time. 
Don't merge this yet.

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -255,22 +255,21 @@ function fbw.read_configs()
...
-    -- Mesh network name
-    local name = configs.ssid
-    -- Format hostname
-    hostname = hostname or config.get("system", "hostname")
+
     -- Save changes in lime-community
-    uci_cursor:set("lime-community", 'wifi', 'ap_ssid', name)
-    uci_cursor:set("lime-community", 'wifi', 'apname_ssid', name..'/%H')
-    uci_cursor:set("lime-community", 'wifi', 'adhoc_ssid', 'LiMe.%H')
-    uci_cursor:set("lime-community", 'wifi', 'ieee80211s_mesh_id', 'LiMe')
```
nicopace, line 269: what happened with these two? shouldn't be used still?

packages/first-boot-wizard/tests/test_firstbootwizard.lua
```diff
@@ -7,6 +7,7 @@ local fs = require("nixio.fs")
 local fbw_utils = require('firstbootwizard.utils')
 
 local uci = nil
+config.log = function () end
```
nicopace, line 10: what is the purpose of this one? doesn't seem to be used anywhere, nor adding any useful functionality

spiccinini: This is to silent some logging performed by different modules during the tests. We should have a config.disable_logging() or something link that.

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
spiccinini: I don't understan why they where needed to be here in the first place. Maybe there were no defaults?
Currently the defaults (in `/lime-system/files/etc/config/lime-defaults`) are:
`option ieee80211s_mesh_id 'LiMe'`
`option adhoc_ssid 'LiMe'`
Note that the adhoc_essis default wa different in FBW. I also don't know really why. Anyway, adhoc is not used by FBW so it is useless here.

nicopace: agreed.

packages/first-boot-wizard/tests/test_firstbootwizard.lua
nicopace: sounds good ... sorry that I can't help with that :(

spiccinini: Thanks for reviewing! it is very helpful

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
ilario: related issue: https://github.com/libremesh/lime-packages/issues/612

PR #702, opened 2020-04-25
Qemu development tools

### LINKED ISSUE #389: create easy, reliable virtual machine workflow for testing
Currently there are three sets of instructions to run LibreMesh in a virtual machine, and none of them work for me. I use Manjaro (Arch derivative). Right now I need this in order to test language handling and translation of LuCi and LimeApp, so any option that doesn't let me access those things isn't useful.

1. in the lime-sdk Readme https://github.com/libremesh/lime-sdk/blob/master/README.md#using-plain-qemu. This part is incomplete: "Using plain QEMU. Plain qemu can be launched straight from the command line, if you don't need to access LibreMesh web interface and just want to have a shell you can issue [...] You can also have access to the web interface configuring a tap device on the host as follows. TODO". Then there's the part in "Using Virt-Manager" that says "Now you have to create a new virtual machine: click on File/New Virtual Machine, select Import existing disk image and choose the arm architecture under Architecture options and virt machine type." But when I launch Virt-Manager, I don't see the option "Import existing disk image".
2. by @gmarcos87 https://github.com/gmarcos87/lime-virtualbox These instructions get me all the way to a VM mesh node that's visible to other nodes on my network, but its web interfaces won't load (LuCi and LimeApp) when I go to the VM's IP in my browser. And sometimes it doesn't finish booting, so I restart until it finishes booting.
3. by @nicopace https://blog.freifunk.net/2017/07/22/streamline-app-development-openwrt/ this doesn't work, gives an error `qemu-system-x86_64: -enable-kvm: unsupported machine type. Use -machine help to list supported machines`.

Can we figure out something that works more reliably? At least on Debian and Arch-based distros? And preferably on Mac and Windows too, since some people will probably find it useful someday. I don't know much about the options, but maybe Docker, Vagrant or LXC are options? I already use Docker to run `cooker`. LXC was promised to be the future of lighweight virtualization years ago, but I haven't seen it in use.

spiccinini:
This PR adds two scripts to ease running qemu images. 
Am option `--libremesh-workdir` is provided to dynamicaly modify the rootfs with a local libremesh repo so that qemu boots directly with the development files. This is very handy to test boot scripts, etc. It is not perfect as it does not delete the old files already present in the images. To support this I modified all the packages to follow the "files directory scheme".

I have built and uploaded images to repo.libremesh.org that will work with this scheme.

I still did not investigate rootfs persistance, so it is always a new fresh boot.

Notes:
* I first started with docker. This worked great until I wanted to have more control over the network configuration.
* In the time I had to spend in this I wasn't able to use only a ramfs image or a vmlinuz and rootfs. So both a ramfs (rootfs + kernel) and rootfs.tar.gz has to be used.
* Should fix #389 
* Inspired on @nicopace work here: https://github.com/libremesh/lime-packages/issues/389#issuecomment-496774005


packages/first-boot-wizard/Makefile
```diff
@@ -24,7 +24,6 @@ endef
 
 define Package/$(PKG_NAME)/install
 	$(CP) ./files/* $(1)/
-	@chmod a+x $(1)/etc/init.d/firstbootwizard
```
nicopace, line 27: I don't think this is related with this PR

spiccinini: To correctly support the "files structure" permissions of files need to be already as they have to be in the target. That's why I changed the permissions of multiple files and removed the now unneded step in the makefile

packages/ubus-lime-utils/Makefile
```diff
@@ -23,7 +23,6 @@ endef
 define Package/$(PKG_NAME)/install
 	$(INSTALL_DIR) $(1)/
 	$(CP) ./files/* $(1)/
-	@chmod a+x $(1)/usr/libexec/rpcd/lime-utils
```
germanferrero, line 26: I was trying out this branch and founded out that we missed to setup execution permissions to lime-utils-admin.
Let's add it in this PR?

germanferrero: I opened this PR in case we prefer to address this issue in its own PR: https://github.com/libremesh/lime-packages/pull/704

spiccinini: Thanks! I cherry picked your commit

Issue #189, opened 2017-07-25
"received packet with own address" batman-adv error in random occasions


ilario:
I observed this old error (see [this](https://lists.open-mesh.org/pipermail/b.a.t.m.a.n/2014-March/011839.html) and [this](https://lists.libremesh.org/pipermail/lime-dev/2014-March/000308.html) old emails, the problem was solved here: ae0672b) happening again but only on first boot.

Only on the first boot, bat0 interface doesn't take the dummy0 interface as main one, so the resulting bat0 MAC address so is taken from an actual interface in bridge with bat0 and results in this error.

I think also @nicoechaniz saw this.

ilario:
I saw this on TP-Link WDR3600, but I didn't have time to confirm recently.

altergui:
I've seen this happen in random occasions, (not particularly on first boot)
it looks like a race condition to me, i.e. at each boot, there's a small chance that the first interface added to bat0 is not dummy0, so the "own address" error will be logged every 10 secs until reboot.

Maybe this small chance is bigger (or even equal to 1) on first boot, as you report. But i think it's irrelevant since i've seen this happen in non-first boots as well. So i would close this as "wontfix" (i.e. "can't really fix")

ilario:
Ok, let's close. I hope that we will find a way to fix this eventually (isn't there a way to delay the up of eth0? Something like a sleep 1` in ifup`?)

ilario:
In the router I'm using now the message in dmesg was always present.

I tried to fix this inserting a sleep in `/etc/hotplug.d/iface/00-netstate` for all interfaces but dummy0 (it was something like 
```
echo "$INTERFACE" | grep -q dummy
[ "$?" = 1 ] && {
	sleep 1
}
```
but this didn't change the order of activation of the interfaces, just made the process slower.

So I solved in another unclean way, setting in /etc/config/network
```
config interface 'lm_net_eth0_2_batadv_if'
	option auto '0'
```
and enabling the interface later adding in `/etc/rc.local` a line with `ifup lm_net_eth0_2_batadv_if`.
This is a dirty solution, a clean one would be to understand how netifd dependencies work (I couldn't find enough documentation).

nicoechaniz:
if you see a non stop stream of these messages you may be experiencing the "Joker node" sindrome.

TP-Link WDR3500 and 3600 have been known to enter this state after an electrical problem due to lightning or other source.

[...]

Issue #56, opened 2016-07-21
Batman-adv ethernet meshing


p4u:
Currently, by default, lime-config do not use BAT-ADV in ethernet interfaces. So the expected behavior would be to have eth0.15 (for instance) which runs BAT-ADV, but this was removed in some point because there were supose to be problems with BAT-ADV.

I would like to review this issue and if it does not exist anymore enable again the bat-adv ethetrnet mesh.

p4u:
Thinking loud: this current setup can also have some negative impact on the DHCP lease share.

Since the ethernet attached nodes are bridged and bat-adv does not know about this connection (not sending OGM packets over it), it might be a situation where ALFRED is not sharing correctly the DHCP leasse file over the bridged nodes thus there might be duplicated IP addresses on the network.

egon0:
i'm heavily using batman-adv via ethernet on our freifunk jena mesh at the moment, i can't see any drawbacks, especially alfred seems to work like a charm. but we use a old batman-adv version (a patched 2013.4)... if we could make a branch with enabled batman-adv-lan-meshing i maybe could test it in the next days.

dangowrt:
On Mon, Jul 25, 2016 at 05:02:53AM -0700, Micha Stöcker wrote:
> i'm heavily using batman-adv via ethernet on our freifunk jena mesh at the moment, i can't see any drawbacks, especially alfred seems to work like a charm. but we use a old batman-adv version (a patched 2013.4)... if we could make a branch with enabled batman-adv-lan-meshing i maybe could test it in the next days.

it surely works if you *only* use batman-adv over ethernet. however,
as batman-adv adds some encapsulation overhead, that either harms
performance (fragmentation) or needs the device's on-board switch to
be able to handle MTU > 1500 properly, which is not the case for all
devices (I had problems e.g. with TL-W8970B having an Infenion
switch IC).
For that reason it makes more sense to run a bridge in one VLAN and
batman-adv in another VLAN. BLA should prevent loops, however, there is
a race-condition if both, batman-adv and the bridged VLAN share the
same physical media. This was discussed during battlemesh this year and
we were wondering if Simon or Linus already came up with a fix.

ilario:
Batadv on ethernet was disabled in [this](/libre-mesh/lime-packages/commit/dd07aeffcca3aeafbe815ad663f182ff443011e5) commit.

altergui:
[...]

Issue #658, opened 2019-12-01
Filter DHCP offers from other nodes on LAN


ilario:
This issue was first reported by @nordurljosahvida [on the lime-users list here](https://lists.libremesh.org/pipermail/lime-users/2017-July/000751.html) and on a similar issue: #172.
In the original report, a commercial router was connected to the LAN port, and to fix this was enough to move it on the WAN port.
Instead, in this case I observed it on a isolated two LibreMesh nodes network while configuring them one by one using FirstBootWizard.

The setup:

- two LibreMesh nodes are connected via ethernet cable at their LAN ports
- my laptop is connected on the AP of one of the nodes
- the two LibreMesh nodes have different ap_ssid AP ESSID (to me it happened simply because one was already configured with a custom ESSID while the other was not yet configured (had the default "LibreMesh.org" ESSID))

The problem:

- the DHCP offers from both nodes reach my client, my client considers both valid and takes a random one
- the two DHCP offers suggest a different network domain and a different gateway
- when the DHCP offer from the wrong node (the distant one) gets accepted, my client cannot reach the gateway (e.g. I am connected to the AP of a node with 10.13.0.1/16 but my laptop accepts an IPv4 in the 10.249.0.0/16 network, and cannot reach the announced 10.249.0.1 gateway)

What I would expect:

some firewall rule should kill the DHCP requests or the DHCP offers going through br-lan

I thought that this ebtable rule (added 6 years ago in 920ba9953abd05be47347b1b804416cb3a40b19f) was in charge of filtering these offers, but seems that is not working:
https://github.com/libremesh/lime-packages/blob/f33051f68fb931a66026e228719a194e09e4f2b7/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua#L66

ilario:
Running various times the DHCP client on my laptop, the IPv4 can change from one network to another randomly with some stickiness:

at this point I was connected to a "libremesh.org" AP with IPv4 10.249.x.x but the DHCP offer that gets accepted by my laptop if the one of the other "LibreMesh.org" node connected to the first node by cable on the LAN port, likely just because it was received few milliseconds before
```
# dhcpcd
dev: loaded udev
wlp3s0: connected to Access Point `libremesh.org'
enp0s25: waiting for carrier
DUID 00:01:00:01:22:29:54:8f:54:ee:75:7a:c2:1f
wlp3s0: IAID ea:06:f3:6f
[...]

PR #703, opened 2020-04-28
Testing Batman-adv on ethernet

ilario:
I did not test this PR yet, please test or wait before testing.

In dd07aeffcca3aeafbe815ad663f182ff443011e5, the ethernet interfaces were excluded from the `bat0` interface so that they're not used by Batman-adv. Please check out the commit comments for the reasons. Actually, the WAN ethernet port always is included in `bat0` as is configured via interface-specific configuration by [lime-hwd-openwrt-wan](https://github.com/libremesh/lime-packages/tree/master/packages/lime-hwd-openwrt-wan) package.

Addresses #56 
This PR reverts dd07aeffcca3aeafbe815ad663f182ff443011e5 and offers an alternative solution to #189 removing the previous solution implemented in ae0672b75b25a5e434a1b4ae6eb5c9bfc3f536a2 (based on the creation of `dummy0` interface, that was working most but not all the times. Due to a race condition in which the `dummy0` interface could be not the first to be included in bat0).
Instead of adding the `dummy0` interface, the MAC addresses of each of the interfaces included in `bat0` is changed randomly (based on the interface name), so that they have all different MAC addresses.

An alternative solution would be to include `br-lan` inside `bat0` instead of `bat0` inside `br-lan`, similarly to #631.


spiccinini:
Hi! Great that you are investigating this. I really don't know how to test it, which things to test exactly.  Maybe we can compile a list of tests? Some ideas:
* Test that there is no loop when connecting two nodes with wifi and LAN.
* Test that DHCP for clients is working as expected (that only one of the nodes is responding?)

ilario:
> Maybe we can compile a list of tests? Some ideas:

Yep, that would be useful!
Some hints are in @altergui comment here dd07aeffcca3aeafbe815ad663f182ff443011e5

* "fragmentation (since mtu is not enough) which impacts performance, in some cases severely (i.e. when the ethernet link is wirelessly bridged)": I suppose that the fragmentation decreasing performances will still be present, maybe could be worth to decrease the MTU of AP interfaces also to 1496? For being honest, I have no idea of these things.
* "instability (racy conditions, it seems) with BLA2": @altergui could you detail this?
* "spurious warnings about 'received packet with own address' on br-lan": I think I solved this randomizing the MAC addresses, do you think this solution could have drawbacks?
* "complicates things when two different clouds are accidentally connected by an plain ethernet cable between them": no idea why, they are on different VLANs anyway...

>     * Test that there is no loop when connecting two nodes with wifi and LAN.

Yess

>     * Test that DHCP for clients is working as expected (that only one of the nodes is responding?)

All nodes connected to br-lan are already answering, this is already broken and this PR does not solve it (but this change should be taken into consideration when we will try to fix that), see #658.

ilario:
Daniel Golle proposed a solution here: https://github.com/libremesh/lime-packages/issues/56#issuecomment-637598835

If we decide to go that way, I can remove the eth mesh-related commit and leave this PR only about the removal of dummy0 interface.

dangowrt:
Testing the current version of the patch (rebased so it would apply) and looks good till now.
* Would be nice to try increasing MTU on batman hardifs to 1532, so full 1500 bytes could be carried. This will most likely not work with all Ethernet switches, but works well with wifi interfaces of any kind (and that's where batman-adv is actually carrying traffic). So we should raise the MTU only for wifi devices and not mess with wired interfaces.
* BLA seems to do the job, traffic goes through the plain bridge rather than batman-adv.
*  no more `received packet with own address on br-lan` or the like

packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
```diff
@@ -74,8 +60,18 @@ function batadv.setup_interface(ifname, args)
...
+	--! Randomize MAC address for each of the interfaces included in Batman-adv.
+	local id = utils.get_id(ifname)
+	local randomMac = network.primary_mac();
+	randomMac[1] = id[1]
+	randomMac[2] = id[2]
+	randomMac[3] = id[3]
+
 	local uci = config.get_uci_cursor()
 	uci:set("network", owrtDeviceName, "mtu", mtu)
+	uci:set("network", owrtDeviceName, "macaddr", table.concat(randomMac, ":"))
```
G10h4ck, line 74: This used to break linux 802.1{ad,q} at least when bridges and vlans operates on same hard interface, if the mac of the vlan is different from the one on the hard interface (what you are doing here) the packets get sucked by the bridge and never get to the vlan interface. I haven't tested this recently but this happened at least until last libremesh stable release.

G10h4ck:
>     * Would be nice to try increasing MTU on batman hardifs to 1532, so full 1500 bytes could be carried. This will most likely not work with all Ethernet switches, but works well with wifi interfaces of any kind (and that's where batman-adv is actually carrying traffic). So we should raise the MTU only for wifi devices and not mess with wired interfaces.

AFAIU we are already increasing the MTU on wireless hard interfaces for the sake of not fragmenting `batman-adv`

https://github.com/libremesh/lime-packages/blob/master/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua#L87

Also it would be really nice to have a way to detect ethernet interface which would support increased MTU (check if they are gigabit?)

dangowrt:
> AFAIU we are already increasing the MTU on wireless hard interfaces for the sake of not fragmenting `batman-adv`
> 
> https://github.com/libremesh/lime-packages/blob/master/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua#L87

nice, i didn't notice.

> Also it would be really nice to have a way to detect ethernet interface which would support increased MTU (check if they are gigabit?)

even that wouldn't not necessarily be true, because some devices got an external switch chip which doesn't support jumbo frames (eg. lantiq devices with external admtek/infineon/lantiq switch chip). i guess this will only be solved properly by using DSA instead of swconfig, as then the call to set the MTU to anything higher than supported would just return -EFAIL.

G10h4ck:
> even that wouldn't not necessarily be true, because some devices got an external switch chip which doesn't support jumbo frames (eg. lantiq devices with external admtek/infineon/lantiq switch chip). i guess this will only be solved properly by using DSA instead of swconfig, as then the call to set the MTU to anything higher than supported would just return -EFAIL.

Does OpenWrt plan to move from swconfig to DSA any time soon?

dangowrt:
> Does OpenWrt plan to move from swconfig to DSA any time soon?

Yes, but for some outdated targets it will most likely never happen.
It has already happened for ramips/mt7621, lantiq/xrx200.
People are working on AR9331, some initial work was done for Ralink-ESW
as found on MT7628. Once this is more complete, we can try supporting
all the older Ralink/MediaTek stuff as well, because the switch IP in
those SoCs is very similar.
For ath79 it will certainly not be ready this year, because there are
too many boards and many got external switch chips (which then also
need DSA drivers, of course).

G10h4ck:
> Does OpenWrt plan to move from swconfig to DSA any time soon?
> Yes, but for some outdated targets it will most likely never happen. It has already happened for ramips/mt7621, lantiq/xrx200. People are working on AR9331, some initial work was done for Ralink-ESW as found on MT7628. Once this is more complete, we can try supporting all the older Ralink/MediaTek stuff as well, because the switch IP in those SoCs is very similar. For ath79 it will certainly not be ready this year, because there are too many boards and many got external switch chips (which then also need DSA drivers, of course).

First of all thanks for the report!

So we could eventually test if DSA is supported (this is as simple as see if some command is in PATH?), if so then test if MTU can be increased, and if one of those two tests fails act as if MTU is 1500?

When we designed LibreRouter I remember to insist on having gigabit ethernet mainly because I thought that would guarantee increased MTU support, do you remember if the chip we used in LR plus it's drivers does supports that well?

Cheers!

packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
```diff
@@ -74,8 +60,18 @@ function batadv.setup_interface(ifname, args)
 
 	local owrtInterfaceName, _, owrtDeviceName = network.createVlanIface(ifname, vlanId, nameSuffix, vlanProto)
 
+	--! Avoid dmesg flooding caused by BLA with messages like "br-lan:
+	--! received packet on bat0 with own address as source address".
+	--! Randomize MAC address for each of the interfaces included in Batman-adv.
+	local id = utils.get_id(ifname)
+	local randomMac = network.primary_mac();
+	randomMac[1] = id[1]
```
G10h4ck, line 68: Some bits of first byte of the mac have special meanings, so it cannot be totally random. Also at first glance I think we can just change the firs two bytes with something fixed making sure multicast bit is 0 and locally administered bit is 1 https://en.wikipedia.org/wiki/MAC_address#/media/File:MAC-48_Address.svg

G10h4ck:
I have reworked this a bit in this PR https://github.com/ilario/lime-packages/pull/1 and started testing it, at first glance it seems I could not reproduce the old problematic behavior which was that when the vlan interface had it's mac address changed, unicast packets never got there (probably sucked in by the bridge), while multicast packets arrived without problem so routing protocols kept discovering the paths but then unicast traffic was lost. Did something big changed in 802.1q or bridge code in the kernel in this years?

I'll keep testing this in this a bit more and if there are no problems we can finally merge this with the extra fixup i made.

Hopefully loops when both WiFi and ethernet link are available will disappear with this.

ilario:
The commits here got messy, sorry for that, continuing on #724

PR #707, opened 2020-05-08
ubus-lime-location improvements

spiccinini:
* Force shared state to update the nodes and links local database to be able to use this information right away in the lime-app so the links in the map can be updated when the position of the node is 
changed.
* Add a convenient method to get the mac addresses with globing support
* Add tests
* Cleanups

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -21,6 +21,27 @@ function network.get_mac(ifname)
 	return utils.split(macaddr, ":")
 end
 
+--! Return a table of macs based on the interface globing filter
+function network.get_own_macs(intarface_filter)
+	if intarface_filter == nil then
+		intarface_filter = '*'
+	end
+
+	local macs = {}
+	local search_path = "/sys/class/net/" .. intarface_filter .. "/address"
+	for address_path in fs.glob(search_path) do
+		mac = io.open(address_path):read("*l")
+		macs[mac] = 1
+	end
+
+	local result = {}
+	for mac, _ in pairs(macs) do
+		table.insert(result, mac)
+	end
+	return result
+end
+
+
 function network.assert_interface_exists(ifname)
 	assert( ifname ~= nil and ifname ~= "",
 	        "network.primary_interface() could not determine ifname!" )
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -21,13 +21,13 @@ function network.get_mac(ifname)
 end
 
 --! Return a table of macs based on the interface globing filter
-function network.get_own_macs(intarface_filter)
-	if intarface_filter == nil then
-		intarface_filter = '*'
+function network.get_own_macs(interface_filter)
+	if interface_filter == nil then
+		interface_filter = '*'
 	end
 
 	local macs = {}
-	local search_path = "/sys/class/net/" .. intarface_filter .. "/address"
+	local search_path = "/sys/class/net/" .. interface_filter .. "/address"
 	for address_path in fs.glob(search_path) do
 		mac = io.open(address_path):read("*l")
 		macs[mac] = 1
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -16,8 +16,7 @@ network.MTU_ETH = 1500
 network.MTU_ETH_WITH_VLAN = network.MTU_ETH - 4
 
 function network.get_mac(ifname)
-	local path = "/sys/class/net/"..ifname.."/address"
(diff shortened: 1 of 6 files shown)


packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
```diff
@@ -32,43 +25,40 @@ end
...
+    uci:set("libremap", "location", "latitude", msg.lon)
+    uci:set("libremap", "location", "longitude", msg.lat)
 
-local function lines(str)
-  -- split a string into lines separated by line endings
-  local t = {}
-  local function helper(line) table.insert(t, line) return "" end
-  helper((str:gsub("(.-)\r?\n", helper)))
-  return t
+    io.popen("shared-state insert nodes_and_links", "w"):write(json.stringify(_nodes_and_links()))
```
nicopace, line 55: This is a side effect, so it shouldn't be bounded to the success of the function.
If the popen fails, it will make the function fail, right?

packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
```diff
@@ -81,19 +71,17 @@ local function nodes_and_links()
...
+  return result
+end
 
+local function nodes_and_links()
+  utils.printJson(_nodes_and_links())
 end
 
 function all_nodes_and_links(msg)
-  print('{"result":'..shell("shared-state get nodes_and_links")..'}')
+  print('{"result":'..utils.unsafe_shell("shared-state get nodes_and_links")..'}')
```
nicopace, line 84: Eventually shared-state could be called a a library instead of as a shell. Not for now of course...

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -16,11 +16,31 @@ network.MTU_ETH = 1500
...
 function network.get_mac(ifname)
-	local path = "/sys/class/net/"..ifname.."/address"
-	local macaddr = assert(fs.readfile(path), "network.get_mac(...) failed reading: "..path):gsub("\n","")
+	local _, macaddr = next(network.get_own_macs(ifname))
 	return utils.split(macaddr, ":")
 end
 
+--! Return a table of macs based on the interface globing filter
+function network.get_own_macs(intarface_filter)
+	if intarface_filter == nil then
```
nicopace, line 25: it seems there is a spelling mistake, intarface/interface (a-e)

packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
spiccinini: Yes, I thought about this a little. We may export an easier/ergonomic shared-state API as now to use it some stuff must be done (locks have to be taken, released, etc).

packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
spiccinini: If popen fails (like calling `io.popen("false")`) execution continues as nothing happened. So It is not an issue.

packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
```diff
@@ -32,43 +25,41 @@ end
...
+  utils.printJson(_get_location(msg))
 end
 
 local function set_location(msg)
-    local result = {}
-    conn:call("uci", "set", {config="libremap", section="location", values={ longitude = msg.lon, latitude = msg.lat } })
-    printJson({ lat = msg.lat, lon = msg.lon });
-end
+    local uci = config.get_uci_cursor()
+    uci:set("libremap", "location", "latitude", msg.lon)
```
germanferrero, line 52: We are switching longitude and latitude here !

packages/ubus-lime-location/tests/test_lime_location.lua
```diff
@@ -0,0 +1,91 @@
...
+
+    it('test set location', function()
+        local f = io.open("/tmp/lime_location_testing", "w")
+        stub(network, "get_own_macs", function () return {"00:11:7f:13:36:16", "02:ce:16:aa:83:52"} end)
+        stub(io, "popen", function () return f end)
+
+        local response  = rpcd_call(ubus_lime_loc, {'call', 'set'}, '{"lat":"1", "lon":"3.14"}')
+        io.popen:revert()
+        assert.is.equal("ok", response.status)
+        assert.is.equal("1", response.lat)
```
germanferrero, line 52: The test could read libremap config or maybe call location get method after set. That would have cached the bug of switching lat and lon

packages/ubus-lime-location/files/usr/libexec/rpcd/lime-location
spiccinini: Fixed!

packages/ubus-lime-location/tests/test_lime_location.lua
spiccinini: Thanks! I improved the tests

Issue #649, opened 2019-11-25
Batman-adv configuration needs to be adapted to upstream changes


ilario:
When compiling LibreMesh on top of the OpenWrt's openwrt-19.07 branch, I had several errors related to Batman-adv interfaces:

```
daemon.notice netifd: Interface 'lm_net_batadv_dummy_if' is setting up now
daemon.notice netifd: Interface 'lm_net_eth0_2_batadv_if' is now down
daemon.notice netifd: Interface 'lm_net_eth0_2_batadv_if' is setting up now
daemon.notice netifd: Interface 'lm_net_wlan0_mesh_batadv_if' is now down
daemon.notice netifd: Interface 'lm_net_wlan0_mesh_batadv_if' is setting up now
daemon.notice netifd: lm_net_wlan1_mesh_batadv_if (22621): Command failed: Invalid argument
daemon.notice netifd: Interface 'lm_net_wlan1_mesh_batadv_if' is now down
daemon.notice netifd: Interface 'lm_net_wlan1_mesh_batadv_if' is setting up now
daemon.notice netifd: lm_net_batadv_dummy_if (22662): Command failed: Invalid argument
daemon.notice netifd: lm_net_eth0_2_batadv_if (22671): Command failed: Invalid argument
```

Asking for help on Batman-adv IRC seems that the configuration has to be updated due to this commit as documented in the same commit:

https://github.com/openwrt-routing/packages/commit/54af5a209e0a0a75b5eb712c0ca8056e66de02c0

Checking on `ubus monitor` the ubus commands causing the error seems to be:

```
<- 6c63e636 #a20df0f7         invoke: {"objid":-1576144649,"method":"notify_proto","data":{"action":0,"ifname":"lm_net_wlan1_mesh_batadv_if","link-up":true,"keep":false,"interface":"lm_net_wlan1_mesh_batadv_if"}}
-> b84bd45b #6c63e636         invoke: {"objid":-1576144649,"method":"notify_proto","data":{"action":0,"ifname":"lm_net_wlan1_mesh_batadv_if","link-up":true,"keep":false,"interface":"lm_net_wlan1_mesh_batadv_if"},"user":"root","group":"root"}
<- b84bd45b #6c63e636         status: {"status":2,"objid":-1576144649}
-> 6c63e636 #a20df0f7         status: {"status":2,"objid":-1576144649}
```

ilario:
On Batman-adv IRC I received this comments:

```
[...]

PR #708, opened 2020-05-09
proto-batadv: adapt to new UCI API

dangowrt:
commit openwrt-routing/packages@54af5a209e0a
("batman-adv: Split batadv proto in meshif and hardif part") changed the
way batman-adv is configured in OpenWrt. Change the lime proto accordingly.
This is needed for lime-proto-batadv to work with OpenWrt 19.07 or newer.

### DIFF
--- a/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
+++ b/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
@@ -14,24 +14,24 @@ function batadv.configure(args)
 	if batadv.configured then return end
 	batadv.configured = true
 
-	if not fs.lstat("/etc/config/batman-adv") then fs.writefile("/etc/config/batman-adv", "") end
-
 	local uci = config.get_uci_cursor()
 
-	uci:set("batman-adv", "bat0", "mesh")
-	uci:set("batman-adv", "bat0", "bridge_loop_avoidance", "1")
-	uci:set("batman-adv", "bat0", "multicast_mode", "0")
+	uci:set("network", "bat0", "interface")
+	uci:set("network", "bat0", "proto", "batadv")
+	uci:set("network", "bat0", "interface")
+	uci:set("network", "bat0", "bridge_loop_avoidance", "1")
+	uci:set("network", "bat0", "multicast_mode", "0")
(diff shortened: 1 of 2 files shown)


ilario:
This is amazing! Thanks!
Fixes #649

aparcar:
lgtm

spiccinini:
Great @dangowrt ! 
Thanks! I will test it and also adapt it to work also with 18.06 configs.

spiccinini:
@dangowrt I didn't know that I could push to your fork, that was not my intention, sorry. Are you ok with that?
I've tested it with 18.06, now I am building a 19.07 image to test.

spiccinini:
> Shouldn't we just make a branch for libremesh running on openwrt-18.06.x? that compatibility code adds codepath to the target in runtime which are never used (because only either of them will be). If branching is not an option, maybe build-time detection is?

In my opinion, at least for this project, it is easier to maintain only one branch. While I agree that the compatibility code adds codepaths and maintenance burden, I don't see big drawbacks for this code that will run once (or sporadically). Having the unittests helps lowering the maintenance.
Maybe if we have a lot of things it may add up, but for now it seems that compatibility between 18.x and 19.x is low, that is why I think that having to maintain two branches is more work.

> Apart from that, the mere presence of `/etc/config/batman-adv` doesn't guarantee that the old API is indeed still used -- that file can be a left-over kept from sysupgrade... Testing for `/lib/netifd/proto/batadv_hardif.sh` would be better, given that we really opt for run-time detection of the OpenWrt version.
You are right about the left-over from a sysupgrade, thanks.

spiccinini:
@dangowrt Do you have an idea of a build-time detection mechanism?

spiccinini:
So I tested it in the 19.07, first without using this branch, to see exactly the problem (and then testing this branch). For my surprice it worked.... Diging I found this commit https://github.com/openwrt-routing/packages/commit/54af5a209e0a0a75b5eb712c0ca8056e66de02c0#diff-fa7687b85fcf63726669395fe5218908 with a migration script from february, so it worked.  But yes this is not enough becase the next time lime-config is run then this migration won't run.

spiccinini:
Tested in openwrt 18 and 19.

PR #712, opened 2020-05-26
Add check-internet package

spiccinini:
Very simple package with a `check-internet` command that uses HTTP to check for internet.

Also use this in `last_internet`.

### DIFF
--- a/packages/check-internet/Makefile
+++ b/packages/check-internet/Makefile
@@ -1,0 +1,37 @@
+#
+# Copyright (C) 2020 Santiago Piccinini <spiccinini@altermundi.net>
+#
+# This is free software, licensed under the GNU Affero General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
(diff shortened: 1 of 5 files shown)


ilario:
And we should really push this package to the OpenWrt's repositories!

ilario:
Seems that OpenWrt is migrating away from wget towards uclient-fetch, see discussion on #720.
Anyway I just realized that the wget dependency was missing.

spiccinini:
> Seems that OpenWrt is migrating away from wget towards uclient-fetch, see discussion on #720.
> Anyway I just realized that the wget dependency was missing.

if I remember correctly wget is part of busybox and it is built by default.

Anyway It is good idea to move to the main openwrt supported http client!

spiccinini:
I've opened https://github.com/libremesh/lime-packages/pull/721

spiccinini:
So I was wrong, `wget` is a symlink to uclient-fetch, so yes the dependency was missing (we may use #721 to correct this)

PR #713, opened 2020-05-29
wireless: add option legacy_rates to allow disabling 11b rates

dangowrt:
802.11b uses 22MHz channels with CCK modulation which causes significant noise on neighbouring 20MHz channels as they are partially overlapping.
Disabling CCK allows for better spectrum utilization as can then use 4 non-overlapping 20MHz channels (1,5,9,13) instead of just 3 (1,6,11) when located within ETSI regulatory domain.

Using `'not wireless.is5Ghz(radioName)` to designate potential 802.11b radios may not cover all cases (dual-band radios, other not yet supported non-5GHz-bands such as 60GHz)... But does the trick for all classic routers.


spiccinini:
We may even change the default to not support legacy 'b' rates ?

dangowrt:
> We may even change the default to not support legacy 'b' rates ?

Yes, that can go into lime-defaults.

PR #714, opened 2020-05-30
[WIP] Upgrade firmware from the lime-app

spiccinini:
This is a work in progress PR to show/discuss.

Changes to safe-upgrade:
- Add verify and board-supported subcommands
- Add some testing
- Add `--preserve-archive` option to upgrade subcommand (like `sysupgrade -f`)

Changes to ubus-lime-admin:
- Add an endpoint to consult if firmware upgrade with confirmation step is supported (only with safe-upgrade)

Some fixes and added utilities / helpers.
It should be easy to review commit by commit.

TODO
- [x] Add firmware verify ubus call
- [x] Add firmware upgrade ubus call
- [ ] Some automatic way to keep ssh keys (but not keep every config)
- [ ] Add firmware confirm ubus call


packages/safe-upgrade/files/usr/sbin/safe-upgrade
```diff
@@ -246,11 +273,20 @@ function bootstrap(args)
...
 
-    -- TODO: validate that the firmware is valid for this board using metadata
+    local fw_metadata = read_fw_metadata(args.firmware)
+    fw_valid = fw_metadata ~= nil and su.is_firmware_valid(fw_metadata)
+    if not fw_valid then
+        print("Invalid firmware!")
+        if args.force then
+            print("Forcing upgrade to continue as requested")
+        else
+            exit(111)
```
germanferrero: should be os.exit?

germanferrero: fixed in later commit.

PR #715, opened 2020-06-02
network: no longer clear WAN proto

dangowrt:
clearing WAN proto kills existing WAN setups using PPPoE or any
other protocol than DHCP. Especially for remote administered gateways
this is pita as everytime I call lime-config, I need to re-edit
/etc/config/network to set WAN proto back to 'pppoe' not to lock
myself out...


G10h4ck:
@dangowrt  should we delegate the clearing to lime-proto-wan or do you use it too?

dangowrt:
> @dangowrt should we delegate the clearing to lime-proto-wan or do you use it too?

I guess I should be using lime-proto-wan, currently I do not on this router as lime-proto-wan also changes the WAN proto (I need pppoe, not dhcp).
Hence clearing in lime-proto-wan is ok as long as lime-proto-wan is not suitable for my use-case anyway.
For me it's kinda unclear why we hard-code wan proto to be 'none' by default in first place.

spiccinini:
Maybe we should add the ability to for a more detailed configuration of lime-proto-wan (for ppoe, etc) ? I always use static or dhcp so I am not sure if we can map it easily, it seems pretty involved.

I would use the lime-config hooks that run as a final step of lime-config at `/etc/hotplug.d/lime-config/`

Also I am not sure why it is cleared to none.

dangowrt:
@spiccinini thanks for the hint of using the lime-config hotplug hook! that's a least a convenient work-around.
And I agree, more detailed handling in lime-proto-wan would be great, in my case I have to set DSL mode to PTM, create a VLAN interface on top of ptm0, then do pppoe for wan, then use dhcpv6 on wan6 to get v6 address and delegated prefix...

G10h4ck:
> @dangowrt should we delegate the clearing to lime-proto-wan or do you use it too?

I remembered, this is not viable, we clear wan so if `lime-proto-wan` is not used the interface is not used for wan. I thing that right now the hotplug hook is the best option to handle your use case ;)

PR #716, opened 2020-06-02
bmx6/7: unconditionally depend on ebtables

dangowrt:
conditional dependencies are deprecated. create hard dependency instead as it seems the only good solution.


dangowrt:
Only saw #673 now, guess it should be merged with that.

PR #721, opened 2020-06-04
check-internet: use uclient-fetch instead of wget

spiccinini:
uclient-fetch is used by opkg and other core utilities.


spiccinini:
Ok I now see  that `wget` is a symlink to `uclient-fetch`, I thought that wget was part of busybox

dangowrt:
doesn't really make any difference.... shared-state would be the place to use uclient-fetch aka. OpenWrt's wget instead of luci-lib-httpclient which pulls tons of dependencies.

ilario (approved):

Issue #666, opened 2019-12-12
Babeld-only network does not work on wireless mesh links


ilario:
I just tested a network with **no Batman-adv but only Babeld**, on top of OpenWrt 18.06 (initially spotted when testing on OpenWrt 19.07-rc2, where the same problem is present): **wireless mesh links do not work**.

When adding all the protocols but `list protocols batadv:%N1` in the `/etc/config/lime-node`, the routing over the wireless links stops working (while the cabled ones are still ok). Needless to say, when both Batman-adv and Babeld are present everything works.
But it works just because the broken routing is "patched" as the packets flow through bat0.

My setup has 4 nodes: 
Commercial AP with internet - - - - LibreMesh wireless client (10.13.40.169) ------ LibreMesh AP + mesh (10.13.104.0) - - - - LibreMesh AP + mesh (10.13.123.8)

the ping to the internet from the third router works (i.e. the cabled link works), but from the fourth does not work (i.e. the wireless link does not work).

The Babeld dump (`echo dump | nc ::1 30003`) looks good both in the third and in the fourth router, and does not change when adding or removing Batman-adv.

Third router routes:
```
# ip route show
default via 10.13.40.169 dev eth0-1_17 proto babel onlink 
10.13.0.0/16 dev br-lan proto kernel scope link src 10.13.104.0 
10.13.0.0/16 dev anygw proto kernel scope link src 10.13.0.1 
```

ping to the internet works (via eth0-1_17) and ping to the second router (10.13.40.169) works (as it is reachable from the LAN network included in br-lan).

Fourth router routes:
```
# ip route show
default via 10.13.104.0 dev wlan1-mesh_17 proto babel onlink 
10.13.0.0/16 dev br-lan proto kernel scope link src 10.13.123.8 
10.13.0.0/16 dev anygw proto kernel scope link src 10.13.0.1
```

Ping to the internet is sent through wlan1-mesh_17 but the reply does not arrive as the third router sends it through br-lan, which does not include the mesh interface.
Ping to the third router (10.13.104.0) is sent through br-lan and does not even get to the third router as the mesh interface is not included in br-lan.

When Batman-adv is present, the connection magically works, even if the routes are identical to the ones I have shown here, but works in a stupid way I think:
the reply of the ping from the fourth router to the internet **finds its way back as bat0 is included in br-lan**; the ping to the third router works even if it's sent through br-lan as it includes bat0 which includes wlan1-mesh_29.

[...]

PR #724, opened 2020-06-14
Testing Batman-adv on ethernet, replaces #703

ilario:
#703 got messy (sorry for that) and this has the same content, including the rework by @G10h4ck.
See discussion on #703 and #56

### DIFF
--- a/packages/lime-proto-batadv/Makefile
+++ b/packages/lime-proto-batadv/Makefile
@@ -21,7 +21,7 @@ define Package/$(PKG_NAME)
   CATEGORY:=LiMe
   MAINTAINER:=Gioacchino Mazzurco <gio@eigenlab.org>
   URL:=http://libremesh.org
-  DEPENDS:=+lime-system +lua +libuci-lua +kmod-batman-adv +kmod-dummy
+  DEPENDS:=+lime-system +lua +libuci-lua +kmod-batman-adv
   PKGARCH:=all
 endef
 
--- a/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
+++ b/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
@@ -58,13 +58,13 @@ end
 function batadv.setup_interface(ifname, args)
 	if not args["specific"] then
 		if ifname:match("^wlan%d+.ap") then return end
-		if ifname:match("^eth") then return end
 	end
 
 	local vlanId = args[2] or "%N1"
 	local vlanProto = args[3] or "8021ad"
 	local nameSuffix = args[4] or "_batadv"
 	local mtu = 1532
+	if ifname:match("^eth") then mtu = 1496 end
 
 	--! Unless a specific integer is passed, parse network_id (%N1) template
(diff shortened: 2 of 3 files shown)


G10h4ck:
opened another one with conflict fixed https://github.com/libremesh/lime-packages/pull/726

markbirss:
Oh this good news, adding back support for Batman-adv over ethernet 

I have been busy with this also just recently without LibreMesh with devices without radio's
e.g

Cisco Meraki MX60
Cisco ON100
Samsung CY-SWR1100 (the onboard radio's never worked with OpenWRT)

but will test also when I get some time

Im also waiting for some 10Gbit Fiber Networks cards that are supported under OpenWRT for X86 builds

markbirss:
Ok, i tested snapshot with LibreMesh changed routing_also to BATMAN_V (to see reported Ethernet throughput) on my 

Cisco ON100 - does not work correctly, only has eth0 and eth1 ethernet

Cisco MX60  seems ok ( yes - 1000 )
Samsung CY-SWR1100 seems ok  ( no - 1.0 ) 

also

I tried separate vlan on each port to have as many as 3 bat interface
to have a switch for batman-adv over Ethernet

batman bonding is not working as speed reduces and then reaches 0 (with batctl tp)

markbirss:
Hi All

How do I confirm if routing is actually over batman-adv ?

ilario:
Gosh... I usually check the output of `ip route` and `batctl o`

markbirss:
@ilario 

im not sure my current test setup configuration or build is actually correct?

Could you further detail a test configuration for ethernet only ?

I have disabled wifi and removed wifi related interfaces totally

i have firmware snapshot openwrt with libremesh snapshot feeds

I have hardware

1 x TP-Link TL-WR1043NDv1 
1 x Cisco Meraki MX60
1 x Samsung CY-SWR1100

I have connected them together either via LAN ports or from LAN to WAN ports and have internet access on the Cisco Meraki MX60 via WAN port

if I for example rmmod batman-adv i can still ping to internet?

ilario:
I think we should move this conversation away from this pull request (in favour of the lime-users mailing list or of the chat room, see https://libremesh.org/communication.html ) and only report here the conclusion.

Anyway I never tested the ethernet only option, it's interesting.
In order to disable the mesh via wifi, simply add a couple of lines in `/etc/config/lime-node`:

```
config lime system

config lime network

config lime wifi
	list modes 'ap'
	list modes 'apname'
```

this will override the default list of wifi modes (which includes `ieee80211s` mesh).

Never never never connect two LibreMesh nodes LAN to WAN, always LAN to LAN.

Do not yet use LibreMesh on top of OpenWrt snapshot (we never tested with the latest OpenWrt changes), on top of OpenWrt 19.07 should work (I would go this way), on top of OpenWrt 18.06 surely works.

If you remove batman-adv module I suppose that in some cases things will work (e.g. networks with < 2 nodes) but normally it should not work (see #666) as nobody implemented the possibility to use LibreMesh without Batman-adv, I believe.

markbirss:
@ilario 
understood

PR #727, opened 2020-06-21
bmx7: allow setting trustedNodesDir

dangowrt:
To make use of bmx7 signature checking one has to set
`bmx7.general.trustedNodesDir='/etc/bmx7/trustedNodes'`
Allow communities to do that by reading the value from lime's
`network.bmx7_trustedNodesDir` option.

### DIFF
--- a/packages/lime-docs/files/lime-example
+++ b/packages/lime-docs/files/lime-example
@@ -55,6 +55,7 @@ config lime network
 	option bmx7_over_batman false
 	option bmx7_pref_gw none			# Force bmx7 to use a specific gateway to Internet (hostname must be used as identifier)
 	option bmx7_wifi_rate_max 'auto'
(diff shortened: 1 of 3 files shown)


dangowrt:
> All lime config options must have a default value (in lime-defaults file).
Ah, that's odd for values where 'undefined' is a meaningful default (which is the case for 'trustedNodesDir'). I also thought that maybe a bool in lime config to either set bmx7.general.trustedNodesDir='/etc/bmx7/trustedNodesDir' or not set it at all (because having it at that path covers all use-cases on OpenWrt and all we need to decide is whether to enable strict signature checking or not.
> Also please add this to lime-example with a comment documenting it.
I'll change it to a bool `bmx7_enable_pki` then which is '0' by default.

> 
> I am in my mobile phone so I can't check but I remember we use named with all lowercase separated with underscores.
I'm trying to just not introduce another spelling for a variable which already exists (in bmx7 and bmx7 uci config) -- using the boolean above solves that one as well :)

PR #728, opened 2020-06-28
CI: build all packages via GitHub action

aparcar:
Signed-off-by: Paul Spooren <mail@aparcar.org>

PR #729, opened 2020-06-29
Use libremesh.mk

aparcar:
Instead of having a complex Makefile per package this commit introduces
    a Makefile template called `libremesh.mk` which is included in all
    compatible LibreMesh packages.
    
    This follow the approach of LuCI where only very basic information is
    required per package and the rest is evaluated automatically.
    
    The following variables can be set:
    
    * LIME_TITLE
    * LIME_MAINTAINER
    * LIME_DESCRIPTION
    * LIME_DEPENDS
    
    Everything else is automatically determined.


ilario:
It sounds very good to me. I hate writing Makefiles and to have a unified template seems very convenient.

Some of the packages could need special stuff (e.g. @aparcar mentioned that [tinc package Makefile](https://github.com/libremesh/lime-packages/blob/master/packages/tinc/Makefile) cannot be simplified this way).
The packages and maintainers of the packages involved in this PR are (maintainers, could you comment if this change is ok?):

@altergui 

* auto-usb-wwan

@p4u 

* bmx7-auto-gw-mode

@G10h4ck 

* lime-proto-babeld
* lime-proto-batadv
* lime-proto-bmx6
* lime-proto-bmx7
* lime-proto-wan
* shared-state-babeld_hosts
* shared-state-bat_hosts
* shared-state

@aparcar 

* lime-smart-wifi

aparcar:
This can be seen as a drop in replacement. All other packages with regular Makefiles keep working. Special cases like tinc, where actual C compiling is required, keep their individual Makefiles.

spiccinini:
It is simpler/cleaner. 
But for new developers IMHO it is not easier as it is different than "normal" openwrt packages, we are adding a layer of indirection (and in makefiles that are not very friendly). So with things like this I tend to think that maybe we can improve upstream somehow. But well it is only a personal opinon i am not -1.
Also not all packages will be converted right? As they would be harder to upstream. So we will have two ways of doing packages, the "simplified lime way" and the normal package way.

I would prefer something that use composition instead of inheritance, something like:

```
include $(INCLUDE_DIR)/lime_package.mk

define Build/Compile
	lime_default_compile
endef
```

aparcar:
I don't see how this makes anything more difficult. Whoever understands the regular Makefile for packages likely also saw a LuCI Makefile, which is pretty much the same as the LiMe Makefile. Also, as it is still possible to use regular Makefile, existing developers can show the new developer how to convert the Makefile.

I did not count but I think the `libremesh.mk` covers 85% of the `lime-packages`, I don't see the point using the same tweaks in multiple Makefiles instead of defining it one.

Regarding upstream, I never saw any ambitions to get this done. Even if so, we can also migrate the `libremesh.mk` file as it's done for e.g. Python, they ship their own file, too[1].

[1]: https://github.com/openwrt/packages/tree/master/lang/python

libremesh.mk
```diff
@@ -0,0 +1,63 @@
...
+
+PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+  SECTION:=lime
+  CATEGORY:=LibreMesh
+  TITLE:=$(LIME_TITLE)
+  DEPENDS:=+lime-system $(LIME_DEPENDS)
```
G10h4ck, line 35: Not every package depends on `lime-system` as an example `shared-state` doesn't, but with this change it would install lime-system too when it is not needed

G10h4ck (dismissed):
The underlining idea is good, but it seems there are also a bunch of unrelated changes mixed in into this PR, that makes review error prone, please split this PR in smaller pieces taking in account also other comments

libremesh.mk
aparcar: Good point, what would you suggest? Should everyone just add the package in the *LiMe* Makefile?

Not relying on `lime-system` sounds like upstream material to me. What about adding those packages to the official package repository?

ilario: I agree that there's plenty of stuff which should be upstreamed (bandwidth-test, batman-adv-auto-gw-mode, bmx6-auto-gw-mode, bmx7-auto-gw-bw-mode, bmx7-auto-gw-mode, bmx7-mdns, check-date-http, hotplug-initd-services, lua-argparse, luci-app-batman-adv, luci-app-openairview, luci-mod-status, pirania-app, pirania, shared-state-babeld_hosts, shared-state-bat_hosts, shared-state-dnsmasq_hosts, shared-state-dnsmasq_leases, shared-state, shared-state-persist, shared-state-pirania, soporteremoto, tinc) but in the meanwhile I suggest to remove lime-system from the libremesh.mk profile and to add it to the single Makefiles.

ilario:
Some duplicate files accidentally appeared in the pull request, @aparcar could you delete them?

packages/lime-proto-bmx6/src/etc/uci-defaults/85-add-bmx6-addresses-to-hosts
packages/lime-proto-bmx6/src/usr/bin/bmx6hosts
packages/lime-proto-bmx6/src/usr/bin/bmx6hosts_dnsmasq_update
 packages/lime-proto-bmx6/src/usr/lib/lua/lime/proto/bmx6.lua

ilario:
And also: this PR includes the `.github/workflows/build.yml` file which better fits #728.

aparcar:
> And also: this PR includes the .github/workflows/build.yml file which better fits #728. @aparcar Can you remove it from here?

@ilario done

ilario:
@aparcar have you seen my other comment on the accidental creation of duplicated files https://github.com/libremesh/lime-packages/pull/729#issuecomment-683289206 ?

libremesh.mk
ilario: Sorry, I changed my mind: it's better to leave the lime-system dependency inside libremesh.mk and use libremesh.mk only for LibreMesh specific packages.
In this PR the following Makefiles are being edited:

auto-usb-wwan - NOT LibreMesh specific
bmx7-auto-gw-mode - NOT LibreMesh specific
lime-proto-babeld - LibreMesh specific
lime-proto-batadv - LibreMesh specific
lime-proto-bmx6 - LibreMesh specific
lime-proto-bmx7 - LibreMesh scecific
lime-proto-wan - LibreMesh specific
lime-smart-wifi - LibreMesh specific

@aparcar could you remove auto-usb-wwan and bmx7-auto-gw-mode Makefiles from this PR?

aparcar: Removed

aparcar:
> @aparcar have you seen my other comment on the accidental creation of duplicated files #729 (comment) ?

Removed.

packages/lime-proto-bmx6/Makefile
```diff
@@ -1,46 +1,7 @@
...
-	$(INSTALL_BIN) ./build/bmx6hosts $(1)/usr/bin/
-	$(INSTALL_BIN) ./build/85-add-bmx6-addresses-to-hosts $(1)/etc/uci-defaults/
-	$(INSTALL_BIN) ./build/bmx6hosts_dnsmasq_update $(1)/usr/bin/bmx6hosts_dnsmasq_update
-endef
-
-$(eval $(call BuildPackage,$(PKG_NAME)))
+LIME_TITLE:=LiMe Bmx6 proto support
+LIME_MAINTAINER:=Gioacchino Mazzurco <gio@eigenlab.org>
+LIME_DEPENDS:=+bmx6 +bmx6-json +bmx6-sms +bmx6-table +bmx6-uci-config \
+	+!PACKAGE_firewall:iptables \ +lime-system +lua +libuci-lua \
```
ilario, line 4: there's a backslash in the middle of the line, can you remove it?

packages/lime-proto-bmx6/src/usr/bin/bmx6hosts_dnsmasq_update
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
ilario, line 1: this file is going to be a duplicate, can you remove this?

ilario (changes requested):
An additional comment: you merged the files in the lime-smart-wifi packages into a srv/ directory instead than into a files/ directory.

ilario:
While we wait for @aparcar to change the location of the files inside lime-smart-wifi, can some other dev say if you're ok with this new style of Makefile for LibreMesh specific packages (up to now we had a positive opinion from @G10h4ck and a neutral opinion from @spiccinini)? @nicopace @gmarcos87 @germanferrero

ilario (approved):

PR #736, opened 2020-07-27
Add generic configuration support

### LINKED ISSUE #719: [RFC] Node and community assets and generic configs
After spliting the settings into lime-node and lime-community there are still some loose ends. This proposal tries to tie them up :)
What do you think?

## Guidelines

1. It should be easy to setup (and reason about) community-wise and node only configurations for any software in the firmware, not only the configurations already supported in LibreMesh.
2. It should be easy to upgrade the firmware by keeping the community/node configurations.
3. Configuration changes that are backwards incompatible should be doable without manual intervention and without loosing the current configuration.

## Proposal

Currently there are package configurations that cannot be specified without being for the community or for the node only.
The same applies to uci-defaults and hotplug hooks.
The idea is to be able to include all these use cases under the lime-node, lime-community paradigm.

* For package configurations that do not yet have their integration (option in lime config + hook) we add the type of section in the lime config called `generic_uci_config` that specifies generic configurations of type `uci set` to modify those of another package.
* For the `uci-defaults` and `hotplug.d/lime-config` hooks, we added the directories, `node-assets` and `community-assets`. And the following section types:
    - `copy_asset` copy an asset into a specified path.
    - `run_asset_on_firstboot` executes at first boot (using a uci-default for that end, eg: `uci-defaults/90_lime_generic`).
    - `run_asset` executes an asset at each lime-config run

This way not longer need to maintain other configuration files during an upgrade that are not exactly those specified by the lime-node/lime-community and the corresponding assets.

It is important to clarify then (in the documentation) that these *unforseen* configurations are a trigger to development and eventually have to end up as internal libre-mesh integrations.
And so hacker communities don't have to be maintaining these extra configurations that require maintenance.

Pros:
- Allows for flexibility in node configuration and easy firmware upgrade by maintaining that configuration.
- Allows to make an adhoc configuration through the generic_uci_config, and at the same time not to get lost when someone else comes and runs lime-config.
- If a firmware does not have custom packages you can just use a vanilla libremesh firmware and just add the lime-node and lime-community files.

Example:
(based on the quintanalibre network-profile https://github.com/libremesh/network-profiles/tree/master/quintanalibre.org.ar/generic/files/etc)

```bash
config generic_uci_config libremap
    list uci_set "libremap.settings.community=quintana.libre.org.ar"
    list uci_set "libremap.settings.community_lat=-31.80461"
    list uci_set "libremap.settings.community_lon=-64.41609"

config copy_asset collectd
    option asset 'collectd.conf'
    option path '/etc/collectd.conf'

config run_asset_on_firstboot cron-workaround-tasks
    option asset 'cron-workaround-tasks'

config run_asset dropbear
    option asset 'dropbear'
    option run_once '0' # once per upgrade
```

Random notes:
* Config asset directories may be something like `/etc/lime-assets/{node, community}`
* We may use the *files convention* (`/etc/lime-assets/community/files/usr/bin/foo`) instead of `copy_asset`. I have mixed feelings, prefer the copy_asset because of it being explicit.
* We should integrate more configs into lime, like the libremap/location, etc.
* This proposal allows to to not keep any configuration beside lime-node / lime-community and its assets. So the recommended way of upgrading will be to use `lime-sysupgrade`  (that internally uses `sysupgrade -f` with the `lime.system.keep_on_upgrade` config )

spiccinini:
This fixes #719

###  generic_uci_config

Specifies generic configurations of type uci set to modify the configs of another package.

Example:
```bash
 config generic_uci_config libremap
    list uci_set "libremap.settings=libremap"
    list uci_set "libremap.settings.community=our.libre.org"
    list uci_set "libremap.settings.community_lat=-200.123"
    list uci_set "libremap.settings.community_lon=500.9"
```

### copy_asset
Copy a file from the assets directory into a specified path `dst`.
```bash
config copy_asset collectd
   option asset 'collectd.conf'
   option dst '/etc/collectd.conf'
```

### run_asset 

Executes a file from the assets directory. The execution may be at first boot (ATFIRSTBOOT) or when lime-config is run (ATCONFIG ~~RECONFIG~~, I don't like the much the RECONFIG name). This configuration provides the functionality of uci-defaults and limeconfig hooks. 
Note: It does not rely in uci-defaults for the first boot executions so it is not exactly the same as uci-defaults as the scripts will run when lime-config is run but the implementation is simpler and I didn't find an example of something that must be  be modified prior to run lime-config. 
Also each script will be run only once (it does not depend on the exit code of each script).

Example:
```bash
config run_asset dropbear
     option asset 'dropbear.sh'
     option when 'ATFIRSTBOOT' # ATFIRSTBOOT, ATCONFIG
```

## Asset directories

This implementation adds two directories `/etc/lime-assets/node` and `/etc/lime-assets/community`.

There are two ways of specifying a different config than the community config for a node:
1. Specifying the asset in the node config, for example in `/etc/config/lime-node`:
```bash
config run_asset dropbear
     option asset 'community/dropbear_mynode.sh'
     option when 'ATFIRSTBOOT'
```
2. Specifying the new asset in the `/etc/lime-assets/node` directory and config at lime-node:
```
config run_asset dropbear
     option asset 'node/dropbear_mynode.sh'
     option when 'ATFIRSTBOOT'
```

What do you think?

I am still working in the documentation (will open a PR in lime-web)

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
@@ -13,9 +13,10 @@ function gen_cfg.clean()
 end
 
 function gen_cfg.configure()
-    gen_cfg.do_generic_uci_config()
-    gen_cfg.do_copy_asset()
+    gen_cfg.do_generic_uci_configs()
+    gen_cfg.do_copy_assets()
 end
+
 
 --! Generic UCI configuration from libremesh. Eg usage:
 --!   config generic_uci_config libremap
@@ -23,7 +24,7 @@ end
 --!     list uci_set "libremap.settings.community=our.libre.org"
 --!     list uci_set "libremap.settings.community_lat=-200.123"
 --!     list uci_set "libremap.settings.community_lon=500.9"
-function gen_cfg.do_generic_uci_config()
+function gen_cfg.do_generic_uci_configs()
     local uci = config.get_uci_cursor()
     local ok = true
     utils.log("Applying generic configs:")
@@ -41,7 +42,15 @@ function gen_cfg.do_generic_uci_config()
     return ok
 end
 
-function gen_cfg.do_copy_asset()
+--! copy_asset copy an file from the assets directory into a specified path.
+--! The node asset directory (/etc/lime-assets/node) has priority over the community directory when
+--! both assets exists. Use this to specify a per node configuration.
+--!
+--! config copy_asset collectd
+--!    option asset 'collectd.conf' # relative to /etc/lime-assets/node or /etc/lime-assets/community/
+--!    option path '/etc/collectd.conf'
+--!
+function gen_cfg.do_copy_assets()
     local uci = config.get_uci_cursor()
     local ok = true
     utils.log("Copying assets:")
--- a/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
@@ -8,7 +8,7 @@ gen_cfg = {}
 gen_cfg.NODE_ASSET_DIR = '/etc/lime-assets/node/'
 gen_cfg.COMMUNITY_ASSET_DIR = '/etc/lime-assets/community/'
 gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE = '/etc/.cfg_first_boot_already_run'
-gen_cfg.RUN_ASSET_FIRSTBOOT = 'FIRSTBOOT'
+gen_cfg.RUN_ASSET_AT_FIRSTBOOT = 'ATFIRSTBOOT'
 gen_cfg.RUN_ASSET_AT_CONFIG = 'ATCONFIG'
 
 function gen_cfg.clean()
@@ -19,7 +19,7 @@ function gen_cfg.configure()
     gen_cfg.do_generic_uci_configs()
     gen_cfg.do_copy_assets()
     if not utils.file_exists(gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE) then
-        gen_cfg.do_run_assets(gen_cfg.RUN_ASSET_FIRSTBOOT)
+        gen_cfg.do_run_assets(gen_cfg.RUN_ASSET_AT_FIRSTBOOT)
         utils.write_file(gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE, '')
     end
     gen_cfg.do_run_assets(gen_cfg.RUN_ASSET_AT_CONFIG)
@@ -100,7 +100,7 @@ end
 --! Executes a file from the assets directory
 --! config run_asset dropbear
 --!     option asset 'dropbear.sh'
---!     option when 'FIRSTBOOT' # FIRSTBOOT, ATCONFIG
+--!     option when 'ATFIRSTBOOT' # ATFIRSTBOOT, ATCONFIG
 --!
 function gen_cfg.do_run_assets(when)
     local uci = config.get_uci_cursor()
--- a/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
@@ -8,6 +8,8 @@ gen_cfg = {}
 gen_cfg.NODE_ASSET_DIR = '/etc/lime-assets/node/'
 gen_cfg.COMMUNITY_ASSET_DIR = '/etc/lime-assets/community/'
 gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE = '/etc/.cfg_first_boot_already_run'
+gen_cfg.RUN_ASSET_FIRSTBOOT = 'FIRSTBOOT'
+gen_cfg.RUN_ASSET_AT_CONFIG = 'ATCONFIG'
 
 function gen_cfg.clean()
     -- nothing to clean, but needs to be declared to comply with the API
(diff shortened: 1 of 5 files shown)


packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
```diff
@@ -0,0 +1,43 @@
...
+
+local config = require("lime.config")
+local utils = require("lime.utils")
+
+gen_cfg = {}
+
+gen_cfg.NODE_ASSET_DIR = '/etc/lime-assets/node/'
+gen_cfg.COMMUNITY_ASSET_DIR = '/etc/lime-assets/community/'
+
+function gen_cfg.clean()
```
nicopace, line 12: empty function. please remove.

packages/lime-system/tests/test_lime_generic_config.lua
```diff
@@ -0,0 +1,37 @@
+local config = require 'lime.config'
+local gen_cfg = require 'lime.generic_config'
+local test_utils = require 'tests.utils'
+
+local uci = nil
+
+describe('LiMe Generic config tests #genericconfig', function()
+
+    it('test config.do_generic_uci_config', function()
```
nicopace: it would be good to test when a config fails, as it is an important usecase (the first alternate path).

packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
spiccinini: ` clean` and `configure` are mandatory for all lime modules, maybe I can improve the comments.

nicopace: thanks!

packages/lime-system/tests/test_lime_generic_config.lua
```diff
@@ -0,0 +1,81 @@
...
+        ]]
+
+        test_utils.write_uci_file(uci, config.UCI_CONFIG_NAME, content)
+        assert.is_true(gen_cfg.do_generic_uci_configs())
+
+        -- check that everything is commited
+        uci:load('libremap')
+        assert.is.equal('our.libre.org', uci:get("libremap.settings.community"))
+        assert.is.equal('-200.123', uci:get("libremap.settings.community_lat"))
+        assert.is.equal('-200.123', uci:get("libremap.settings.community_lat"))
```
germanferrero, line 26: community_lat is tested twice instead of community_lon

packages/lime-system/tests/test_lime_generic_config.lua
spiccinini: Thanks, I am also adding more tests for other paths that werent exercised

packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
```diff
@@ -0,0 +1,129 @@
...
+    -- nothing to clean, but needs to be declared to comply with the API
+end
+
+function gen_cfg.configure()
+    gen_cfg.do_generic_uci_configs()
+    gen_cfg.do_copy_assets()
+    if not utils.file_exists(gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE) then
+        gen_cfg.do_run_assets('FIRSTBOOT')
```
G10h4ck, line 20: Should we use some enum like variable for `FIRSTBOOT` and `RECONFIG` ?

spiccinini: done

spiccinini:
I've renamed RECONFIG to ATCONFIG. Although it still does not seem to me a good name it is better than the previous one. What do you think?

G10h4ck:
Are the `.placeholder` files just to keep the dorectory in git? Isn't it easier to put an empty `.gitignore`? Anyway it is ready to merge IMHO!

spiccinini:
@G10h4ck what do you think about the asset directories? I am still not very happy of having two ways of doing things. I think that it is easier to maintain (from the perspective of the network maintainer) to have only one asset directory, and not the community and node directories, and use different file names. So the only file that is different per node is the lime-node file, and every node has all the other files. Obviously this can be done having both directories and puting all the files in the community directory but it is less intuitive when to do what... and of course there are use cases of not having all the asset files in all the nodes.

germanferrero:
wrt having two ways of giving priority to a node asset over an equal named community asset.
I would choose only the option of being explicit about the usage of the asset in lime-node file, and be forced to put the new asset at the node's asset directory. 
In this way, to check if the node has any special configuration you only need to check the lime-node file. No need to check into any asset directory nor the lime-community file.

packages/lime-system/files/usr/lib/lua/lime/generic_config.lua
```diff
@@ -17,10 +19,10 @@ function gen_cfg.configure()
     gen_cfg.do_generic_uci_configs()
     gen_cfg.do_copy_assets()
     if not utils.file_exists(gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE) then
-        gen_cfg.do_run_assets('FIRSTBOOT')
+        gen_cfg.do_run_assets(gen_cfg.RUN_ASSET_FIRSTBOOT)
         utils.write_file(gen_cfg.CONFIG_FIRST_BOOT_SIGNAL_FILE, '')
     end
-    gen_cfg.do_run_assets('RECONFIG')
+    gen_cfg.do_run_assets(gen_cfg.RUN_ASSET_AT_CONFIG)
```
nicopace, line 25: RUN_ASSET_AT_CONFIG uses AT, and RUN_ASSET_FIRSTBOOT not, please fix using AT in both or in none.

spiccinini: Thanks! Fixed

spiccinini:
I did a change that I believe address @germanferrero and my concerns about the unexplicitness of the hierarchy. I reflected that in the description of the PR.

PR #737, opened 2020-07-28
safe-upgrade improvements

spiccinini:
Many improvements in safe-upgrade! This changes allows to expose better a firmware upgrade mechanism through ubus.

* upgrade now reboots  after upgrade (add `--no-reboot`), just like sysupgrade.  Rebooting right away after upgrading is almost always what the user want to do.
* improved upgrade backup functionality adding `--preserve-archive` option to the upgrade (like `sysupgrade -f`). By default uses config 'lime.keep_on_upgrade' to provide the list of files to preserve.
* remove short option `-f` (as this means something else in sysupgrade it may lead to confusion)
* add `verify` cmd to veryfy of a firmware (like `sysupgrade --test`)
* add `board_supported` subcommand (to be able to ask the firmware if safe-uppgrade can be performed or a fallback to sysupgrade has to be used)
* testing, cleanups and refactoring to be able to add unittesting

Also other improvements:
*  tests: add `setup_test_dir()` and `teardown_test_dir()` helpers. Use this to get a working directory to create files for tests. The teardown will remove the directory.
* utils: refactor literalize to return only one argument. The idea of literalize was always to only return the new string
but by mistake it returned two arguments. So tests passed and also some (all?) current usages were working as expected.

### DIFF
--- a/packages/safe-upgrade/files/usr/sbin/safe-upgrade
+++ b/packages/safe-upgrade/files/usr/sbin/safe-upgrade
@@ -28,12 +28,13 @@ local firmware_size_bytes = 7936*1024
 local fw1_addr = 0x9f050000
 local fw2_addr = 0x9f050000 + firmware_size_bytes
 su._supported_devices = {'librerouter-v1'}
+su._mtd_partitions_desc = '/proc/mtd'
 
 -- safe upgrade script, generated with bootscript.py, DO NOT edit here!
 local bootcmd = 'run preboot; boot_part=${stable_part}; if test ${testing_part} -ne 0; then echo Testing part ${testing_part}; boot_part=${testing_part}; set testing_part 0; saveenv; fi; if test ${boot_part} -eq 2; then fw_addr=${fw2_addr}; run boot_2; else fw_addr=${fw1_addr}; run boot_1; fi; run boot_1; bootm ${fw1_addr};'
 
-local STABLE_PARTITION_NAME = 'stable_part'
-local TESTING_PARTITION_NAME = 'testing_part'
+su.STABLE_PARTITION_NAME = 'stable_part'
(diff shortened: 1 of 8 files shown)


packages/safe-upgrade/files/usr/sbin/safe-upgrade
```diff
@@ -336,6 +336,11 @@ local function upgrade(args)
     local boot_script = string.format(boot_script_tpl, fw_mtd_str, partitions.other)
     set_uboot_env(string.format('boot_%d', partitions.other), boot_script)
     set_testing_partition(partitions.other)
+
```
nicopace, line 338: For a future improvement, the upgrade function relies on args, and which args it uses are not explicit. would be better to make them explicit.

spiccinini: Make them explicit how? Document it? Do you have an idea? Thanks!

packages/safe-upgrade/files/usr/sbin/safe-upgrade
```diff
@@ -359,6 +382,13 @@ else
         confirm(args)
     elseif args['test-other-partition'] then
         test_other_partition(args)
+    elseif args['board-supported'] then
+        if su.is_current_board_supported() then
+            os.exit(0)
+        else
+            print("This board is not supported")
+            os.exit(198)
```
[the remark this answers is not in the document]
spiccinini: Thanks, already pushed the patch

packages/safe-upgrade/files/usr/sbin/safe-upgrade
spiccinini: Ok I understand now, thanks! I will make the arguments explicit in a future PR

PR #738, opened 2020-07-28
Firmware upgrade through ubus

spiccinini:
This PR adds the ability to upgrade libremesh using ubus calls using the password authenticated `lime-utils-admin` rpcd.
It is built on #737 so it should be merged after it.

I am opening this as WIP/RFC in to receive some feedback.

# API proposed

## File upload

Using `cgi-io` like luci, so no new code but added cgi-io as a dependency.

## `is_upgrade_confirm_supported()`

Returns true if safe-upgrade is supported by the board (only LibreRouter for now)

## `firmware_verify(fw_path)`

Verifyies a firmware  using `sysupgrade --test` or  `safe-upgrade verify`

## `firmware_upgrade(fw_path, preserve_config, metadata)`

Uses `lime-sysupgrade` to perform the upgrade.
If preserve_config is true then the files configured at `lime.system.keep_on_upgrade` will be preserved.
Metadata provided will be stored in the new firmware. This can be used for example to provide the installation timestamp to the new firmware, set an ID, or store other useful information related to the upgrade process.

## `last_upgrade_metadata()`

Return the metadata stored by the last upgrade.

## `firmware_upgrade_confirm()`

Confirm a firmware flashed with `safe-upgrade`

## `safe_upgrade_confirm_remaining_s()`
Return the seconds that remain until safe-upgrade performs the reboot.

# This PR also does
- lime-sysupgrade: 
   * add safe-upgrade support
   * dehardcode backup archive
- utils: 
  * add the function `execute_daemonized(cmd, out_file)`: Run a command in a shell and daemonized directly in pid 1 (similar to diswon/nohup)
- lime.config: remove base-files-essential from the default keep_on_upgrade list. `base-files-essential` is not really needed and may cause problems when upgrading for multiple reasons.  The main reason is that openwrt config is not always forward compatible (some files have to be generated with new names, like /etc/config/wireless).

TODO:

- 

- [x]  Add the firmware_confirm step
- [x] Handle some files that would be good to preserve (ssh server key, vpn keys, and other locally generated)


packages/ubus-lime-utils/tests/test_lime_utils_admin.lua
```diff
@@ -55,33 +63,74 @@ describe('ubus-lime-utils-admin tests #ubuslimeutilsadmin', function()
         local response  = rpcd_call(ubus_lime_utils, {'call', 'firmware_verify'},
                                     '{"fw_path": "/foo"}')
         assert.is.equal("ok", response.status)
-        os.execute:revert()
-        utils.file_exists:revert()
     end)
 
-    it('test firmware_upgrade', function()
+    it('test firmware_upgrade without new metadata', function()
```
germanferrero, line 68: @spiccinini these tests cover the metadata functionality partially, which is understandable given the effort needed to test it completely.
It could be nice to test that the metadata will be persisted. I think the best way to test it is with an integration test, but needs to mock fs and 'reboot' somehow.

packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils-admin
```diff
@@ -62,49 +62,66 @@ local function firmware_upgrade(msg)
         return utils.printJson({status = 'error',
                                 message = "Firmware file not found"})
     end
+    local metadata = msg.metadata or {}
+    local backup = ""
+    if not msg.preserve_config then
+        backup = "DO_NOT_BACKUP=1"
```
nicopace, line 68: isn't this an implicit message passing approach?
wouldn't it better to be added as a flag in lime-sysupgrade?

spiccinini: It is not implicit in my opinion, but certainly it is not nice to have to do it using env variables. The thing is that lime-sysupgrade is a bash script that has to be reworked a lot to be able to receive parameters.

nicopace: ok... maybe for future work then... thanks!

packages/pirania/files/usr/lib/lua/voucher/utils.lua
```diff
@@ -23,7 +23,7 @@ local function escapeCSV (s)
 end
 
 local function dateNow ()
-  return os.time() * 1000
```
nicopace, line 26: Might this affect the lime-app pirania integration? It looks that you are changing units, and the interface might depend on it...

spiccinini: Yes,  but current code is broken as far as I understand (int32 overflow in the voucher logic code)

spiccinini: Or I can comment/delete the tests and leave the code as is.

nicopace: Ok, please be aware of this change @luandro

G10h4ck (approved):

PR #739, opened 2020-08-09
Allow building a libremesh network with multiple qemu nodes

spiccinini:
Also dehardcode the wan interface


tools/qemu_dev_start
```diff
@@ -130,33 +145,48 @@ parse_commandline "$@"
...
     # DHCP server for WAN ifc
-    dnsmasq -F 172.99.0.100,172.99.0.100 --dhcp-option=3,172.99.0.1 -i lime_tap1 --dhcp-authoritative --log-dhcp
+    dnsmasq -F 172.99.0.100,172.99.0.100 --dhcp-option=3,172.99.0.1 -i "$WAN_IFC" --dhcp-authoritative --log-dhcp
 
     # enable forwarding and NAT
     echo 1 > /proc/sys/net/ipv4/ip_forward
     iptables -t nat -A POSTROUTING -o wlan0 -j MASQUERADE
     iptables -A FORWARD -m conntrack --ctstate RELATED,ESTABLISHED -j ACCEPT
-    iptables -A FORWARD -i  lime_tap1  -o wlan0 -j ACCEPT
+    iptables -A FORWARD -i  "$WAN_IFC"  -o wlan0 -j ACCEPT
```
nicopace, line 189: It assumes that the wireless interface is called wlan0 (in my computer for example is wlp2s0), and that that one is the access to the internet.

spiccinini: Yes, maybe we can use an program argument for this.

spiccinini: The WAN functionality is fragile also beacause of how the dhcp server is implemented (some computers have a dhcp server already binded to all the interfaces)

spiccinini: Fixed :)

tools/qemu_dev_stop
```diff
@@ -1 +1 @@
-echo system_powerdown | nc -N 127.0.0.1 45454
+echo system_powerdown | nc -N 127.0.0.1 45450
```
germanferrero, line 1: Should we add a parameter to kill nodes with node_number not equal to 0?

tools/qemu_dev_start
```diff
@@ -16,21 +16,26 @@
 #    along with this program.  If not, see <https://www.gnu.org/licenses/>.
 #
 # ARG_OPTIONAL_SINGLE([libremesh-workdir],[],[Optional path to a libremesh working directory to include it as a layer over the rootfs.])
-# ARG_OPTIONAL_BOOLEAN([enable-wan],[],[Enable wan (needs dnsmasq installed)],[])
+# ARG_OPTIONAL_SINGLE([node-number],[],[Optional node number from 0 to 9. Use this when running multiple qemu nodes.],[0])
```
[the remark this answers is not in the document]
spiccinini: yes, also maybe a better name?

TESTING.md
```diff
@@ -175,6 +175,12 @@ Prebuilt development images can be downloaded from here:
 
 Install the package `qemu-system-x86_64` if you don't have already installed.
 
+### Build a mesh network
+
+Up to 10 qemu nodes can be setup. Use the `--node-number N`. All the node's LAN interfaces are
+bridged together. You can use `--enable-wan` in only one of the nodes to share your internet connection
```
G10h4ck, line 181: Can we have only one upstream connection?

spiccinini: Not necesarly but the current code of the setup script yes has this limitation as it runs a dnsmasq listening on a fixed (so a second dnsmasq will collide) and also it uses a fixed IP.  
We can improve the script to give more than one node an upstream connection. We should check if dnsmasq is already running and use a different IP per node.

G10h4ck (approved):

PR #740, opened 2020-08-09
Fix unittests broken and use docker pull instead of local build

spiccinini:
Unittests are broken from #737 . I don't know how github allowed merging without a green status after the last push, or maybe travis did not run at each push.
This fixes this.
Also changes the behaviour of ./run_tests to use docker pull instead of docker build. This speed things up in  the first use.

### DIFF
--- a/packages/safe-upgrade/tests/test_safe-upgade.lua
+++ b/packages/safe-upgrade/tests/test_safe-upgade.lua
@@ -139,11 +139,6 @@ mtd8: 00010000 00010000 "ART"
         assert.are.same(expected, files_preserved)
     end)
 
-    it('test cmdline help', function()
-        local out = utils.unsafe_shell('lua ./packages/safe-upgrade/files/usr/sbin/safe-upgrade --help')
-        assert.not_nil(string.match(out, "Safe upgrade mechanism"))
-    end)
-
     before_each('', function()
         test_dir = test_utils.setup_test_dir()
     end)
--- a/run_tests
+++ b/run_tests
@@ -1,7 +1,9 @@
 #!/bin/bash
 DOCKER_IMAGE="libremesh/luatest:1.1"
 if ! docker inspect --type=image $DOCKER_IMAGE > /dev/null 2>&1; then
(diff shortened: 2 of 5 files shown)


packages/safe-upgrade/tests/test_safe-upgade.lua
```diff
@@ -139,11 +139,6 @@ mtd8: 00010000 00010000 "ART"
         assert.are.same(expected, files_preserved)
     end)
 
-    it('test cmdline help', function()
```
nicopace, line 142: and we can't add lua-argparse to the image?

run_tests
```diff
@@ -1,7 +1,9 @@
 #!/bin/bash
 DOCKER_IMAGE="libremesh/luatest:1.1"
 if ! docker inspect --type=image $DOCKER_IMAGE > /dev/null 2>&1; then
-    docker build -f Dockerfiles/Dockerfile.unittests -t "$DOCKER_IMAGE" .
+    echo "Pulling docker image '$DOCKER_IMAGE'."
+    echo "If you want to build it locally run 'docker build -f Dockerfiles/Dockerfile.unittests -t $DOCKER_IMAGE .'"
+    docker pull "$DOCKER_IMAGE"
```
nicopace, line 6: The thing about using pull instead of build is that if we update this image it will require us to update it manually. Doing the build doesn't require us to update the DOCKER_IMAGE value.
If it is an optimization that was required, we should do it... if not, it might not be that important. Still, your call :) :+1:

spiccinini: The image build process does not guarantee that every build makes exactly the same image. The build process tries to make it reproducible but the result may be different. 
So this way we are sure that everyone is using exactly the same images.
And of course the con is that we have to manually do a docker push when we update the image. I have been doing this but noone was using it :P

packages/safe-upgrade/tests/test_safe-upgade.lua
spiccinini: Yes, we have to host it somewhere (as this package is from lime-packages).

PR #748, opened 2020-08-25
Fix specific wireless radio config not having always priority

### LINKED ISSUE #746: Some specific radio configs (distance, etc) overrided by general configs
The following config should set a distance of 1000 but the resulting distance is 5000:
```
config lime 'wifi'
        option distance_5ghz '5000'

config wifi radio1
    option distance '1000'
```
This happens because the code looks first in the `_freqprefix` options and then in the configs without prefix. So the general config wins over the specific radio: `local distance = options["distance"..freqSuffix] or options["distance"] or 10000`

spiccinini:
Fixes #746 

Also added `htmode_2ghz` missing default.

### DIFF
--- a/packages/lime-system/files/etc/config/lime-defaults
+++ b/packages/lime-system/files/etc/config/lime-defaults
@@ -51,6 +51,7 @@ config lime wifi
 	option channel_2ghz '11'
 	list channel_5ghz '48'
 	list channel_5ghz '157'
+	option htmode_2ghz 'HT20'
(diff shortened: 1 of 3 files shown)


G10h4ck (approved):

ilario (approved):
Did not test but code looks perfect.

ilario:
I just tested, I can confirm the issue and that the patch fixes the issue :D

PR #756, opened 2020-09-01
lime-example improve documentation for WWAN

ilario:
I was configuring a wireless WAN on a router which already had an ethernet WAN port configured by lime-hwd-openwrt-wan, using an image based on OpenWrt 19.07.3.

With my surprise I noticed that the configuration section generated by lime-hwd-openwrt-wan for the ethernet WAN does not appear anywhere in /etc/config/lime-* but it still gets applied.
Is this normal?

Anyway, in order to have the WWAN indicated in /etc/config/network as the WAN interface, the ethernet WAN needs to be disabled and this PR indicates in /docs/lime-example how to disable it.


ilario:
Sorry, the history of this commit got dirty, give me some time to fix it.

ilario:
Ok, ready to be merged :)

G10h4ck (approved):

PR #757, opened 2020-09-04
lime-utils: fix get_node_status most_active field

germanferrero:
When we started using limewireless.mesh_ifaces() we broke the expected api.
This adapt lime-utils.get_node_status to the new api and recover most_active in the response


nicopace:
Great work @germanferrero .
Would be great to add a test of the broken functionality, so we don't break it again :D 
Could we add that to this PR?

germanferrero:
@nicopace I agree, I propose we address it soon and have created an Issue for it https://github.com/libremesh/lime-packages/issues/761

nicopace (approved):

PR #759, opened 2020-09-15
Replace nonworking change_config by set_hostname

spiccinini:
The change_config functionality was not working and was also incompatible with the lime-app.
Let's at least make changing the hostname work again.  Also moved this API behind root/community password.

Also add utils.is_valid_hostname() that checks that a hostname is also a  "DNS compatible"

Here is the lime-app PR https://github.com/libremesh/lime-app/pull/274


germanferrero (approved):
Tested!

packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils-admin
```diff
@@ -29,6 +30,18 @@ local function set_root_password(msg)
     return utils.printJson(result)
 end
 
+local function set_hostname(msg)
```
nicopace, line 33: isn't there any validation in relation to what string is a valid hostname?
Here some info: https://stackoverflow.com/a/3523068

nicopace (changes requested):
check the hostname validation request please. thanks!

packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils-admin
spiccinini: We have a validation in the lime-app side (well I am not actually sure about). But yes we may implement one also in the backend. "Sadly" linux accepts almost anything you put as a name, but then the name won't work as a domain name.

nicopace: yes... the same validation done in the userspace should happen here i think

spiccinini: I would say that we should do the validation only in the backend and not twice (or only in the frontend and if you want to shoot in your feet well....), but yes I understand that it may not play nicely with how apps work (thanks to the market we will have more work and bugs).
I will add the validation in the backend, hopefully compatible with the frontend validation.

PR #762, opened 2020-09-18
Fix shared-state get candidates neigh

spiccinini:
Use our own implementation of the conversion of mac to ipv6linklocal to workarround this bug in luci.ip.tolinklocal https://github.com/openwrt/luci/pull/4415

Also provided is the utility script `mac2ipv6ll`


G10h4ck (approved):

ilario:
Is this needed even now that openwrt/luci#4415 has been fixed?

spiccinini:
As far as I can see it was not applied to the 19.x and 18.x branches of luci so I would say that we need it at least for some months.

germanferrero (approved):

PR #772, opened 2020-10-03
Improvements to lime-sysupgrade

### LINKED ISSUE #769: search for a possebillaty to sysupgrade (-n) keep-settings-light  just with lime_node
for keep config clean
thanks.
and the keep the ssh-key would be nice too. but no other settings at all. 
my reason for this is, that i try to admin a mesh with 17 nodes and my most important settings are the node-name, ap ssid and the cannels. and if i make keep settings, the routers make strange things (for exapel do not schow the libremesh logo in the terminal)
and if i make sysupgade -n /tmp then the logo is there but i have to update and rename every router and have a very hard work to make a plan to do not lose the connections. there are more reasons for me but the list is to long and my english to bad.

ilario:
Fix #769 
Here goes three changes to lime-sysupgrade script from lime-system package:

1) when one of the files from a collection was missing, for example /root/.ssh/known_hosts from dropbear collection, a misleading warning was printed, for example:

`Warning: Cannot read /lib/upgrade/keep.d/dropbear file`

even if `/lib/upgrade/keep.d/dropbear` exists.

2) the checksums since OpenWrt 19.07 are SHA256 by default, so I added the output of sha256sum. I think that sha256sum was not present in OpenWrt 18.06, so I added a check for its presence.

3) in case the router was not supported by safe-upgrade (or did not have installed it), a broken check caused the lack of preservation of the configuration.

### DIFF
--- a/packages/lime-system/files/usr/bin/lime-sysupgrade
+++ b/packages/lime-system/files/usr/bin/lime-sysupgrade
@@ -11,7 +11,7 @@ BACKUP_EXTRA_DIR="/tmp/lime-sysupgrade/preserve/"
 
 add_file() {
   for l in `cat $1 | sed -e s/#.*$// -e /^$/d`; do
-    [ -e "$l" ] && BACKUP_FILES="$BACKUP_FILES $l"
+    [ -e "$l" ] && BACKUP_FILES="$BACKUP_FILES $l" || true
   done
 }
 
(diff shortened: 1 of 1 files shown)


spiccinini (approved):
Great thanks!

aparcar (approved):

Issue #325, opened 2018-02-11
Attended Sysupgrade inclusion


ilario:
[Attended Sysupgrade](https://blog.freifunk.net/2017/08/25/attended-sysupgrade-final-evaluation-update/) system helps the user to upgrade the router with one click. Developed by @aparcar 
What's missing:

* subdomain upgrade.libremesh.org for upgrade server (already accessible [here](https://ledeupdate.planetexpress.cc/))
* upgrade button in lime-app
* ?

aparcar:
@gmarcos87 can you help with the lime-app?

ilario:
@aparcar updates? Is this planned for inclusion in 18.06 release?

aparcar:
@ilario actually @gmarcos87 must do the magic, I'm not working on lime-app

ilario:
@aparcar @gmarcos87 what is missing?

ilario:
[luci-app-attendedsysupgrade](https://github.com/openwrt/luci/tree/openwrt-18.06/applications/luci-app-attendedsysupgrade) is already present in OpenWrt 18.06 and there's also stuff related to ubus that I don't understand in the [attendedsysupgrade-common](https://github.com/openwrt/packages/tree/openwrt-18.06/utils/attendedsysupgrade-common) package which is also included in OpenWrt 18.06.

ilario:
@spiccinini is there any planned automatic updating method?
In this still the one that is going to be used at some point in the future?

PR #773, opened 2020-10-08
Add easy upgrades support

spiccinini:


ilario:
Sounds good, can you tell us what is this?
Is this related to #325?

spiccinini:
> Sounds good, can you tell us what is this?
> Is this related to #325?

This is related, yes. This package provides ubus and commands for downloading and validating signatures of upgrades. The mechanism is somehow similar to unnatended upgrades but without the ability to request specific builds dinamically.

packages/eupgrade/files/usr/lib/lua/eupgrade.lua
```diff
@@ -0,0 +1,177 @@
...
+eup.STATUS_DEFAULT = 'not-initiated'
+eup.STATUS_DOWNLOADING = 'downloading'
+eup.STATUS_DOWNLOADED = 'downloaded'
+eup.STATUS_DOWNLOAD_FAILED = 'download-failed'
+
+
+local uci = libuci.cursor()
+
+function eup.set_workdir(workdir)
+    if not utils.file_exists(workdir) then
```
G10h4ck, line 16: what if exists but it is not dir?

packages/eupgrade/Makefile
```diff
@@ -0,0 +1,24 @@
...
+include ../../libremesh.mk
+
+PKG_NAME:=eupgrade
+
+define Package/$(PKG_NAME)
+  SECTION:=utils
+  CATEGORY:=Utilities
+  TITLE:=$(PKG_NAME) provides semi automated firmware upgrades
+  MAINTAINER:=Santiago Piccinini <spiccinini@altermundi.net>
+  DEPENDS:=+lua +lime-system
```
G10h4ck, line 16: Is this yet another package which only on `lime.utils`?

G10h4ck (approved):

packages/eupgrade/Makefile
spiccinini: Yes!

packages/eupgrade/files/usr/lib/lua/eupgrade.lua
spiccinini: What do you suggest, to loga warning if it is not a directory?

G10h4ck: in that case it probably should fail all together and report an error

germanferrero (approved):

Issue #747, opened 2020-08-24
Removing lime-community and lime-node files from lime-system package


ilario:
This is more a question than a proposal:
why the `/etc/config/lime-community` and `/etc/config/lime-node` are included in `lime-system` package?

Isn't this likely going to conflict with their versions included in the [network-profiles](https://github.com/libremesh/network-profiles/find/master) if the network-profiles are included at build time (either from source via buildroot or as packages via SDK)?

Wouldn't be better to not forcefully require them (so they would not exist unless a user creates them) or to create them at the first boot (if they are really needed)?

spiccinini:
> This is more a question than a proposal:
> why the `/etc/config/lime-community` and `/etc/config/lime-node` are included in `lime-system` package?

Well I see some reasons for thouse files to be in a package (not necesarly lime-system):
* UCI needs the files to be present to be able to "use or modify them" (with uci set, etc).
* In general is good to have all the files in the system defined in a package. This simplifies discoverability, thinking about the system, ease maintenance.

> Isn't this likely going to conflict with their versions included in the [network-profiles](https://github.com/libremesh/network-profiles/find/master) if the network-profiles are included at build time (either from source via buildroot or as packages via SDK)?

If the files are included from source using the `files/` directory it won't be a conflict (as everything in this directory overwrites what it is in the packages). But yes it will conflict if using the network-profiles through packages via SDK. I see that there may be interest in installing and upgrading the network-profiles through opkg, I don't use this way.

> Wouldn't be better to not forcefully require them (so they would not exist unless a user creates them) or to create them at the first boot (if they are really needed)?

In my opinion If they don't exist in the system it would be a lot more difficult to modify them in the target.
Creating them at the first boot is not ideal, it adds more complexity.

It would be good to be able to use the opkg Provides and Replaces for this.
For example we may create a `lime-config` package that `PROVIDES:=libremesh-config` and then it would be able to have 
packages for each network like `network-profile-foo` that `PROVIDES:=libremesh-config` and `REPLACES:=lime-config`.
I did a test of this and it works but openwrt ipkg glue does not support `REPLACES` :(  I don't understand why as openwrt's opkg do supports Replaces, it is just a limitation of the buildroot glue that creates the metadata of the package.

WIth this patch it works:
```
--- a/include/package-ipkg.mk
+++ b/include/package-ipkg.mk
@@ -159,6 +159,7 @@ Version: $(VERSION)
[...]

PR #774, opened 2020-10-11
configs: ship configs in usr/share and populate if needed

spiccinini:
I am not fully happy with this change but it may be helpful solving #747 

I did not test yet building a firmware not with network-profiles (I am not using network-profiles) to check. If someone can test great, as I won't have time this week.


aparcar:
I don't have a test setup. What is the advantage of this approach?

spiccinini:
> I don't have a test setup. What is the advantage of this approach?

This enable to provide lime-node and lime-community in the network-profiles packages. Do you suggest other approach?

aparcar (approved):

aparcar:
Sounds great

spiccinini:
> Sounds great

Thanks for the review. I won't mare it until is tested.

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -198,8 +189,17 @@ function config.uci_commit_all()
 end
 
 function config.main()
-	config.sanitize()
-	config.sanitize(config.UCI_NODE_NAME)
+	config.initialize_config_file(config.UCI_AUTOGEN_NAME)
```
ilario, line 192: Why are you initializing lime-autogen if a few lines after config.uci_autogen() will empty it?

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -41,23 +41,14 @@ function config.get_config_path()
...
-	f:write('')
-	f:close()
-	config.sanitize(config.UCI_NODE_NAME)
+	config.initialize_config_file(config.UCI_NODE_NAME)
 end
 
---! Minimal /etc/config/lime santitizing
-function config.sanitize(config_name)
-	local config_name = config_name or config.UCI_AUTOGEN_NAME
+function config.initialize_config_file(config_name)
```
ilario, line 47: Why should we have both initialize_config_file and reset_node_config if they do the same function?
I noticed that reset_node_config is being used by first-boot-wizard, so that name should not change.
In my opinion, rather than having reset_node_config calling initialize_config_file, this function can be done directly from reset_node_config and we can remove both config.initialize_config_file and config.sanitize.

ilario:
> What is the advantage of this approach?

The advantage is that we can include lime-community files directly in the network-profiles, without the need to use uci-defaults scripts (e.g. [this one](https://github.com/libremesh/network-profiles/blob/master/libremesh/encrypt-11s%2Bap/etc/uci-defaults/80-set-11s-encryption)) to set the options.

For the testing I did the following:

using the lime-packages master branch (without these edits) selected the `LibreMesh/network-profiles/profile-es.hackmeeting.org-hm2019generic` which [includes an /etc/config/lime-community file](https://github.com/libremesh/network-profiles/blob/master/es.hackmeeting.org/hm2019generic/root/etc/config/lime-community), the network-profiles feed is now suggested even in the development instructions, since https://github.com/libremesh/lime-web/commit/0b9a082d10f610dab30c94be146da2763717432c .
This resulted in the expected make error:
```
Collected errors:
 * check_data_file_clashes: Package profile-es.hackmeeting.org-hm2019generic wants to install file /home/ilario/software/openwrt/build_dir/target-mips_24kc_musl/root-ath79/etc/config/lime-community
	But that file is already provided by package  * lime-system
 * opkg_install_cmd: Cannot install package profile-es.hackmeeting.org-hm2019generic.
```

using the branch from this PR and selecting the `profile-es.hackmeeting.org-hm2019generic` package, no error arose. Flashing the router with the obtained image resulted in a working router with the network-profile configuration in place. Using the lime-sysupgrade utility (manually taking its latest version) I then flashed an image an image without the network-profile and the custom lime-community was still in place, as it was the manually modified lime-node. Then (after re-flashing the first image with sysupgrade -n) I flashed using lime-sysupgrade an image with another network-profile (LibreMesh/network-profiles/profile-libremesh-default) and the lime-community from the first flashing was kept in place.

using the branch from this PR and not selecting any network-profile no error arose. Flashing the router (without preserving anything) resulted in a normal image with the /etc/config/lime-* files being as usual (lime-defaults 70 lines, lime-community 16 lines, lime-node 16 lines, lime-autogen 68 lines).

So it looks good to me :D

I added a couple of comments in the code.

Thanks!!!

packages/lime-system/files/usr/lib/lua/lime/config.lua
spiccinini: Right!, just because I didn't read all the code I only changed what it was needed. Will fix

packages/lime-system/files/usr/lib/lua/lime/config.lua
spiccinini: Because I think that reset_node_config may do other things than just initializing one config file. Also now that I am thinking more clearly this initialization may be done copying the template from usr/share. I will do this, let me know what do you think.

spiccinini:
Thanks for testing! Great that it is everything working now.

ilario (approved):

PR #776, opened 2020-10-12
Set LibreMesh packages repository location

### LINKED ISSUE #767: Set new location of repository for compiled packages
Thanks to @aparcar we have a compiled packages and network-profiles repository (see #728 and https://github.com/libremesh/network-profiles/pull/64) on https://github.com/libremesh/lime-feed

The location location of the old repository needs to be replaced, it is here:
https://github.com/libremesh/lime-packages/blob/903655831d74098ea2eba63da36d905636c6d4d0/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos#L25-L31

The files on the new repository cannot be listed (currently) but can be accessed using their explicit name, for example: https://libremesh.github.io/lime-feed/master/Packages.manifest . This should not be a problem for OPKG.

ilario:
Thanks to @aparcar and @spiccinini now we have again a binary repository.
This sets its location in the firmware so that OPKG can use it.
Fix #767

### DIFF
--- a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
+++ b/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
@@ -14,13 +14,12 @@ feeds_file="/etc/opkg/limefeeds.conf"
 
 [ "$LIME_CODENAME" == "development" ] && {
 	base_url="http://feed.libremesh.org/master";
-	key_name="a71b3c8285abd28b"
-	key_content="RWSnGzyChavSiyQ+vLk3x7F0NqcLa4kKyXCdriThMhO78ldHgxGljM/8"
 } || {
 	base_url="http://feed.libremesh.org/$LIME_RELEASE"
-	key_name="a71b3c8285abd28b"
-	key_content="RWSnGzyChavSiyQ+vLk3x7F0NqcLa4kKyXCdriThMhO78ldHgxGljM/8"
 }
+
+key_name="a71b3c8285abd28b"
+key_content="RWSnGzyChavSiyQ+vLk3x7F0NqcLa4kKyXCdriThMhO78ldHgxGljM/8"
 
 echo "Configuring official LibreMesh opkg feeds"
 echo "src/gz libremesh $base_url/libremesh" > "$feeds_file"
--- a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
+++ b/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
@@ -22,11 +22,11 @@ ARCH="$OPENWRT_ARCH"
 }
 
 [ "$LIME_CODENAME" == "development" ] && {
-	base_url="http://snapshots.libremesh.org/packages/$ARCH";
-	key_name="7546f62c3d9f56b1"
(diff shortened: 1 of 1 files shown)


aparcar (approved):

ilario:
I could not test if the second case works (the one in which a LibreMesh release is being used) as currently there is not build for any release in https://github.com/libremesh/lime-feed but it should work.

packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
```diff
@@ -10,28 +12,17 @@ feeds_file="/etc/opkg/limefeeds.conf"
...
 [ "$LIME_CODENAME" == "development" ] && {
-	base_url="http://snapshots.libremesh.org/packages/$ARCH";
-	key_name="7546f62c3d9f56b1"
-	key_content="RWR1RvYsPZ9WsTsmoVajHX9dyl2wL7grjcfFua1q8A99RWr2gF94lRdJ"
+	base_url="http://feed.libremesh.org/master";
+	key_name="a71b3c8285abd28b"
+	key_content="RWSnGzyChavSiyQ+vLk3x7F0NqcLa4kKyXCdriThMhO78ldHgxGljM/8"
 } || {
-	base_url="http://repo.libremesh.org/releases/$LIME_RELEASE/packages/$ARCH"
+	base_url="http://feed.libremesh.org/$LIME_RELEASE"
```
spiccinini, line 20: there is no more need in providing a ARCH dependent URL because most of the packages are ARCH agnostic right?

ilario: Yep, all the packages have `PKGARCH:=all` except `soporteremoto` (for no reason at all) and `tinc` (don't know why, anyway an other version of the same package is also present upstream: https://github.com/openwrt/packages/blob/master/net/tinc/Makefile).
Anyway this should be considered first in the Github Action on https://github.com/libremesh/lime-packages/blob/master/.github/workflows/build.yml

spiccinini: Perfect thanks for the clarification

spiccinini: As we are using the same key maybe we can just add they key to files/etc/opkg/keys instead ?

aparcar: Related https://github.com/libremesh/lime-packages/pull/777

ilario: > As we are using the same key maybe we can just add they key to files/etc/opkg/keys instead ?

I moved the key variables out of the if block. I would not add it into files/etc/opkg/keys as I'm not sure it will not conflict with other OpenWrt packages... Do you know that this will not happen?

spiccinini: I am not 100% sure that this will not happen.

spiccinini (approved):

packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
ilario: I just realized that neither the auto-usb-wwan have the PKGARCH specified.

PR #775, opened 2020-10-12
Update category of packages to category in libremesh.mk, de-hardcode lime-system dependency and review lime-system dependency

ilario:
As highlighted by @jkimmel in https://github.com/libremesh/lime-web/pull/112 the packages category included in `libremesh.mk` file by @aparcar is `LibreMesh` instead of the old `LiMe`.
This observation increased the priority of #766 to a release-blocker.
With this PR, I applied the new category to all the packages without actually migrating them to `libremesh.mk` (see #766). While this migration is still needed, now it can have a much lower priority and stop being a release-blocker.
Some of these packages should not be in the LibreMesh category but this is another issue: #768 

As suggested by @dangowrt in https://github.com/libremesh/lime-packages/issues/766#issuecomment-700214316, I removed the hardcoded dependency `+lime-system` from `libremesh.mk` enabling its usage for non-LibreMesh specific packages.

Additionally, in https://github.com/libremesh/lime-packages/issues/766#issuecomment-702940526 I realized that some of the packages should depend upon `lime-system` and I took advantage of this PR to add the dependency. If you prefer, I can do a separate PR.

Issue #766, opened 2020-09-28
Migrate LibreMesh-specific packages' Makefile to libremesh.mk new format


ilario:
On #729 a new format for the Makefile of LibreMesh packages has been proposed.
Here is a list of the packages which depends on `lime-system` package and which would be good to migrate to the new format.
It could be possible that for some of these packages the migration is not possible, please check each case.

- [x] lime-proto-babeld #729
- [x] lime-proto-batadv #729
- [x] lime-proto-bmx6 #729
- [x] lime-proto-bmx7 #729
- [x] lime-proto-wan #729
- [x] lime-smart-wifi #729
- [ ] deferable-reboot
- [ ] lime-hwd-ground-routing
- [ ] lime-hwd-openwrt-wan
- [ ] lime-hwd-usbradio
- [ ] lime-proto-anygw
- [ ] lime-proto-bgp
- [ ] lime-proto-olsr2
- [ ] lime-proto-olsr6
- [ ] lime-proto-olsr
- [ ] lime-webui
- [ ] safe-upgrade
- [ ] ubus-lime-batman-adv
- [ ] ubus-lime-bmx6
- [ ] wifi-unstuck-wa

Additionally, there are packages with `lime` in their name but which suspiciously do not explicitly depend on `lime-system`, these also should be checked and maybe migrated also:

lime-app
lime-ap-watchping
lime-debug (metapackage, do not require lime-system)
lime-docs (just docs, do not require lime-system)
lime-map-agent
lime-report (works also without LibreMesh stuff)
luci-app-lime-location
ubus-lime-fbw
ubus-lime-groundrouting
ubus-lime-location
ubus-lime-metrics
ubus-lime-openairview
ubus-lime-utils

dangowrt:
Why only use `libremesh.mk` for packages which depend on `lime-system`? I thought it was mostly to automate versioning (like for luci-* packages)...?

ilario:
> Why only use `libremesh.mk` for packages which depend on `lime-system`?

I'm not sure I got the question, but I try to answer: inside `libremesh.mk` there is a hardcoded dependency on `lime-system`.

I think that we should add also another .mk file identical but without `lime-system` dependency (what should be the filename?) to be used in the aforementioned packages: the LibreMesh-specific but not depending on lime-system.

[...]

PR #778, opened 2020-10-12
soporterremoto: convert to libremesh.mk

aparcar:

### DIFF
--- a/packages/soporteremoto/Makefile
+++ b/packages/soporteremoto/Makefile
@@ -1,39 +1,15 @@
-#
-# Copyright (C) 2006-2013 OpenWrt.org
-#
-# This is free software, licensed under the GNU General Public License v3.
-#
+LIME_TITLE:=soporteremoto allows AlterMundi to provide remote support through SSH root access
+LIME_MAINTAINER:=Gui Iribarren <gui@altermundi.net>
+LIME_DEPENDS:= +zerotier +auto-usb-wwan +kmod-ath9k-htc +ath9k-htc-firmware
+LIME_SECTION:=utils
+LIME_CATEGORY:=Utilities
 
-include $(TOPDIR)/rules.mk
-
-PKG_NAME:=soporteremoto
-PKG_VERSION=$(GIT_COMMIT_DATE)-$(GIT_COMMIT_TSTAMP)
-GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
-GIT_COMMIT_TSTAMP:=$(shell git log -n 1 --pretty=%at . )
-
-include $(INCLUDE_DIR)/package.mk
-
-define Package/$(PKG_NAME)
-  SECTION:=utils
-  CATEGORY:=Utilities
-  TITLE:=$(PKG_NAME) allows AlterMundi to provide remote support through SSH root access
-  MAINTAINER:=Gui Iribarren <gui@altermundi.net>
-  DEPENDS:= +zerotier +auto-usb-wwan +kmod-ath9k-htc +ath9k-htc-firmware
-endef
-
-define Package/$(PKG_NAME)/description
+define LIME_DESCRIPTION
   Installs auto-usb-wwan to provide a zero-configuration WWAN via USB dongle
-  zerotier to provide VPN transport
-  and installs a predefined set of ssh authorized_keys of AlterMundi members
-  that will be granted root access
+  zerotier to provide VPN transport and installs a predefined set of ssh
+  authorized_keys of AlterMundi members that will be granted root access.
 endef
 
-define Build/Compile
-endef
+include ../../libremesh.mk
 
-define Package/$(PKG_NAME)/install
-	$(INSTALL_DIR) $(1)/
-	$(CP) ./files/* $(1)/
-endef
-
-$(eval $(call BuildPackage,$(PKG_NAME)))
+# call BuildPackage - OpenWrt buildroot signature


packages/soporteremoto/Makefile
```diff
@@ -1,39 +1,13 @@
...
 endef
 
-define Package/$(PKG_NAME)/install
-	$(INSTALL_DIR) $(1)/
-	$(CP) ./files/* $(1)/
-endef
+include ../../libremesh.mk
 
-$(eval $(call BuildPackage,$(PKG_NAME)))
+# call BuildPackage - OpenWrt buildroot signature
```
spiccinini, line 13: I have a semi-off topic question. If I want to override the `install` definition, what is the way to do it?

Addin the install definition after the include of libremesh.mk and Calling `$(eval $(call BuildPackage,$(PKG_NAME)))` again here?

aparcar: I think after the inclusion similar to [here](https://github.com/openwrt/luci/blob/7edd635026594da577b73f059c0d8b95d653f8fc/modules/luci-base/Makefile). However can't you improve the `libremesh.mk` version?

spiccinini: Thanks, yes maybe I can improve the LibreMesh.mk version (after understanding how does it work), I was just asking not implying it can't be done. I don't know if it is only me or makefiles are very difficult to understand, follow and debug so I try to maintain them as tiny and simple as possible.

ilario:
Thanks for this!
But this package should be in `SECTION:=utils` and `CATEGORY:=Utilities` which is not possible with current `libremesh.mk`.
@aparcar could you edit the `libremesh.mk` to allow the definition of `LIME_SECTION` and `LIME_CATEGORY`? Their fallback values can well be `lime` and `LibreMesh` (as they current values).

aparcar:
I added the requested change in #779 , however I don't see a big gain by distributing `libremesh.mk` packages over multiple sections.

ilario:
> I added the requested change in #779 

Thanks!
So this PR has to be merged **after** #779

> however I don't see a big gain by distributing `libremesh.mk` packages over multiple sections.

Well, the discussion on this is in #766 (check out dangowrt's comments) and the +lime-system hardcoded dependency is being removed in #775.

ilario (approved):

packages/soporteremoto/Makefile
ilario: I also find Makefiles an annoying and complex overhead, that's why I love the approach to have a centralised one in libremesh.mk :)

G10h4ck:
I don't get the advantage of using `LIME_TITLE LIME_MAINTAINER LIME_DEPENDS LIME_SECTION LIME_CATEGORY` versus the without `LIME_` prefix ones, as of today the only advantage of having `libremesh.mk` is having a default install procedure and avoid the versioning code duplication, all the rest seems just more breakage surface, and steeper learning curve for openwrt developers... moreover packages that uses `libremesh.mk` are less upstreamizable, a few of this issues had been pointed out also in the PR that added `libremesh.mk` for first time but where more or less ignored...

aparcar:
> I don't get the advantage of using LIME_TITLE LIME_MAINTAINER LIME_DEPENDS LIME_SECTION LIME_CATEGORY versus the without LIME_ prefix ones, as of today the only advantage of having libremesh.mk is having a default install procedure and avoid the versioning code duplication

The variables have to be defined in `define Package/$(PKG_NAME)`. Everything else would define them globally or you define ` define Package/$(PKG_NAME)` which than forces you to add redundant information.

> ...all the rest seems just more breakage surface, and steeper learning curve for openwrt developers...

The steepness should be close to zero as it's the same concept as for LuCI or shouldn't it?

 > moreover packages that uses libremesh.mk are less upstreamizable

I'm a somewhat late member (2 years in) of LibreMesh and have never seen any efforts to get things up-streamed. When was the last try? Freifunk just recently added their own feed to the upstream repositories. But sure, whatever is LibreMesh independent should become part of `packages.git`, PRs welcome.

dangowrt:
> > I don't get the advantage of using LIME_TITLE LIME_MAINTAINER LIME_DEPENDS LIME_SECTION LIME_CATEGORY versus the without LIME_ prefix ones, as of today the only advantage of having libremesh.mk is having a default install procedure and avoid the versioning code duplication
> 
> The variables have to be defined in `define Package/$(PKG_NAME)`. Everything else would define them globally or you define ` define Package/$(PKG_NAME)` which than forces you to add redundant information.
> 
> > ...all the rest seems just more breakage surface, and steeper learning curve for openwrt developers...
> 
> The steepness should be close to zero as it's the same concept as for LuCI or shouldn't it?

This is true for people having previously dealt with OpenWrt or LuCI sources.
However, the pink elephant in the room here is the assumption that everyone potentially contributing to or modifying LibreMesh has touched OpenWrt or LuCI before. It was also one of my fundamental illusions that this would be the case... In reality, many users downstream are somehow familiar with TCP/IP networking and UNIX in general, but have never touched OpenWrt before...

> 
> > moreover packages that uses libremesh.mk are less upstreamizable
> 
> I'm a somewhat late member (2 years in) of LibreMesh and have never seen any efforts to get things up-streamed. When was the last try? Freifunk just recently added their own feed to the upstream repositories. 

Not exactly. All the (mostly abandoned) freifunk stuff was previously part of the openwrt/luci repository and has been removed from there...

> But sure, whatever is LibreMesh independent should become part of `packages.git`, PRs welcome.

+1

aparcar:
I'll close this, mu involvement in LibreMesh is basically non existent.

PR #779, opened 2020-10-13
libremesh.mk: allow change of SECTION/CATEGORY

aparcar:
Using LIME_{SECTION,CATEGORY} it is now possible to change the defaults.

### DIFF
--- a/libremesh.mk
+++ b/libremesh.mk
@@ -29,8 +29,8 @@ PKG_BUILD_DIR:=$(BUILD_DIR)/$(PKG_NAME)
 include $(INCLUDE_DIR)/package.mk
 
 define Package/$(PKG_NAME)
-  SECTION:=lime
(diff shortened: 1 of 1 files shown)


ilario:
Thanks!
I like better the `LibreMesh` category rather than the `LiMe` category, do you agree?
Can you change it?
The rest of the packages are already being transitioned towards the `LibreMesh` category in #775 as a temporary measure waiting for #766 to be completed.

aparcar:
Done

ilario (approved):

spiccinini (approved):

spiccinini:
I still think that customization of the package definition should be done directly in the Makefile and not exposing a LIME_FOO for each option....it does not make sense to me, it adds more complexity than reducing it. For me what we need to do is to just provide sane defaults.

PR #781, opened 2020-10-15
Add network-profiles packages repo to OPKG feeds

ilario:
Sorry for not thinking this before but we can add also the packaged network-profiles repository to OPKG available packages.

Then, in order to install a network profile using OPKG it could be necessary to use the `--force-overwrite` option from `opkg` command. This is needed only for profiles providing files already existing in the router (like [this](https://github.com/libremesh/network-profiles/tree/master/libremesh/default/root/etc/config)) but will not be needed for some other network-profiles which set the options in other ways (like [this](https://github.com/libremesh/network-profiles/tree/master/libremesh/encrypt-11s/root/etc/uci-defaults)).

See also discussion on #747

### DIFF
--- a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
+++ b/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
@@ -13,15 +13,18 @@ feeds_file="/etc/opkg/limefeeds.conf"
 }
 
 [ "$LIME_CODENAME" == "development" ] && {
-	base_url="http://feed.libremesh.org/master";
+	packages_url="http://feed.libremesh.org/master";
 } || {
-	base_url="http://feed.libremesh.org/$LIME_RELEASE"
+	packages_url="http://feed.libremesh.org/$LIME_RELEASE"
 }
+
+profiles_url="http://feed.libremesh.org/profiles"
 
 key_name="a71b3c8285abd28b"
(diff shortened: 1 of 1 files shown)


packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
```diff
@@ -13,15 +13,18 @@ feeds_file="/etc/opkg/limefeeds.conf"
...
 
+profiles_url="http://feed.libremesh.org/profiles"
+
 key_name="a71b3c8285abd28b"
 key_content="RWSnGzyChavSiyQ+vLk3x7F0NqcLa4kKyXCdriThMhO78ldHgxGljM/8"
 
 echo "Configuring official LibreMesh opkg feeds"
-echo "src/gz libremesh $base_url" > "$feeds_file"
+echo "src/gz libremesh $packages_url" > "$feeds_file"
+echo "src/gz profiles $profiles_url" > "$feeds_file"
```
spiccinini, line 28: Concatenation with `>>` should be used right?

spiccinini (changes requested):

packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
ilario: UOOOPPPSSS yesss! Thanks!

aparcar (approved):

spiccinini (approved):

PR #782, opened 2020-10-15
SSH banner: Added pointer to offline documentation in /docs

### LINKED ISSUE #616: Point out in SSH banner that the documentation is in /docs
See discussion on #195

ilario:
Fix #616


spiccinini (approved):
great!

ilario:
Hey, I just thought: the line I added is only true when lime-docs* is installed, which is very likely but not 100% sure. Do you think it's worth to use an uci-defaults script in lime-docs-minimal to append this line to the banner? Or this way is good enough?

aparcar:
Yea I'd say make it optional that it only shows up if /docs exists.

ilario:
> Yea I'd say make it optional that it only shows up if /docs exists.

Mh, why is this better than using another uci-defaults script in lime-docs-minimal package?

aparcar:
I guess deleting the docs later on would cause to keep the message, but as that's unlikely I'm fine with both.

ilario:
I moved the thing to the lime-docs-minimal package.
And I took advantage to advertise that the docs are accessible also via web interface and to rename lime-example to lime-example.txt so that the browsers open it as a text file (instead than proposing the download as a binary file).

aparcar (approved):

PR #790, opened 2020-10-23
Update Babeld option "wired" -> "type"

ilario:
See original report on https://lists.libremesh.org/pipermail/lime-users/2020-October/001741.html
I tested coarsely, also because I have no idea on how to check if the option is actually doing anything.

### DIFF
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -124,7 +124,7 @@ function babeld.setup_interface(ifname, args)
 	--! It is quite common to have dummy radio device attached via ethernet so
 	--! disable wired optimization always as it would consider the link down at
 	--! first packet lost
-	uci:set("babeld", owrtInterfaceName, "wired", "false")
+	uci:set("babeld", owrtInterfaceName, "type", "wireless")
 
 	uci:save("babeld")
 end


aparcar (approved):

ilario:
I checked if this should work also on the Babeld version from OpenWrt 18.06 and yes, it should.

In [OpenWrt 19.07, Babeld 1.9.2 is used](https://github.com/openwrt-routing/packages/blob/openwrt-19.07/babeld/Makefile) and in there, the [usage of the wired option is deprecated](https://github.com/jech/babeld/blob/babeld-1.9.2/configuration.c#L561).

In [OpenWrt 18.06, Babeld 1.8.5 is used](https://github.com/openwrt-routing/packages/blob/openwrt-18.06/babeld/Makefile) and in there, the [usage of the wired option is deprecated](https://github.com/jech/babeld/blob/babeld-1.8.5/configuration.c#L550).

spiccinini (approved):

G10h4ck:
@ilario @aparcar @spiccinini did you check/test if the openwrt babeld init script support the `type` option too?

AFAIR when I wrote the lime module babeld C code already deprecated the `wired` option but the Openwrt babeld init script didn't seem support the `type` option yet.

In current openwrt-routing repository init script https://github.com/openwrt-routing/packages/blob/master/babeld/files/babeld.init#L129 it seems that the option would be just passed down "as is" but that is just what it seems to  me at first glance.

Be very careful with this kind of changes, they seems very little but they may unexpectedly break entire networks in a non trivial to debug way at next update.

spiccinini:
I have tested ir only with the LibreRouterOs babeld init script that is not the upstreap script. Maybe que can revert this until que are very sure that it is working?

spiccinini:
@ilario can you test using  `wirelesstypo` instead of `wireless` and see if it does not start (so we can conclude that the babeld script is working) ?

ilario:
> @ilario can you test using `wirelesstypo` instead of `wireless` and see if it does not start (so we can conclude that the babeld script is working) ?

Yes it works also for me.
With the typo I have:
`Couldn't parse configuration from file /var/etc/babeld.conf (error at line 4).`
without the typo I have:
```
Warning: couldn't determine channel of interface eth0-1_17.
Warning: couldn't determine channel of interface wlan0-mesh_17.
Warning: couldn't determine channel of interface wlan1-mesh_17.
Warning: couldn't determine channel of interface eth0_17.
Warning: couldn't determine channel of interface eth0-2_17.
```

spiccinini:
Great thanks!

spiccinini:
I've cherrypicked this into 20.09

Issue #763, opened 2020-09-20
Lime-App, Un unexpected error occurred. 


rallep71:
Lime-App, Un unexpected error occurred. 

Hello,

once again I have a problem with the display in the browser of the Lime-App. I don't need the First-Boot-Wizzard, that's why it is not compiled. I only need to see in which node I am located in the Lime-App. See photo. 

systems:
Openwrt 19.07.4, following routers are in use, TP-Link Archer C50 v3, TP-Link Archer C50 v4, TP-Link TL-WDR4300 v1.

Config File:(this file is for C50, but identical except for the different routers)
https://pastebin.com/sdpGx1Ca
feeds.conf:
src-git packages https://git.openwrt.org/feed/packages.git^eae1bb397f064befe003b01a36ae318321e81d35
src-git luci https://git.openwrt.org/project/luci.git^0d0ab01a64f25d15325d3c98a7df252085382f1d
src-git routing https://git.openwrt.org/feed/routing.git^b77498bd56d5e45ab4577a1f4ad6ffc55b4a86b7
src-git telephony https://git.openwrt.org/feed/telephony.git^91b34e7eefa1d37a18390d0c6881978cd76804a1
src-git libremesh https://github.com/libremesh/lime-packages.git
src-git libremap https://github.com/libremesh/libremap-agent.git
src-git networkprofiles https://github.com/libremesh/network-profiles.git

![lime](https://user-images.githubusercontent.com/2786491/93696802-ae935e80-fb15-11ea-89a6-d8ca0ded2b90.png)

one more thing i have to say, i used the network profile of libremesh to encrypt mesh, 

80-set-11s-encryption

#!/bin/sh
uci set lime-defaults.wifi.ieee80211s_encryption='psk2+aes'
uci set lime-defaults.wifi.ieee80211s_key='libremesh'
uci commit lime-defaults

have I forgotten to compile anything, or what could I do?

THX

spiccinini:
Hi @rallep71 indeed the lime-app error is related to firstbootwizard not being installed. Thanks for reporting it :)

In order to use encrypted mesh, the wpad-mini package have to be replaced with wpad-mesh-wolfssl package.

rallep71:
Thanks for the quick answer!
I had mentioned it before in another posting, I am very impressed with libremesh! I have been looking for solutions to build a mesh for over a year. Sure, there are ready-made solutions from different vendors including hardware, but I don't have control over them, and usually after two or three years there are no more updates. 

[...]

PR #791, opened 2020-10-24
firstbootwizard: now needs root privilege to run

germanferrero:
This PR changes acl for firstbootwizard, in order to fix the current privilege escalation reported at: https://github.com/libremesh/lime-packages/issues/763#issuecomment-696670006
It should be merged along with https://github.com/libremesh/lime-app/pull/277 at the LimeApp


spiccinini (approved):

G10h4ck (approved):

dangowrt:
As root is always allowed to access everything on ubus, it really doesn't make any sense to add an ubus acl rule for user 'root'. Why not just remove that completely?

Issue #786, opened 2020-10-20
FBW initial prompt: add a "Do not ask anymore" button


ilario:
Currently, in the FirstBootWizard web interface the user is welcomed with "Please configure your network" prompt.
The possible interactions are "OK" and "CANCEL".
A button like "Stop asking, I'll configure my router by hand!" is missing.
Up to now the way to do this is editing the /etc/config/lime-node  
https://github.com/libremesh/lime-packages/blob/6de76482829a774d9dec071c5d442db1c35da146/packages/lime-docs/files/www/docs/lime-example.txt#L25
setting:

```
config lime system
	option firstbootwizard_configured true
```

as commented in https://github.com/libremesh/lime-packages/issues/640#issuecomment-557199692

ilario:
https://github.com/libremesh/lime-packages/pull/792 has been merged.
If I got it correctly, once https://github.com/libremesh/lime-app/pull/278 is also merged this will be fixed.

germanferrero:
It got merged, this issue is now solved :)

PR #792, opened 2020-10-24
Fbw add dismiss option

germanferrero:
This PR adds the ability to dismiss fbw from an ubus endpoint, which will be called by the LimeApp. Addressing #786 
The corresponding PR in the limeapp side is https://github.com/libremesh/lime-app/pull/278

Comments and suggestions about the implementation are welcomed. I thought it would be nice to keep track of whether the fbw has been performed firstbootwizard_configured=true or it has been dismissed firstbootwizard_dismissed=true.
Also, when dismissing, the config option is not reloaded until a reboot or restart of the rpcd service, so a temp file is used to reflect the dismiss too.

Disclaimer: This branch is rebased with the one of #791. They introduce changes in the same acl

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -25,6 +25,7 @@ local fbw = {}
 
 
 fbw.WORKDIR = '/tmp/fbw/'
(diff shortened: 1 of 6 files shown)


spiccinini:
Great @germanferrero !  I will be reviewing ASAP

Probably not to be done in this PR but now that we can do generic configs in libremesh we can use that to remove the options from the lime system config (as they are not core options).

spiccinini (approved):

G10h4ck (approved):

Issue #788, opened 2020-10-23
libremap-agent errors


ilario:
In the logs from logread of my router I can see:

```
user.warn libremap-agent[2792]: unable to load plugin "location"; module 'luci.libremap.plugins.location' not found: 	no field package.preload['luci.libremap.plugins.location'] 	no file './luci/libremap/plugins/location.lua' 	no file '/usr/share/lua/luci/libremap/plugins/location.lua' 	no file '/usr/share/lua/luci/libremap/plugins/location/init.lua' 	no file '/usr/lib/lua/luci/libremap/plugins/location.lua' 	no file '/usr/lib/lua/luci/libremap/plugins/location/init.lua' 	no file './luci/libremap/plugins/location.so' 	no file '/usr/lib/lua/luci/libremap/plugins/location.so' 	no file '/usr/lib/lua/loadall.so' 	no file './luci.so' 	no file '/usr/lib/lua/luci.so' 	no file '/usr/lib/lua/loadall.so'
user.warn libremap-agent[2792]: unable to load plugin "babeld"; module 'luci.libremap.plugins.babeld' not found: 	no field package.preload['luci.libremap.plugins.babeld'] 	no file './luci/libremap/plugins/babeld.lua' 	no file '/usr/share/lua/luci/libremap/plugins/babeld.lua' 	no file '/usr/share/lua/luci/libremap/plugins/babeld/init.lua' 	no file '/usr/lib/lua/luci/libremap/plugins/babeld.lua' 	no file '/usr/lib/lua/luci/libremap/plugins/babeld/init.lua' 	no file './luci/libremap/plugins/babeld.so' 	no file '/usr/lib/lua/luci/libremap/plugins/babeld.so' 	no file '/usr/lib/lua/loadall.so' 	no file './luci.so' 	no file '/usr/lib/lua/luci.so' 	no file '/usr/lib/lua/loadall.so'
```

The `location` plugin is preset to enabled here:
https://github.com/libremesh/libremap-agent/blob/fbca90abd1e98b732f58942dc46653c567b6ada8/libremap-agent/files/etc/uci-defaults/80_libremap-agent#L25
and causes the error as the plugin is not present because it is a dependency only of [lime-map-agent](https://github.com/libremesh/lime-packages/tree/master/packages/lime-map-agent) which is not recommended in the [compilation guide](https://libremesh.org/development.html).
It should either be added as a dependency of libremap-agent or be preset to disabled (and added to the compilation guide if we want to recommend it).

[...]

PR #797, opened 2020-11-08
babeld: do not forcefully enable libremap plugin

ilario:
See issue description in #788.
In #788 two issues are described, the second one is addressed here.

### DIFF
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -83,12 +83,6 @@ function babeld.configure(args)
 
 	uci:save("babeld")
 
-
-	uci:set("libremap", "babeld", "plugin")
-	uci:set("libremap", "babeld", "enabled", "true")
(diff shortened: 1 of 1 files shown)


spiccinini (approved):

G10h4ck:
So this just leaves the libremap plugin un-configured! Maybe what you wants is to not configure it if not installed?

ilario:
> So this just leaves the libremap plugin un-configured!

Citing from the reference issue #788 description:

> The babeld plugin is preset to disabled

See https://github.com/libremesh/libremap-agent/blob/fbca90abd1e98b732f58942dc46653c567b6ada8/libremap-agent/files/etc/uci-defaults/80_libremap-agent#L39-L40

> and libremap-agent enables it if if detects that is actually installed here:
https://github.com/libremesh/libremap-agent/blob/fbca90abd1e98b732f58942dc46653c567b6ada8/libremap-agent/files/etc/uci-defaults/80_libremap-agent#L63-L69

So, by default it is configured as not-enabled.

G10h4ck (approved):

PR #799, opened 2020-11-09
shared-state and ubus-lime-metrics disable httpS redirection in uhttpd

ilario:
In https://github.com/libremesh/lime-web/pull/118 I wrote

> And I did not test if shared-state is affected.

and turns out that it is affected indeed :/
Adding the `--no-check-certificate` fixes and does not generate a problem when having nodes without httpS.

Another point where uclient-fetch (wget) is used towards nodes is in ubus-lime-metrics, so I added the same `--no-check-certificate` option also there.


dangowrt:
Why are we using https here in first place then? Isn't that just a waste of CPU cycles, as no additional security is gained in that way?
If the reason for that is that uhttpd redirects all http requests to https by default, then we should probably rather change that instead imho.

ilario:
> Why are we using https here in first place then? Isn't that just a waste of CPU cycles, as no additional security is gained in that way?

Yep, in this case no security is gained.
The httpS was enabled for accessing the web interface and was enabled simply specifying which packages to install in the development page of the website, see libremesh/lime-web#118 and libremesh/lime-packages#613 

> If the reason for that is that uhttpd redirects all http requests to https by default, then we should probably rather change that instead imho.

Yes, that's what happens: all http connections gets redirected to httpS.
I agree with you that would make much more sense to not have shared-state waste CPU for httpS.

Maybe we should create exceptions in the redirect rules for `/cgi-bin/hostname` and for `/cgi-bin/shared-state/*`?

dangowrt:
> Maybe we should create exceptions in the redirect rules for `/cgi-bin/hostname` and for `/cgi-bin/shared-state/*`?

That's beyond uhttpd's capabilities. What we can do (but it's also a bit of a waste) is having an additional uhttpd instance which doesn't do https ever and use that for shared-state stuff.

ilario:
Or we can disable the automatic redirection to httpS completely. So that if a user is aware of the existence of httpS he will use it, otherwise it will be plain http.

Additionally, if it can be done I don't know, we could have the redirection to httpS inside lime-app, for example here:
https://github.com/libremesh/lime-packages/blob/master/packages/lime-app/files/www/lime_app_index.html

dangowrt:
```
uci set uhttpd.main.redirect_https='0'
uci commit uhttpd
```
seems like the best thing to do, better than using costly but useless httpd handshakes for something like shared-state...

ilario:
> ```
> uci set uhttpd.main.redirect_https='0'
> uci commit uhttpd
> ```
> 
> seems like the best thing to do, better than using costly but useless httpd handshakes for something like shared-state...

Ok, sounds good to me, I'll modify the PR adding these commands as a uci-defaults script inside `shared-state` and in `ubus-lime-metrics` packages (unless you see a better way to do this). Then I'll add a phrase in the development page indicating that the httpS has to be specified manually (no redirect).

ilario:
As suggested by @dangowrt I disabled uhttpd `redirect_https` which [by default is enabled](https://github.com/openwrt/openwrt/blob/0b31713c856a0d8c6fe2961113a9317f7c3659dc/package/network/services/uhttpd/files/uhttpd.config#L13).

In order to do this, I added an uci-defaults script to each of the packages using wget/uclient-fetch to get information from other nodes.

> Additionally, if it can be done I don't know, we could have the redirection to httpS inside lime-app, for example here:
https://github.com/libremesh/lime-packages/blob/master/packages/lime-app/files/www/lime_app_index.html

We could try this using JS: https://stackoverflow.com/questions/11401897/get-the-current-domain-name-with-javascript-not-the-path-etc (I don't know how to do this)

Let me know your opinions on this last point, in case we don't want to do this I'll specify in the development page that httpS has to be explicitly requested by the users.

dangowrt (approved):

spiccinini:
Another way to do it is to add this config to lime-defaults. As uhttpd is a somewhat default package I would do that this way.
```
        config generic_uci_config uhttpd_https
            list uci_set "uhttpd.main.redirect_https=0"
```

ilario:
> Another way to do it is to add this config to lime-defaults.

Yes, that would also be a valid solution.
The pro of adding this just where needed (shared-state and ubus-lime-metric) is that if one user is not using those packages (rather unlikely actually...), the default behaviour of uhttpd will be in place.

The pro of your suggestion would be that a user can re-enable it. But re-enabling it makes sense just if neither shared-state nor ubus-lime-metric are installed, otherwise the user will break things.

So I see these cases:

|  | With shared-state and ubus-lime-metric | Without shared-state and ubus-lime-metric (unusual) |
| ----------- | ----------- | ----------- |
| With hardcoded conf in the packages | No uttpd redirect, never. | By default redirect activated. Configurable via uci. |
| Configurable in lime-system  | By default redirect disactivated. Configurable via lime-node (breaking things!).       | By default redirect disactivated. Configurable via lime-node. |

I think that it's better to leave (at least) the hardcoded setting in the packages. Otherwise we're offering the users an easy way to break things in change of a easier configuration for a small minority of users.

A third option would be to hardcode it into lime-system (not in lime-defaults).

G10h4ck:
Please add also the snipped proposed by @spiccinini so every-time `lime-config` run it check for the option being in place

ilario:
> Please add also the snippet proposed by @spiccinini so every-time `lime-config` run it check for the option being in place

If we want to enforce more strictly that option I would find other ways. For example, I can add a check in shared-state throwing an error or warning if the redirect is not disabled.

I just thought yet another way to solve this (a mix-up of the previously proposed solutions), maybe the most satisfactory so far:

* use @spiccinini configuration code in lime-defaults;
* add the `--no-check-certificate` option to wget inside shared-state and ubus-lime-metric (as it was initially proposed).

ilario:
I pushed the alternative in which lime-defaults disables the redirect_https but nothing explodes if a user enables it (I did not test yet).

ilario:
I just tested and it works:


without touching anything the redirection is disabled and shared-state works;

manually enabling the redirection setting
```
config generic_uci_config uhttpd_https
        list uci_set 'uhttpd.main.redirect_https=1'
```
 in `/etc/config/lime-node` re-enables the redirection but shared-state still works.

@spiccinini can we merge this?

spiccinini (approved):

PR #804, opened 2020-11-11
ubus-lime-location do not depend on libremap-agent

ilario:
Followup from my [comment here](https://github.com/libremesh/lime-packages/issues/788#issuecomment-725038277) on #788:

>> in a cascade of dependencies (`libremap-agent` => `luci-lib-libremap` => `luci-base`) it selects `luci-base` and increases a lot the size of the image (`luci-base` is also selected in a cascade by `check-date-http` package, which is suggested as optional [in the development page](https://libremesh.org/development.html#compiling_libremesh_from_source_code)).
>> and get rid of all these issues altogether. Currently, libremap-agent is selected by ubus-lime-location which is selected by lime-app.

> I tried to understand how ubus-lime-location works and seems that it does not require libremap-agent in order to work. The selection of libremap-agent can be suggested to users willing to update the network condition on a public map server, and can be removed as a hard dependency from ubus-lime-location.
> Is this correct?
> This would have the bonus to allow the generation of images with lime-app without LuCI (less dependencies smaller image!).

### DIFF
--- a/packages/ubus-lime-location/Makefile
+++ b/packages/ubus-lime-location/Makefile
@@ -13,7 +13,7 @@ define Package/$(PKG_NAME)
   MAINTAINER:=Marcos Gutierrez <gmarcos87@gmail.com>
   SUBMENU:=3. Applications
   TITLE:=Libremap ubus status module
-  DEPENDS:= +lua +libubox-lua +libubus-lua +libuci +libremap-agent
+  DEPENDS:= +lua +libubox-lua +libubus-lua +libuci
   PKGARCH:=all
 endef
 


ilario:
@nicopace to me it looks like ubus-lime-location should depend on shared-state-nodes_and_links, see b722f921dca901c508afc3fb080e7f7e674c143b

If we agree on both changes, I'll add the shared-state-nodes_and_links dependency in this PR (or in another, just tell me).

nicopace:
I would say yes, but wait for @spiccinini as I am not that active in the development work lately.

ilario:
Closing this as #805 is much better, thanks @spiccinini!
I'll open another one for the addition of shared-state-nodes_and_links dependency in ubus-lime-location.

ilario:
>> @nicopace to me it looks like ubus-lime-location should depend on shared-state-nodes_and_links

> I would say yes, but wait for @spiccinini as I am not that active in the development work lately.

I just realised that shared-state-nodes_and_links already depends on ubus-lime-location (see #754) so the reverse cannot happen (OpenWrt [does not like circular dependencies](https://openwrt.org/docs/guide-developer/dependencies#caveats)).

PR #805, opened 2020-11-13
location: refactor to not depend on libremap-agent

spiccinini:
Provides the location config from the ubus-lime-location itelf. Maybe we can move this config  into its own package?

Also this migrates the config from libremap if available into a generic uci config.


packages/ubus-lime-location/files/etc/uci-defaults/93-migrate_libremap_cfg
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/lua
+
```
G10h4ck, line 2: Should we put some kind of time to live on this file, and mark it for removal for some 2022 release?

G10h4ck (approved):

packages/ubus-lime-location/files/etc/uci-defaults/93-migrate_libremap_cfg
ilario: Like the concept. We should open a permanent issue where we collect obsolete code which should be removed with LibreMesh 21.xx release.

ilario: @G10h4ck please consider listing this file in #809.

ilario:
I just tested quickly:

* set the location with a firmware previous to this PR;
* upgraded using lime-sysupgrade to a firmware with this PR;
* check if the location has been preserved.

And the location is not being preserved.
This happens because the [OpenWrt's init boot script tries to run the stuff in /etc/uci-defaults using the current shell](https://github.com/openwrt/openwrt/blob/openwrt-19.07/package/base-files/files/etc/init.d/boot#L15), which is not Lua. So it fails and the migration does not happen.

The rest of the PR works :D

spiccinini:
> I just tested quickly:
> 
>     * set the location with a firmware previous to this PR;
> 
>     * upgraded using lime-sysupgrade to a firmware with this PR;
> 
>     * check if the location has been preserved.
> 
> 
> And the location is not being preserved.
> This happens because the [OpenWrt's init boot script tries to run the stuff in /etc/uci-defaults using the current shell](https://github.com/openwrt/openwrt/blob/openwrt-19.07/package/base-files/files/etc/init.d/boot#L15), which is not Lua. So it fails and the migration does not happen.
> 
> The rest of the PR works :D

Thanks Ilario for testing it! Will fix it

ilario:
I think that the migration Lua script can be removed from uci-defaults, moved to /usr/bin and offered for manual execution in the documentation of the new release.

spiccinini:
> I think that the migration Lua script can be removed from uci-defaults, moved to /usr/bin and offered for manual execution in the documentation of the new release.

Based in your suggestion I moved the script into /usr/bin but I maintained the uci-defaults calling the script. What do you think? I am not very convinced. I think the ideal option would be to be able to run the migration previous to the firmware upgrade (as part of the upgrade process).

ilario:
Just tested the latest changes, the migration script works amazingly. I'm going to merge.

ilario (approved):

PR #806, opened 2020-11-13
fbw: use non absolute wget path

### LINKED ISSUE #798: FBW: hardcoded wget absolute path is wrong on OpenWrt 19.07
In OpenWrt 18.06 `wget` link pointing to uclient-fetch was located in `/bin/`, [see here](https://github.com/openwrt/openwrt/blob/openwrt-18.06/package/libs/uclient/Makefile#L46), but in OpenWrt 19.07 this was moved to `/usr/bin/`, [see here](https://github.com/openwrt/openwrt/blob/openwrt-19.07/package/libs/uclient/Makefile#L35).

For this reason, two wget commands are failing inside FirstBootWizard:
https://github.com/libremesh/lime-packages/blob/31e6a58d36ca637fab640f423df42ce7363a2677/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua#L160

https://github.com/libremesh/lime-packages/blob/31e6a58d36ca637fab640f423df42ce7363a2677/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua#L166

spiccinini:
wget path changed from 18.06 to 19.07 so this is to be more compatible.
Current uses of firstbootwizard (init.d and ubus daemon) have the PATH
env variable correctly populated so this should not break anything.

Fixes #798

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -157,13 +157,13 @@ end
 function fbw.fetch_config(data)
     fbw.log('Fetch config from '.. json.stringify(data))
     local host = data.host
-    local hostname = utils.execute("/bin/wget http://["..data.host.."]/cgi-bin/hostname -qO - "):gsub("\n", "")
+    local hostname = utils.execute("wget http://["..data.host.."]/cgi-bin/hostname -qO - "):gsub("\n", "")
     fbw.log('Hostname found: '.. hostname)
     if (hostname == '') then hostname = host end
     local signal = data.signal
     local ssid = data.ssid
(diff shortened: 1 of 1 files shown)


packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -157,13 +157,13 @@ end
 function fbw.fetch_config(data)
     fbw.log('Fetch config from '.. json.stringify(data))
     local host = data.host
```
G10h4ck, line 159: either we avoid to define this local variable or later we use `host` instead of `data.host`

G10h4ck (approved):

ilario (approved):
Tested on top of Openwrt 19.07.4. Using this patch and manually fixing #807, FBW works.

PR #808, opened 2020-11-14
makefiles: depend on libuci-lua instead of libuci

spiccinini:
Depending on libuci creates a binary dependency tied to a specific compile time defined version of libuci. As these packages use libuci through lua it is better to depend on the correct package that also makes not depending on any compiled stuff.

This fixes the following error:
```
root@LiMe-1efa02:~# opkg install ubus-lime-fbw
Installing ubus-lime-fbw (2020-11-14-1605375181) to root...
Downloading http://feed.libremesh.org/master/ubus-lime-fbw_2020-11-14-1605375181_all.ipk
Collected errors:
 * satisfy_dependencies_for: Cannot satisfy the following dependencies for ubus-lime-fbw:
 * 	libuci20130104
 * opkg_install_cmd: Cannot install package ubus-lime-fbw.
```


ilario:
Did you test if this fixes?

`libuci-lua` depends on `libuci` (see [here](https://github.com/openwrt/openwrt/blob/openwrt-19.07/package/system/uci/Makefile#L51)) so that the problem should be still there.

Additionally, the libuci20130104 version is available on OpenWrt repositories, for example here:
https://downloads.openwrt.org/releases/packages-19.07/mipsel_24kc/base/libuci20130104_2019-09-01-415f9e48-4_mipsel_24kc.ipk

G10h4ck (approved):

spiccinini:
Yes it fixes the issue. The other ubus-lime packages depend on this dependency

ilario:
> Yes it fixes the issue.

Mmh it still does not make sense to me, if the opkg packages list is updated there should be no problem in obtaining libuci20130104 from OpenWrt repositories.

> The other ubus-lime packages depend on this dependency

lubuci-lua is not a dependency of ubus-lime-* but is a dependency of many other packages (e.g. lime-proto-*) so it's ok to merge.
I approve this but it still makes no sense to me.

ilario (approved):

spiccinini:
Closing as this was introduced in #815

ilario:
> Closing as this was introduced in #815

So is this still needed or not?
In #815 that dependency was introduced for ubus-lime-fbw and first-boot-wizard.
In this PR that dependency would be added to ubus-lime-groundrouting, ubus-lime-location, ubus-lime-utils.

Issue #280, opened 2018-01-16
Use OpenWrt firewall (drop all connections incoming from WAN) or add a fail2ban-like functionality to dropbear and uhttpd


nicopace:
Many routers are accesible from outside, so they need to be considered as production servers.
Looking at the logs of dropbear of one of our nodes, I saw that a public IP tried to access the router many many times, and the dropbear allowed it to try all the times (none passed).

One thing that can be done is to add fail2ban in order to reduce the chance for bots to access or to do a DOS to the nodes.

fail2ban is implemented in python, so it is a no-go, but on openwrt they are suggesting to use logtrigger:
https://forum.openwrt.org/viewtopic.php?id=27955

ilario:
+1
More alternatives:
https://github.com/robzr/bearDropper
https://sourceforge.net/projects/mf2b

nicopace:
I liked the logtrigger idea, because it will allow us not only to react to fail2ban situations, but also others like the ones we are doing workarounds (cause some of them can be detected on logs):

https://github.com/enryIT/logtrigger

aparcar:
@nicopace have you tried the tool? It looks nice but 6 years without any development?
We could create a PR to add it to the official OpenWrt packages repo?

nicopace:
We have been working with @gmarcos87 and @nicoechaniz to filter the current logs of the system in order to clear them and be able to capture those valuable.
We haven't published this (yet), but it is based on rsyslogd's liblognorm: http://www.liblognorm.com/
We also implemented openwrt support for the library... it is still pending to use this tool to take actions based on the logs within the router... 

Right now it is only being used to collect the output of the local routers within the networks in argentina, and to filter those logs afterwards in a raspberry pi.

Will have more news soon :)

ilario:
This would be very good to have, is there any update?
The work with liblognorm in LibreMesh can be found [here](https://github.com/libremesh/lime-log-review).

nicopace:
The current implementation of [remotefilteredlog](https://github.com/libremesh/lime-log-review/tree/master/packages/remotefilteredlog) sends ubus messages when a message that wanted to be alerted happens.

Also, doing some skim over https://github.com/enryIT/logtrigger looks old but healthy.

ilario:
This issue is closely related to this one reported in 2016:
https://lists.libremesh.org/pipermail/lime-dev/2016-December/000821.html

Access to web interface is also dangerous because of #640.

We could:

[...]

PR #812, opened 2020-11-23
shared-state-dnsmasq_servers: new package

altergui:
Signed-off-by: Gui Iribarren <gui@altermundi.net>

* a uci-defaults script points dnsmasq to use `/var/shared-state/dnsmasq_servers` as `servers-file`
* each node then publishes through shared-state their own dhcp.dnsmasq.domain and anygw ipv6
* each node populates the local `/var/shared-state/dnsmasq_servers` with all the domains and ipv6s from the other clouds.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -153,6 +153,8 @@ function network.setup_dns()
 			uci:set("dhcp", s[".name"], "local", "/"..cloudDomain.."/")
 			uci:set("dhcp", s[".name"], "expandhosts", "1")
 			uci:set("dhcp", s[".name"], "domainneeded", "1")
+			--! allow queries from non-local ips (i.e. from other clouds)
+			uci:set("dhcp", s[".name"], "localservice", "0")
 			uci:set("dhcp", s[".name"], "server", resolvers)
 		end
 	)
--- a/packages/shared-state-dnsmasq_servers/files/usr/bin/shared-state-generate_dnsmasq_servers
+++ b/packages/shared-state-dnsmasq_servers/files/usr/bin/shared-state-generate_dnsmasq_servers
@@ -1,0 +1,46 @@
+#!/usr/bin/lua
+
+--! LibreMesh
+--! Copyright (C) 2020 Gui iribarren <gui@altermundi.net>
+--!
+--! This program is free software: you can redistribute it and/or modify
+--! it under the terms of the GNU Affero General Public License as
+--! published by the Free Software Foundation, either version 3 of the
+--! License, or (at your option) any later version.
+--!
+--! This program is distributed in the hope that it will be useful,
+--! but WITHOUT ANY WARRANTY; without even the implied warranty of
+--! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+--! GNU Affero General Public License for more details.
+--!
+--! You should have received a copy of the GNU Affero General Public License
+--! along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+local JSON = require("luci.jsonc")
+
+local uci = require("uci"):cursor()
+
+uci:foreach("dhcp", "dnsmasq",
+  function(s)
+    localDomain = uci:get("dhcp", s[".name"], "local"):gsub("/","")
+  end
+)
+
+local localHostname = io.input("/proc/sys/kernel/hostname"):read("*line")
+
+local content = {}
+
+--! example key="cloud1.mesh", example value.data="fd0d:fe46:8ce8::ab:cd00"
+for key,value in pairs(
+ JSON.parse(io.stdin:read("*all")) ) do
+ if value.data and value.author ~= localHostname and key ~= localDomain then
+   table.insert(content, "server=/" .. key .. "/" .. value.data)
+ end
+end
+
+local outputFile = io.open("/var/shared-state/dnsmasq_servers", "w")
+if outputFile then
+	outputFile:write(table.concat(content,"\n").."\n")
+	outputFile:close()
+	os.execute("killall -HUP dnsmasq 2>/dev/null")
+end
--- a/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
+++ b/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
@@ -100,6 +100,7 @@ function anygw.configure(args)
 	uci:foreach("dhcp", "dnsmasq",
 		function(s)
 			uci:set("dhcp", s[".name"], "add_local_fqdn", "0")
+			uci:set("dhcp", s[".name"], "add_local_hostname", "0")
 		end
 	)
 
--- a/packages/shared-state-dnsmasq_hosts/files/usr/bin/shared-state-generate_dnsmasq_hosts
+++ b/packages/shared-state-dnsmasq_hosts/files/usr/bin/shared-state-generate_dnsmasq_hosts
@@ -19,13 +19,10 @@
 local JSON = require("luci.jsonc")
 
 local outputTable = {}
-local localHostname = io.input("/proc/sys/kernel/hostname"):read("*line")
 
 for key,value in pairs(
 	JSON.parse(io.stdin:read("*all")) ) do
-	if value.data and value.author ~= localHostname then
 		table.insert(outputTable, key)
-	end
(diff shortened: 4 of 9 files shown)


altergui:
this is the start of getting DNS "right"
some notes to self:
* `shared-state-dnsmasq_hosts` and `shared-state-dnsmasq_leases` have no concept of "cloud domain", and they should. both dhcp leases and dns hosts should be local to the cloud, and not distributed across clouds.

packages/shared-state-dnsmasq_servers/files/usr/bin/shared-state-generate_dnsmasq_servers
```diff
@@ -0,0 +1,47 @@
...
+--! example key="cloud1.mesh", example value.data="fd0d:fe46:8ce8::ab:cd00"
+for key,value in pairs(
+ JSON.parse(io.stdin:read("*all")) ) do
+ if value.data and value.author ~= localHostname and key ~= localDomain then
+   table.insert(content, "/" .. key .. "/" .. value.data)
+ end
+end
+
+local outputFile = io.open("/var/shared-state/dnsmasq_servers", "w")
+print(outputFile)
```
G10h4ck, line 42: This print is just some residual debugging code?

altergui: correct, thanks for spotting it!
i amended the commit and force pushed a clean tree

altergui:
now it works nicely. tested this on a virtual testbed with 12 nodes split among 4 different clouds. i spotted some quirks and added some fixes in various files.

altergui:
wishlist: implement `rev-servers` as well
right now, each node publishes it's own domain and ipv6
and the result is the following
```
server=/cloudA.test/fdeb:75d8:5c6d::1
server=/cloudD.test/fd30:7030:8ee3::1
server=/cloudB.test/fd54:721a:3043::1
```
ideally, each node would also publish it's ipv6 prefix, so the result would be
```
server=/cloudA.test/fdeb:75d8:5c6d::1
rev-server=fdeb:75d8:5c6d::/64,fdeb:75d8:5c6d::1
server=/cloudD.test/fd30:7030:8ee3::1
rev-server=fd30:7030:8ee3::/64,fd30:7030:8ee3::1
server=/cloudB.test/fd54:721a:3043::1
rev-server=fd54:721a:3043::/64,fd54:721a:3043::1
```

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -153,6 +153,8 @@ function network.setup_dns()
 			uci:set("dhcp", s[".name"], "local", "/"..cloudDomain.."/")
 			uci:set("dhcp", s[".name"], "expandhosts", "1")
 			uci:set("dhcp", s[".name"], "domainneeded", "1")
+			--! allow queries from non-local ips (i.e. from other clouds)
+			uci:set("dhcp", s[".name"], "localservice", "0")
```
G10h4ck, line 157: Does this mean that our dnsmasq ill attempt to answhere any type of queries or only the ones relatives to the domain of the cloud?  The latter is ok, the first could expose our dnsmasq to become a DNS botnet considering a community network that have public IPv6

G10h4ck (approved):

packages/lime-system/files/usr/lib/lua/lime/network.lua
altergui: any type of queries (including recursive), as far as i understand.
as long as `fw3` is installed and running (which is, by default), it will block any incoming queries over the public internet, both on ipv4 and ipv6. maybe `lime-firewall` does this as well by default? (it really should)
ideally, instead of disabling `local-service`, we would enable `auth-server` in dnsmasq conf. but the standard `dnsmasq` package shipped with openwrt does not compile in `auth` support. (`dnsmasq-full` does, with a myriad other features)

G10h4ck: How does `fw3` distinguish from a public IPv6 from the internet and a public IPv6 from another cloud? Maybe we should create a package `dnsmasq-authoritative` package with just the options we need, and eventually upstream it?

altergui: `fw3` AFAIU distinguish from incoming interface. if it's WAN port, it blocks incoming traffic.

creating `dnsmasq-auth` sounds great in principle... in my experience, it's easy to somehow during menuconfig end up with dnsmasq (vanilla) installed. and quite frustrating to debug why `auth-server` is not working

altergui: we had this issue with `dnsmasq-dhcpv6` (instead of `dnsmasq`)

G10h4ck: Apparently that have been solved and now if the wrong dnsmasq is selected it fails at compile time

G10h4ck: > `fw3` AFAIU distinguish from incoming interface. if it's WAN port, it blocks incoming traffic.

Well that's not good as we support WAN-WAN connections as a way to connect different clouds, thought I think we can eventually change that, with the current config system that would not be that problematic IMHO

ilario: > as we support WAN-WAN connections as a way to connect different clouds

We never did, as far as I know the right way to connect different clouds is a lan-lan connection where the two clouds have different batman-adv vlan id.
People where using wan-wan because batman-adv was not running there, I suppose, but with the new release batman-adv runs also on wan so using wan-wan connection makes even less sense than before.

ilario: > as long as fw3 is installed and running (which is, by default)

fw3 is not installed by default, see the instructions in the [development page](https://libremesh.org/development.html).
For more discussion about whether selecting fw3 or not see #280

G10h4ck: > We never did, as far as I know the right way to connect different clouds is a lan-lan connection where the two clouds have different batman-adv vlan id.
> People where using wan-wan because batman-adv was not running there, I suppose, but with the new release batman-adv runs also on wan so using wan-wan connection makes even less sense than before.

Nope you are missing a piece about this, if you connect two cloud LAN-LAN it will become a huge mess, all the MAC addresses of the second cloud would be seen as batman-adv clients in the second cloud, telling batman-adv to send hello packets on LAN doesn't help in this case even if it helps a lot in other.

G10h4ck: > > `fw3` AFAIU distinguish from incoming interface. if it's WAN port, it blocks incoming traffic.
> 
> Well that's not good as we support WAN-WAN connections as a way to connect different clouds, thought I think we can eventually change that, with the current config system that would not be that problematic IMHO

Also how does `fw3` distinguish between a request coming to resolve something like `myraspberry.mycommunity.info` which is legitimate even from WAN and one that requests `www.mybllaaaaaaaadomain.com`. Using `fw3` or any other firewall to do this will break that legitimate use case which is one of the purposes we are implementing this.

ilario: > > We never did, as far as I know the right way to connect different clouds is a lan-lan connection where the two clouds have different batman-adv vlan id.
> > People where using wan-wan because batman-adv was not running there, I suppose, but with the new release batman-adv runs also on wan so using wan-wan connection makes even less sense than before.
> 
> Nope you are missing a piece about this, if you connect two cloud LAN-LAN it will become a huge mess, all the MAC addresses of the second cloud would be seen as batman-adv clients in the second cloud, telling batman-adv to send hello packets on LAN doesn't help in this case even if it helps a lot in other.

Sorry but I still don't understand.
From my point of view, the fact that Batman-adv runs on VLAN both on LAN than on WAN makes them identical for Baman-adv.
And I repeat: Batman-adv is also running on WAN (with VLAN, identical as on LAN).

ilario:
Sorry for not noticing this before: can you update the Makefile to use https://github.com/libremesh/lime-packages/blob/master/libremesh.mk or just change the "category" from LiMe to LibreMesh?

ilario:
Re-reading the discussion here after years what I understand is:
The pull request is ready for merging but requires a firewall to be in place for filtering some stuff coming from the internet, so this depends on #280 and https://github.com/libremesh/libremesh.github.io/pull/139

PR #813, opened 2020-11-24
Qemu 12 nodes in 4 different clouds

altergui:
started working in a virtualized mesh cloud

### DIFF
--- a/tools/qemu_cloud_start
+++ b/tools/qemu_cloud_start
@@ -1,7 +1,0 @@
-#!/bin/bash
-for id in `seq 0 10` ; do
-  ( sleep ${id}s && ./tools/qemu_dev_start --node-id $id $@ ) &
-done
-
-echo "done, now you can launch"
-echo 'clusterssh fe80::5000:ff:feab:c2{00..10}%lime_br0'
--- a/tools/qemu_cloud_start
+++ b/tools/qemu_cloud_start
@@ -1,0 +1,7 @@
+#!/bin/bash
+for id in `seq 0 10` ; do
+  ( sleep ${id}s && ./tools/qemu_dev_start --node-id $id $@ ) &
+done
+
+echo "done, now you can launch"
+echo 'clusterssh fe80::5000:ff:feab:c2{00..10}%lime_br0'
--- a/tools/qemu_dev_start
+++ b/tools/qemu_dev_start
@@ -215,10 +215,8 @@ if [ "$_arg_libremesh_workdir" ]; then
     done
 fi
 
-# build the cdpio. Don't know how to do it without changing the directory
-cd /tmp/lime_rootfs_${NODE_ID}
-find . | cpio --quiet -o -H newc  > /tmp/lime_rootfs_${NODE_ID}.cpio
-cd - > /dev/null
+# build the cpio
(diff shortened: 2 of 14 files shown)


spiccinini:
Looks good! I will test it :)

tools/qemu_cloud_start
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+for id in `seq 0 10` ; do
+  #./tools/qemu_dev_start --no-serial --node-id $id $@ &
```
spiccinini, line 3: This is commented on purpose?

spiccinini:
What is the idea behind the ETH2 interface?

tools/qemu_cloud_start
altergui: nope. sorry about that, anyway refactored everything into an ansible-playbook, much more powerful

altergui:
after major work, made an ansible playbook that reads a YAML describing a complex set of nodes distributed accross different clouds with links connecting them.

G10h4ck (approved):

tools/qemu_dev_start
```diff
@@ -62,9 +65,12 @@ print_help()
 	printf '\t%s\n' "<rootfs>: Path to a x86-64-generic-rootfs.tar.gz"
```
spiccinini, line 65: The usage line (--help) has to be updated also.  
We have been using argbash https://argbash.io/generate to maintain the options of the script (and helper line, etc), can you please use it? (you can copy from the line 0 until # ARGBASH_GO() into the argbash textbox and then copy the result back to the script).

altergui: thanks! was unaware of that. rebased and force-pushed my branch, using argbash

altergui:
with this rebase i dropped the `--no-serial` argument (it's not really important) and used argbash to handle arguments, as indicated
https://github.com/libremesh/lime-packages/compare/1143e39753d7a43953366e17ca79254223db1bc9..5407071393ed5c3868d377e0973e2b4d3b502804

spiccinini:
I tested the non cloud operation and it is working :)

At first it the `--enable-wan` options wasn't working and It took me some time to discover that as the MAC configuration  changed the dnsmasq lease time was preventing dnsmasq to offer an ip as the range is only one IP .... Anyway, everything good, I've just deleted the lease file (`/var/lib/misc/dnsmasq.leases` in my distro).

spiccinini (approved):

spiccinini:
One thing before merging. Which license the python files have?

tools/qemu_cloud_start
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
```
spiccinini, line 1: This and qemu_cloud_stop are no longer needed?

spiccinini:
@altergui please can you answer the remaining questions? I want to merge this :)

altergui:
> One thing before merging. Which copyright and license the python files have?

good catch! no license whatsoever. wrote an email to the author asking for clarification.
i think  (based on other code by same author) it's safe to assume the author intended to publish under MIT license and just ommitted it.
we can wait a bit until a proper author statement, and if that doesn't happen... no idea

tools/qemu_cloud_start
altergui: true. i just removed them

altergui:
success. MIT license it is.
https://gist.github.com/upa/94f82e00dfffbb25d896/revisions

spiccinini:
Thank gui!

PR #814, opened 2020-11-26
fbw: share lime-community as cgi-bin

### LINKED ISSUE #807: FBW: 403 error when fetching lime-community
Originally, the `/etc/config/lime-community` file has `-rw-r--r--` permissions.
When creating a new network, the `/etc/config/lime-community` file has read permissions only for root: `-rw-------`.
This causes the failure in wget (403: Forbidden) on the other nodes when trying to download the lime-community file.

Before creating a new network:

```
# ls -l /etc/config/lime*
-rw-r--r--    1 root     root           196 Sep  6 16:19 /etc/config/lime-app
-rw-------    1 root     root          2300 Nov 10 12:19 /etc/config/lime-autogen
-rw-r--r--    1 root     root           209 Sep  6 16:19 /etc/config/lime-community
-rw-r--r--    1 root     root          2377 Sep  6 16:19 /etc/config/lime-defaults
-rw-r--r--    1 root     root           374 Nov 10 12:19 /etc/config/lime-node
```

after creating a new network using FirstBootWizard:

```
# ls -l /etc/config/lime*
-rw-r--r--    1 root     root           196 Sep  6 16:19 /etc/config/lime-app
-rw-------    1 root     root          2199 Nov 10 13:08 /etc/config/lime-autogen
-rw-------    1 root     root           237 Nov 10 13:08 /etc/config/lime-community
-rw-r--r--    1 root     root          2377 Sep  6 16:19 /etc/config/lime-defaults
-rw-------    1 root     root           134 Nov 10 13:08 /etc/config/lime-node
```

I am testing on top of OpenWrt 19.07.4

spiccinini:
Sharing the lime-community file as a symlink needs read permissions and uci changes these file permissions in 19.07 when uci commit is run.
Sharing the file as a cgi-bin prevents this problems and is also more hackable. 

Fixes #807

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -157,13 +157,19 @@ end
 function fbw.fetch_config(data)
     fbw.log('Fetch config from '.. json.stringify(data))
     local host = data.host
+
     local hostname = utils.execute("wget http://["..data.host.."]/cgi-bin/hostname -qO - "):gsub("\n", "")
     fbw.log('Hostname found: '.. hostname)
     if (hostname == '') then hostname = host end
     local signal = data.signal
     local ssid = data.ssid
     local filename = fbw.WORKDIR .. fbw.HOST_CONFIG_PREFIX .. hostname
-    utils.execute("wget http://[" .. data.host .. "]/lime-community -O " .. filename)
+
+    utils.execute("wget http://[" .. data.host .. "]/cgi-bin/lime/lime-community -O " .. filename)
+    if not utils.file_exists(filename) then
+        -- For backwards compatibility
+        utils.execute("wget http://[" .. data.host .. "]/lime-community -O " .. filename)
+    end
 
     -- Remove lime-community files that are not yet configured.
     -- For this we asume that no ap_ssid options equals not configured.
--- a/packages/first-boot-wizard/files/www/cgi-bin/lime/lime-community
+++ b/packages/first-boot-wizard/files/www/cgi-bin/lime/lime-community
@@ -1,0 +1,6 @@
+#!/bin/sh
+
+echo "Content-Type: text/plain"
+echo
+
+cat /etc/config/lime-community
--- a/packages/first-boot-wizard/files/www/lime-community
+++ b/packages/first-boot-wizard/files/www/lime-community
@@ -1,1 +1,0 @@
-/etc/config/lime-community


ilario:
Thanks @spiccinini! I think this is a valid way to fix. I will test as soon as possible :)
Just one reminder: both this PR and #799 modify the same line so we'll have to fix the conflict in one of the two.

nicopace (approved):

ilario (approved):
Tested, it works :D

ilario:
One question (a bit late): the exposed lime-community file exposes the md5crypt of the community password and the hash is a bit weak.

In the lime-example.txt stronger alternatives are mentioned, this is great, but these are not supported by FBW and neither by OpenWrt! (are not available by default OpenWrt 19.07 (but can be activated) and in the case they are enabled, doubts are present on the higher security due to lack of salting: https://forum.openwrt.org/t/passwd-a-sha256-sha512-less-secure-than-default/13117/ )

We should warn users on the risk to use important passwords there.

spiccinini:
> One question (a bit late): the exposed lime-community file exposes the md5crypt of the community password and the hash is a bit weak.
> 
> In the lime-example.txt stronger alternatives are mentioned, this is great, but these are not supported by FBW and neither by OpenWrt! (are not available by default OpenWrt 19.07 (but can be activated) and in the case they are enabled, doubts are present on the higher security due to lack of salting: https://forum.openwrt.org/t/passwd-a-sha256-sha512-less-secure-than-default/13117/ )
> 
> We should warn users on the risk to use important passwords there.

I would say that md5crypt it is not the strongest available in the crypto world but it is also not very weak:

If the password has a length of 10 (the suggested length) with letters and numbers then in my computer it would take 17000 years to break using bruteforce with hashcat. My computer is 10000 times slower than the used at https://terahash.com with 8 Nvidia cards. So it would take 1 year to break in this powerful system.
If the password has mayus and minus letters and numbers then my computer would take 4millon years to break it.
Attackers with that kind of resources would break a linux routers much much faster using some existing security issues here and there. I am more inclined to let libremesh users know about these issues.

G10h4ck (approved):

ilario:
> > One question (a bit late): the exposed lime-community file exposes the md5crypt of the community password and the hash is a bit weak.
> > In the lime-example.txt stronger alternatives are mentioned, this is great, but these are not supported by FBW and neither by OpenWrt! (are not available by default OpenWrt 19.07 (but can be activated) and in the case they are enabled, doubts are present on the higher security due to lack of salting: https://forum.openwrt.org/t/passwd-a-sha256-sha512-less-secure-than-default/13117/ )
> > We should warn users on the risk to use important passwords there.
> 
> I would say that md5crypt it is not the strongest available in the crypto world but it is also not very weak:
> 
> If the password has a length of 10 (the suggested length) with letters and numbers then in my computer it would take 17000 years to break using bruteforce with hashcat. My computer is 10000 times slower than the used at https://terahash.com with 8 Nvidia cards. So it would take 1 year to break in this powerful system.
> If the password has mayus and minus letters and numbers then my computer would take 4millon years to break it.

All true, but most of the passwords are not complex and bruteforce is not needed, for example it took less than 1 day for onlinehashcrack.com to crack the example password "mypassword" (FBW would not accept this as it does not have any number, but still, a well designed cracking dictionary can include numbers).

> Attackers with that kind of resources would break a linux routers much much faster using some existing security issues here and there. I am more inclined to let libremesh users know about these issues.

Consider:

* the additional risk if someone is connecting the WAN to a modem, as described in #280: in that case an attacker from the internet can take the root password hash, crack it and enter via SSH;
* that someone could install FBW but set a weak password manually via SSH (not that common right now as it is not suggested by default).

I have no idea how to warn the users in the latter case.

PR #816, opened 2020-11-30
pirania: pirania-dnsmasq should not read /etc/dnsmasq.conf

altergui:
Right now, /etc/dnsmasq.conf points to /etc/dnsmasq.d/,
which contains lime-proto-anygw-20-ipv6.conf and makes
pirania-dnsmasq do IPv6 router-advertisements on br-lan and anygw
very confusing and totally unrelated to pirania-dnsmasq function

Signed-off-by: Gui Iribarren <gui@altermundi.net>


G10h4ck (approved):

spiccinini (approved):

nicopace:
there is something that needs to be considered in relation to local domain resolution.
I am not sure if this fix breaks it or not, but just want to raise it as it is tied to dnsmasq configs and the files that are populated by the dhcp service.

PR #818, opened 2020-12-02
github-ci: replace deprecated set-env

spiccinini:
Hopefuly fIxes github action.
I am not exactly sure but as per the following documentation  https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files this should do the work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -31,7 +31,7 @@ jobs:
           echo "DEST_DIR=$TAG" >> $GITHUB_ENV
 
       - name: Upload packages to GitHub Pages
-        uses: peaceiris/actions-gh-pages@v3.7.0-8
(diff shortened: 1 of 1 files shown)


spiccinini:
It says that failed but it doesn't ... I don't understand
image

ilario:
The error seems to be in peaceiris/actions-gh-pages old version we're using.
Can you edit the action file to point to a newer release?

https://github.com/peaceiris/actions-gh-pages/releases

spiccinini:
Ok it should be working now

ilario (approved):

germanferrero (approved):

ilario:
network-profiles actions are failing for the same reason: 
https://github.com/libremesh/network-profiles/actions

@spiccinini could you apply the same fix also there? Thanks!

PR #820, opened 2020-12-08
lime-proto-batadv change MAC also of wlan interfaces

ilario:
In order to avoid the "received packet on bat0 with own address as source address" annoying (but harmless) message we are customizing the MAC address of the ethernet VLAN interface, see #726 (and #703 and #724).
But in some cases (see [here](https://lists.libremesh.org/pipermail/lime-users/2020-December/001855.html)) for some reason the wireless VLAN interface can be taken as MainIF by batman-adv and therefore needs to have a customized mac address also.

This PR simply moves the MAC customization code out of the `if` block, with no further edit.

### DIFF
--- a/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
+++ b/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
@@ -87,19 +87,19 @@ function batadv.setup_interface(ifname, args)
 		--! TODO: Use DSA to check if ethernet device is capable of bigger MTU
 		--! reducing it
 		mtu = network.MTU_ETH_WITH_VLAN
-		
-		--! Avoid dmesg flooding caused by BLA with messages like "br-lan:
-		--! received packet on bat0 with own address as source address".
-		--! Tweak MAC address for each of the interfaces used by Batman-adv
-		--! 00 + Locally administered unicast .. 2 bytes from interface name
-		--! .. 3 bytes from main interface
-		local id = utils.get_id(ifname)
-		local vMacaddr = network.primary_mac();
-		vMacaddr[1] = "02"
-		vMacaddr[2] = id[2]
-		vMacaddr[3] = id[3]
-		uci:set("network", owrtDeviceName, "macaddr", table.concat(vMacaddr, ":"))
 	end
+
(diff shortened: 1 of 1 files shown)


G10h4ck:
The change looks good, did you tested it on a few different devices/architectures, every-time a mac is changed it could tickle up some unexpected bug even in the hardware driver :-\

ilario:
I quickly checked on YouHua WR1200JS and TP-Link WDR3600, please help with the testing as I will not have time to test further or on other routers. @G10h4ck @spiccinini

G10h4ck:
> I quickly checked on YouHua WR1200JS and TP-Link WDR3600, please help with the testing as I will not have time to test further or on other routers. @G10h4ck @spiccinini

ok that seems enough as those devices have quite different wifi stack

G10h4ck (approved):

spiccinini:
This won't change the mac address seen by the other wireless stations right? sothe ipv6 link local addresses of the wireless interfaces won't change? Sorry if this seems a silly question

G10h4ck:
> This won't change the mac address seen by the other wireless stations right? sothe ipv6 link local addresses of the wireless interfaces won't change? Sorry if this seems a silly question

I expect the macaddress of the wlanX-mesh interface so the one seen as station from the others and it's relative link local ipv6 address should not change. Instead I expect all of those properties change for the vlan interface. So just to make it explicit I expect wlanX-mesh_123 to get a different mac address and a different IPv6 link local address. Do we assume wlanX-mesh and wlanX-mesh_123 have same properties in some part of the stack?

spiccinini:
> > This won't change the mac address seen by the other wireless stations right? sothe ipv6 link local addresses of the wireless interfaces won't change? Sorry if this seems a silly question
> 
> I expect the macaddress of the wlanX-mesh interface so the one seen as station from the others and it's relative link local ipv6 address should not change. Instead I expect all of those properties change for the vlan interface. So just to make it explicit I expect wlanX-mesh_123 to get a different mac address and a different IPv6 link local address. 

Thanks Gio! 

> Do we assume wlanX-mesh and wlanX-mesh_123 have same properties in some part of the stack?

Not that I am aware.

spiccinini (approved):

PR #823, opened 2020-12-10
shared-state: multiple fixes

G10h4ck:
As busybox ping6 doesn't support float intervals, ping6 from iputils is needed, in recent versions of OpenWrt the package iputils-ipv6 has been removed and meld into iputils-ping, the output of the command also changed son a few tricks to keep compatibility with all this different versions were needed.
As a bonus also lime-debug get its dependencies fixed for OpenWrt master in a retro-compatible way.

### DIFF
--- a/packages/lime-debug/Makefile
+++ b/packages/lime-debug/Makefile
@@ -16,13 +16,16 @@ PKG_VERSION=$(GIT_COMMIT_DATE)-$(GIT_COMMIT_TSTAMP)
 
 include $(INCLUDE_DIR)/package.mk
 
+include ../../ping6retrocompat.mk
+
 define Package/$(PKG_NAME)
 	TITLE:=libremesh debug utils
 	CATEGORY:=LibreMesh
 	MAINTAINER:=Gioacchino Mazzurco <gio@eigenlab.org>
 	URL:=http://libremesh.org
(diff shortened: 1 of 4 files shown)


spiccinini (approved):

ilario:
In
```
ifeq ($(shell git -C $(TOPDIR) merge-base --is-ancestor 98b3526bf23e8d1b48939c937c9b12e4f2160415 HEAD ; echo $$?),0)
```
what does 98b3526bf23e8d1b48939c937c9b12e4f2160415 refer to?
Can you put a comment on this also next to that code line?

G10h4ck:
> In
> 
> ```
> ifeq ($(shell git -C $(TOPDIR) merge-base --is-ancestor 98b3526bf23e8d1b48939c937c9b12e4f2160415 HEAD ; echo $$?),0)
> ```
> 
> what does 98b3526bf23e8d1b48939c937c9b12e4f2160415 refer to?
> Can you put a comment on this also next to that code line?

It is explained here https://github.com/libremesh/lime-packages/pull/823/files#diff-299866dd193ecc8035ca88a8483a263c1aac45549a498149e1e28c43d6f8fcd3R13

IMHO putting it at same line or just before the line where it is used would make the stuff even more unreadable

nicopace (approved):

PR #830, opened 2020-12-17
Lime proto babeld fixes

altergui:
right now, babeld doesn't route between clouds over ipv6, since `main_ipv6_address` is set to `fd%N1:%N2%N3:%N4%N5::/64`

also, conceptually it makes total sense to redistribute `fc00::/7` the same way `10.0.0.0/8` is redistributed

### DIFF
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -82,9 +82,9 @@ function babeld.configure(args)
 	uci:set("babeld", "localdeny", "action", "deny")
 
 	-- Avoid redistributing enything else
-	uci:set("babeld", "deny", "filter")
-	uci:set("babeld", "deny", "type", "redistribute")
-	uci:set("babeld", "deny", "action", "deny")
+	uci:set("babeld", "denyany", "filter")
+	uci:set("babeld", "denyany", "type", "redistribute")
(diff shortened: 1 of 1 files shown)


packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
```diff
@@ -77,9 +82,9 @@ function babeld.configure(args)
 	uci:set("babeld", "localdeny", "action", "deny")
 
 	-- Avoid redistributing enything else
-	uci:set("babeld", "dany", "filter")
-	uci:set("babeld", "dany", "type", "redistribute")
-	uci:set("babeld", "dany", "action", "deny")
+	uci:set("babeld", "deny", "filter")
```
G10h4ck, line 85: `dany` here was a contraption of "deny any", as this is the `uci` section name it should not change the sematic, but I hope we don't get some naming collision in the future :p

G10h4ck (approved):

spiccinini (approved):

packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
altergui: :facepalm: ok, maybe rename it to deny_any?

G10h4ck: yeah that seems better, could you do that before merging this?

PR #834, opened 2020-12-18
Refactor lime location as lib and fix location shared state publishing

spiccinini:


G10h4ck (approved):

nicopace (approved):
Did a quick proof reading, and it looks ok, but have not run it.
Thanks for adding tests to it!

PR #835, opened 2020-12-18
[2020.1] Fix shared-state-nodes_and_links

G10h4ck:


ilario:
Maybe 2020.2?

spiccinini (approved):

spiccinini:
> Maybe 2020.2?

Well we can backport fixes to the 2020.1 and then release a 2020.2 from it in the short time (well we only have 13 days remaining in 2020!

nicopace (approved):

PR #839, opened 2021-01-11
Add ubus-tmate to expose tmate control for terminal sharing

nicopace:
This PR proposes an interface to a package that allows console sessions to be shared amongst many people.

This is a piece of a feature that will allow, through the lime-app, for users to request support to other networks and be able to monitor what they do with the access it was given, and remove the access once the intervention is done or you don't want them to be accessing your router anymore. It will have as a lime-app screen.

https://user-images.githubusercontent.com/1136597/104211357-bd7a8580-5412-11eb-9d9b-74123590c28e.mp4
(Chrome plays it, Firefox doesn't :( )

Opening it up so to get your general feedback on:
- using tmate instead of lime-remotesupport: I think it is good that we can generalize it, we will eventually be able to upstream it

You will have to add the tmate feed to your OpenWRT buildroot to build it `src-git tmate https://github.com/project-openwrt/openwrt-tmate.git`. It needs OpenWRT 19.07.0 as it needs libssh 0.7.6.


packages/tmate-ubus/files/usr/libexec/rpcd/tmate
```diff
@@ -0,0 +1,70 @@
...
+        utils.printJson({status = "ok", session = { rw_ssh = rw_ssh, ro_ssh = ro_ssh, clients = clients}})
+    else
+        utils.printJson({status = "error", msg = "no session"})
+    end
+    return
+end
+
+local function open_session(msg)
+    utils.unsafe_shell("tmate -S "..TMATE_SOCK.." new-session -d")
+    utils.unsafe_shell("tmate -S "..TMATE_SOCK.." wait tmate-ready")
```
spiccinini, line 38: tengo una duda, esto del wait que pasa si no hay internet o no se puede conectar? Puede llegar a tardar mucho y no responder?

packages/tmate-ubus/files/usr/libexec/rpcd/tmate
```diff
@@ -0,0 +1,70 @@
...
+    utils.printJson({status = "ok"})
+    return
+end
+
+local function close_session(params)
+    if unix_socket_listening(TMATE_SOCK) then
+        utils.unsafe_shell("tmate -S "..TMATE_SOCK.." kill-session -t 0")
```
spiccinini, line 45: I would try to wrap all the repeated `tmate -S .. TMATE_SOCK` into something like `tmate_cmd_as_str('kill-session -t 0')`

packages/tmate-ubus/files/usr/libexec/rpcd/tmate
nicopace: si... tal cual.
en paralelo chequeamos la conexión, estuvimos hablando de eso con @germanferrero

germanferrero: No chequeamos la conexión antes de ese wait. El timeout es el chequeo ahora. Gracias por cazarlo SAn.

Yes need to fix open_session timeout waiting time

packages/tmate-ubus/Makefile
```diff
@@ -0,0 +1,31 @@
+include $(TOPDIR)/rules.mk
+
+PKG_NAME:=tmate-ubus
```
germanferrero, line 3: All other ubus packages follow the naming convention ubus-* .We should do the same for this one.

packages/tmate-ubus/files/usr/libexec/rpcd/tmate
```diff
@@ -0,0 +1,70 @@
...
+
+local function get_session(msg)
+    if unix_socket_listening(TMATE_SOCK) then
+        local rw_ssh = utils.unsafe_shell("tmate -S "..TMATE_SOCK.." display -p ''#{tmate_ssh}|head -c -1")
+        local ro_ssh = utils.unsafe_shell("tmate -S "..TMATE_SOCK.." display -p ''#{tmate_ssh_ro}|head -c -1")
+        local clients = utils.unsafe_shell("tmate -S "..TMATE_SOCK.." display -p ''#{tmate_num_clients}|head -c -1")
+
+        utils.printJson({status = "ok", session = { rw_ssh = rw_ssh, ro_ssh = ro_ssh, clients = clients}})
+    else
+        utils.printJson({status = "error", msg = "no session"})
```
germanferrero, line 31: I duplicate my previous comment about the API here to keep the feedback in a single place:

Reserve status error for unhappy paths. We could use 
{status="ok", session="no initialized"}
As session=nil will end up in a json without that field: {"status": "ok"}

germanferrero (commented):
Looks great @nicopace. I'll test it out with final version of the limeapp and report back

packages/ubus-tmate/etc/tmate/tmate.conf
```diff
@@ -0,0 +1,5 @@
+# in this source-file we expect to have the server connection info, don't remove. If server config doesn't exist (in this or that file), will use default tmate.io servers
```
germanferrero, line 1: Why is this file not under /files ?
Is it a mistake?

nicopace: it is indeed! thanks for rising that!

nicopace:
How do we do with this package that requires a package from a different feed?
Should we add something in the readme to instruct which feeds need to go with this one?

packages/ubus-tmate/files/usr/lib/lua/tmate.lua
```diff
@@ -0,0 +1,50 @@
...
+
+
+local TMATE_SOCK = "/tmp/tmate.sock"
+local TMATE_CONFIG = "/etc/tmate/tmate.conf"
+
+local tmate = {}
+
+function tmate.cmd_as_str(cmd)
+   final_cmd = "tmate -f "..TMATE_CONFIG.." -S "..TMATE_SOCK.." "..cmd
+   return utils.unsafe_shell(final_cmd)
```
germanferrero, line 15: utils (lime.utils) it's not imported in this file.

spiccinini: Indeed and it is working because lime.utils is "installed" (by mistake) in the global namespace. There are one or two other modules that have the same issue.

spiccinini (changes requested):
Great PR! Good work Nico!

This is just a "administrative" request (and a reminder for everyone included myself), plase can you modify your commit mesages so they start with the module name that is modified? (like `tmate: fix bla bla`) And while you are at it please remove the final dot and use 'fix bla bla' instead of 'fixes bla bla'. Yes, we should add this recomendations in the contributing info.
A quick way to do it is to do: `git rebase -i master` and then replace all the `pick` by `reword` (save and close) and then all the commit mesages will be desplayed for you to be changed.

spiccinini (approved):

germanferrero (approved):
I've test it against https://github.com/libremesh/lime-app/pull/289 and works great

Issue #699, opened 2020-04-10
FirstBootWizard setup country code


spiccinini:
When starting a new network le the user select its conutry.

ilario:
+1
This was suggested in https://github.com/libremesh/lime-packages/issues/327#issuecomment-560132371
Consider that some country code could be a problem with the default 5 GHz channels: #594.

PR #843, opened 2021-02-04
fbw: add optional country config

spiccinini:
This change adds the ability to configure the country (for the wireless config) at network creation in the First Boot Wizard.
This needs to be exposed in the lime-app to be useful.


ilario:
Thanks! This goes in the direction to fix #699.

nicopace (approved):

G10h4ck (approved):

PR #851, opened 2021-03-01
shared-state-bat_hosts: mv acl file to the correct directory

### LINKED ISSUE #850: lime-app fetching name
hello @all,

ive following issue in lime app---fetching name,,,,search search search, openwrt 19.07.7, i have given each router its own name in lime-node so that i can distinguish them better. but the status fetching name does not change, see pictures, no matter on which router, this is an example of router "rootnode"
fetchingname
fetchingname1
when i open the page for the first time it remains empty, this is another node in the same mesh
fetchname3

spiccinini:
Fixes  #850  lime-app fetching name 

There are two acl directories, the file was in the incorrect one.


germanferrero (approved):
Thanks!

nicopace (approved):

PR #852, opened 2021-03-01
fbw: support community lime-assets

### LINKED ISSUE #846: Firstbootwizard should retrieve /etc/lime-assets/community directory
Since we merge https://github.com/libremesh/lime-packages/pull/736 users can define assets/scripts to be run at firstboot or at running lime-config for every node in a community network. But, we are not copying those assets when performing firstbootwizard.

I'll say this is a bug, that we should fix asap

spiccinini:
The lime-assets are documented here https://libremesh.org/docs/en_config.html#lime-assets
This PR only adds the lime-assets support to fbw.

The lime-assets of level community are shared as a tar.gz and fetched as part of the first boot wizard config step.

Fixes #846

### DIFF
--- a/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
+++ b/packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
@@ -22,15 +22,18 @@ local uci = require "uci"
 
 local fbw = {}
 
-
-
 fbw.WORKDIR = '/tmp/fbw/'
-fbw.HOST_CONFIG_PREFIX = 'lime-community__host__'
+fbw.COMMUNITY_HOST_CONFIG_PREFIX = 'lime-community__host__'
+fbw.COMMUNITY_ASSETS_TMPL = 'lime-community_assets__host__%s.tar.gz'
 
 utils.execute('mkdir -p ' .. fbw.WORKDIR)
 
 function fbw.log(text)
     nixio.syslog('info', '[FBW] ' .. text)
+end
+
+local function lime_community_assets_name(hostname)
+    return fbw.WORKDIR .. string.format(fbw.COMMUNITY_ASSETS_TMPL, hostname)
 end
 
 -- Write lock file at begin
@@ -42,7 +45,7 @@ end
 
 -- Remove old results
 function fbw.clean_tmp()
-    utils.execute('rm -f ' .. fbw.WORKDIR .. fbw.HOST_CONFIG_PREFIX .. '*')
+    utils.execute('rm -f ' .. fbw.WORKDIR .. '*')
 end
 
 -- Save working copy of wireless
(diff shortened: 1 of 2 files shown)


packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -161,27 +164,32 @@ function fbw.fetch_config(data)
...
     if (hostname == '') then hostname = host end
-    local signal = data.signal
-    local ssid = data.ssid
-    local filename = fbw.WORKDIR .. fbw.HOST_CONFIG_PREFIX .. hostname
 
-    utils.execute("wget --no-check-certificate http://[" .. data.host .. "]/cgi-bin/lime/lime-community -O " .. filename)
-    if not utils.file_exists(filename) then
+    local lime_community_fname = fbw.WORKDIR .. fbw.COMMUNITY_HOST_CONFIG_PREFIX .. hostname
+
+    utils.execute("wget --no-check-certificate http://[" .. data.host .. "]/cgi-bin/lime/lime-community -O " .. lime_community_fname)
```
nicopace, line 170: it is odd that wget uses ` --no-check-certificate` when the request is http... not from this commit, but still :)

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
```diff
@@ -210,6 +218,14 @@ function fbw.apply_file_config(file, hostname)
     -- Clean previus lime configuration and replace lime-community
     config.reset_node_config()
     utils.execute("cp " .. filePath .. " /etc/config/" .. config.UCI_COMMUNITY_NAME)
+
+    -- Setup the shared lime-assets
+    local remote_hostname = string.sub(file, #fbw.COMMUNITY_HOST_CONFIG_PREFIX + 1)
+    local lime_community_assets_fname = lime_community_assets_name(remote_hostname)
+    if utils.file_exists(lime_community_assets_fname) then
+        utils.execute(string.format("tar xfz %s -C /etc/lime-assets/", lime_community_assets_fname))
```
nicopace, line 226: the use of string.format instead of the tradicional string concat might confuse the reader, as it is less used, more in these cases... not an issue per se, but I guess for readability might be better using the double dot expression

nicopace (commented):
It lacks documentation of what lime-assets is and what is it used for, as right now for me is difficult to understand what you wanted to do with it.
How would you document it?
The code looks ok though...

spiccinini:
> It lacks documentation of what lime-assets is and what is it used for, as right now for me is difficult to understand what you wanted to do with it.
> How would you document it?
> The code looks ok though...

Thanks for the review nico!! The lime-assets are documented here https://libremesh.org/docs/en_config.html#lime-assets 
This PR only adds the lime-assets support to fbw.

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
spiccinini: Indeed, that is because uhttpd may have "redirect to https" enabled with a self signed certifcate (check this PR https://github.com/libremesh/lime-packages/pull/799)

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
spiccinini: Well I would say that string concatenation is not "traditional" at all :P 
But yes maybe it uses a too advanced feature?

nicopace: not too advance :) but felt that it was the thing that is used more often, and there is no specific use of the format method in this case, so I would go with the one more frequently used... 
this is an opinion though, so please do as you see fit :)

packages/first-boot-wizard/files/usr/lib/lua/firstbootwizard.lua
nicopace: Thanks! was not aware of that!

germanferrero (approved):
Thanks @spiccinini! It looks great, I didn't test it but I know you did

PR #854, opened 2021-03-08
Fix shared state location publisher

germanferrero:
At https://github.com/libremesh/lime-packages/pull/834 we have broken the shared state publisher for nodes_and_links.

The broken version would publish an entry for each field (links, hostname, macs, etc) in shared-state db, breaking the non-robust lime-app ui for the map.

### DIFF
--- a/packages/shared-state-nodes_and_links/files/usr/bin/shared-state-publish_nodes_and_links
+++ b/packages/shared-state-nodes_and_links/files/usr/bin/shared-state-publish_nodes_and_links
@@ -18,5 +18,7 @@
 local JSON = require("luci.jsonc")
 local location = require("lime.location")
 
-io.popen("shared-state insert nodes_and_links", "w"):write(JSON.stringify(location.nodes_and_links()))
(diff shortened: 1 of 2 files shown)


spiccinini (approved):

spiccinini:
Thanks @germanferrero ! We need to find a good way to test the binary files

nicopace (approved):

nicopace:
I give it a little more of thought, and I think these issues can be the drivers of extending the coverage of our test suite.
When something like these happens, we should be able to reproduce it in a test so it reduces the chance of it ever happening again (at least on this part of the code).

spiccinini:
> I give it a little more of thought, and I think these issues can be the drivers of extending the coverage of our test suite.
> When something like these happens, we should be able to reproduce it in a test so it reduces the chance of it ever happening again (at least on this part of the code).

Yes! In this case I think that the only Easy way is to move all the code to a library, test it as we do with unit tests, and leave only one function call in the binary file.

PR #857, opened 2021-03-11
lime-utils: on upgrade preserve configs by default

spiccinini:
If no preserve config option is passed (`preserve_config = nil`) then preserve the config instead of not preserving it. This semantic matches better the sysupgrade semantic.
As in the last lime-app release there is no longer a preserve config checkbox, this change enforces (fixes) the intended behaviour of preserving the config instead of not doint it.


nicopace (approved):

germanferrero (approved):
Great!

PR #859, opened 2021-03-12
fbw: use a more permisive temporary wifi config

spiccinini:
The temporary wifi config is used to download the config from the
detected remote nodes. Using a restrictive wireless config (high
mcast_rate, short distance) makes this download phase to take more
time or even to fail. Also it does not make a lot of sense for a
node that was already selected as a candidate to download the config.
If we want to restrict the nodes per their signal we should do that
in the scan phase and not in the temporary config.


nicopace:
Will it work if only one side of the conversation has a permissive wifi config?
I mean, if the other network nodes are already setup, then they are using a restrictive wifi config and that is ok.

I usually encourage community members to visit each other in order to do a "pairing", so there is no chance they don't get the fbw config... would that be enough? or are there any usecases that require it?
could you explain a little bit more what happens to you in which scenary?

spiccinini:
> Will it work if only one side of the conversation has a permissive wifi config?
Not guaranteed to work, but It would work better.

> I usually encourage community members to visit each other in order to do a "pairing", so there is no chance they don't get the fbw config... would that be enough? or are there any usecases that require it?

When you run the FBW a download for all the scaned nodes will be performed. Going near a node does not guarantee that other nodes, will low quality signal, won't  be in the nodes list.

> could you explain a little bit more what happens to you in which scenary?

Two scenarios came to my mind (and both happened to me in practice):
* There are only two nodes and the signal quality is low because of the distance (and the antenna is not yet correctly aligned) and you can't go to the other node because of time, etc. In this scenario you want to increase the chance of downloading the config.
* There are N nodes visible, 1 with high signal and the others with low signal. FBW takes a lot of time when trying to download from the low quality nodes (like several minutes in total time, as it takes like 1 minute before giving up per node, and per radio).

Even if we tell people to came close to a node, the other nodes of the netwok may disturb the functionality. 

The `distance` config change  is very clear for me as configuring a node with max distance 1000m does not imply that it won't see the other node, it will just work very very bad.
The signal config change is not that clear but I think that it won't hurt. A clear path to a better solution may be to filter the nodes list using a minimum signal quality level. The problem with this is that each equipment has a different dBm value for minimum quality.

spiccinini:
Also, the current default multicast rate is 6000 for 5ghz and 24000 for 2.4ghz

nicopace:
> When you run the FBW a download for all the scaned nodes will be performed. Going near a node does not guarantee that other nodes, will low quality signal, won't be in the nodes list.

Definitely... but the one you already know you have line of sight with will, as it is the one that you get nearby...
Still it is not always obvious which one you should pair with, so this might not be the best approach

nicopace (approved):

PR #860, opened 2021-03-13
Delete packages/luci-mod-status directory

aparcar:
The package also exists in the luci.git repository. No reason to have it twice under the same name.


spiccinini:
I don't know if there is functionality in this module that is not provided by the package in luci.git 
Do we know anyone using the luci modules of libremesh? If there are no users we may delete/archive also the other luci modules.

spiccinini:
If there are active users we may rename this module?

dangowrt:
> If there are active users we may rename this module?

Yes, that would be a good solution. Just having it use an already existing name makes it impossible to actually install this package in unpatched OpenWrt, as the package found in luci.git would always be preferred.

aparcar:
Fine with me, someone please come up with a name and rename it. Apart from that I solved the problem on my setup differently so I don't mind it overwriting anything.

Issue #863, opened 2021-03-19
Mention to a specific commit breaks builds without full git history


ilario:
Issue reported by @dangowrt in this comment:

https://github.com/libremesh/lime-packages/commit/ad8d1a3dbaa52362d6859e720aec08cb8c180a51#commitcomment-48372702

It refers to this code line:

https://github.com/libremesh/lime-packages/blob/ad8d1a3dbaa52362d6859e720aec08cb8c180a51/ping6retrocompat.mk#L17

dangowrt:
Is anyone working on resolving this?

spiccinini:
> Is anyone working on resolving this?

Hi! Not myself, if you have a patch please share :) 
Also I think that we may drop 18.06 support if it simplifies things (I don't know if it does in this case).

ilario:
OpenWrt 18.06 support should for sure be abandoned from master branch as it is in End-of-Support.

PR #870, opened 2021-04-05
ping6retrocompat.mk: use grep instead of relying in git history

dangowrt:
Fixes https://github.com/libremesh/lime-packages/issues/863


spiccinini (approved):
Great, I've tested it only in a 19.07 build environment

dangowrt:
I found an even better way using GNU Make `wildcard` function. In this way we don't need to launch shell and it works great in SDK and buildroot.

aparcar:
Lgtm

aparcar (approved):

PR #871, opened 2021-04-08
prometheus-node-push-influx: new package

altergui:
Collects the same statistics as prometheus-node-exporter-lua,
  but instead of exposing them as a server, with prometheus syntax,
  pushes them using influx line protocol over TCP
  to a remote VictoriaMetrics server.


ilario:
Wow!!!
Please consider (sooner or later) pushing to OpenWrt repository, where the other [prometheus-* packages](https://github.com/openwrt/packages/tree/openwrt-19.07/utils) are.

packages/prometheus-node-push-influx/files/usr/bin/prometheus-node-push-influx
```diff
@@ -111,8 +111,16 @@ function push_loop()
...
+      local collectors_output = ""
+      output = function (str) collectors_output = collectors_output..str.."\n" end
       run_all_collectors(col_names)
+      tcp:send("POST /write HTTP/1.1\n")
+      tcp:send("User-Agent: prometheus-node-push-influx\n")
+      tcp:send("Host: "..server_address..":"..server_port.."\n")
+      tcp:send("Content-Type: application/x-www-form-urlencoded\n")
+      tcp:send("Content-Length: "..string.len(collectors_output).."\n")
+      tcp:send("\n")
+      tcp:send(collectors_output)
```
spiccinini, line 123: If I am not wrong, if collectors_output is too long then the `send` call may not send all the bytes, and the remaining should be sent in another call to `send` (checking the output of  `send` as it provides the bytes sent (well at least in C, I am not sure on lua's tcp wrapper to send). 
Is possible that the collectors_output does not fit in "one buffer" ?

altergui: i'm so happy i don't do C
i haven't seen truncated output
but double-checking the docs, you might be right
http://w3.impa.br/~diego/software/luasocket/tcp.html
and it doesn't hurt to add a check. will do

ilario:
This looks very useful!
But I think that two separate contents are being mixed here: the `altermundi-grafana` should be on LibreMesh repositories but the rest should be on OpenWrt ones. In my opinion, clearly.

ilario:
Ah no, I just saw that it checks for its presence before importing it. Nice :)

Issue #867, opened 2021-03-26
List network nodes service


spiccinini:
A list of nodes that belong to the network with their current status of being connected/disconnected and other info (like the firmware version, board, etc) is needed for multiple use cases like:

* List all the nodes of the network that its firmware version is outdated.
* List which nodes are not online but they should (they don't have power, or the antenna is misaligned, etc). 
* Have a table with the nodes and its boards. These may help to maintain
* Get a list of the online and offline nodes so that a config upgrade that needs to be performed in all the nodes at the same time can take place when all the nodes are online.
* Many many more community features

A list like this is currently maintained by hand (in spread sheets, text files, etc) in many communities. To provide these features by software we need a way to get these lists programmatically and in a way that these information is not lost when the mesh is powered off. 

Requirements:
* The database of the list of nodes have to be populated without user/community intervention
* The database must be persistent
* The community should be able to mark a node as not participating anymore (for example people moved, or the name of the node changed, etc). When retrieving the nodes the nodes marked as not participating are not listed (at least by default). This change should be able to be performed from any node.
* At least an attribute of the node is to be connected or disconnected (like reachable / unreachable). More attributes may be added at later.

Idea of implementation: 
Provide an ubus and a cli to show the network nodes list.

Use shared state in two ways:
* A multi-writer (aka multi-master) and persistent mode to store the "non volatile information": hostname, firmware version, board, etc. All the information is stored in all the nodes using /etc as storage (an not /tmp). 
* use the current shared-state mode  (single-writer and non persistent) to store the "volatile information": the connection status.

G10h4ck:
I don't understand why the multi-writer is needed for the node list, can't each node publish it's own perspective and then the perspectives melded together on visualization to create a global view of the network?

spiccinini:
> I don't understand why the multi-writer is needed for the node list, can't each node publish it's own perspective and then the perspectives melded together on visualization to create a global view of the network?

Yes it is possible to persist all the perspectives of all the nodes in each node and then perform the meld in the visualization of the data (we would have to store all because if one node says that another left the network then that will be needed also stored in each node). 
In my point of view it is better to address this directly in the data layer instead of doing that above, not only for the bandwidth and storage savings (that may be negligeable) but because the system is easier to understand and matches better this use case: we want that each node has the same view of the system, if they have a different view is because it did not converge yet.
What do you think?

germanferrero:
[...]

PR #872, opened 2021-04-14
Add shared state multiwriter

### LINKED ISSUE #868: shared-state multi writer mode proposal
Currently shared-state can't store information in a way that multiple nodes can modify all its keys (multi-writer).
This is because the author of a key is re-publishing it with the biggest TTL all the time, not leaving space for others to modify the value of that key. It was though for a _one writer per key_ scheme.

The proposal is to add a multi-writer mode that can also persist the information so that it survives reboots of the full network (for example due to a big power loss in the region).

There are a lot of use cases, for example: List network nodes service #867

The following are considered as premises:
* Each key is created only by one node "at a time". So for example there is not a "mynetwork" key in the database that all the nodes create on first boot. The typical key may be the name of the the hostname that adds information of itself but other schemes are intended to be suported, for example storing information with uuids like vouchers, notifications, etc.
* The information stored does not vary often (for example, firmware versions may be stored but not current signal levels)

Proposal:
* Provide the comand `shared-state-multiwriter` to access this mode/part of the database
* Use a permanent storage by default (maybe selectable to temporary in the future if we find an use case for that)
* Use an ordered way of modifying the value of a key so that it can be merged "monotonically". The simplest scheme is to use a timestamp (the last timestamp wins in the merge). Another scheme is to use a counter and a random string nounce. For example the key starts at with `counter: 1, nonce: "s6qgHkMT"`, and then other node changes it to `counter: 2, nonce: "mEVfaCb7"`.  If the counters are equal the lexicographic order is used is used to select the winner for the merge:  
```python
if local.counter == other.counter: 
    return local if local.nonce > other.nonce else other
```

spiccinini:
This PR implements the Multi Writer proposal from #868 . Thanks @germanferrero for the ideas and implementation, I have only fixed some bugs.

Fixes #868

Currently shared-state can't store information in a way that multiple nodes can modify all its keys (multi-writer).
This is because the author of a key is re-publishing it with the biggest TTL all the time, not leaving space for others to modify the value of that key. It was though for a one writer per key scheme. This scheme can be used to create "streams" of data applying inteligence above shared-state (for example the scheme used by pirania to use a new key for each modification of each voucher) but this is not only more complex, it has to be redone for each use case. So we believe that being able to have keys that can be modified by any node is a step forward to simplicity in some use cases (mainly when the informacion needs persistance).

This adds a multi-writer mode that can also persist the information so that it survives reboots of the full network (for example due to a big power loss in the region).

There are a lot of use cases, for example: List network nodes service #867

The following should me considered as premises when using the multiwriter:
* Each key is created only by one node "at a time". So for example there is not a "mynetwork" key in the database that all the nodes create on first boot. The typical key may be the name of the the hostname that adds information of itself but other schemes are intended to be suported, for example storing information with uuids like vouchers, notifications, etc.
* The information stored does not vary often (for example, firmware versions may be stored but not current signal levels)

This PR:
* Provides the comand shared-state-multiwriter to access this mode/part of the database
* Uses a permanent storage by default (maybe selectable to temporary in the future if we find an use case for that)
* Uses a timestamp as an ordered way of modifying the value of a key so that it can be merged "monotonically".


packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -543,4 +543,20 @@ function utils.is_valid_mac(string)
     end
 end
 
+function utils.deepcompare(t1,t2)
```
nicopace, line 546: As this implementation is recursive, it might have a heap issue.
You can explore an iterative one, that is cheaper and equally fast.

nicopace: would also be good to have tests for these implementations, both for documenting their use and for robustness

nicopace (commented):
great work Ger!
i've started reviewing it, though haven't finished... so gave a few comments and will come back later.

packages/lime-system/files/usr/lib/lua/lime/utils.lua
germanferrero: I added tests for this function, but couldn't come up with a non recursive implementation quickly.
I understand you mean stack issue instead of heap.

packages/shared-state/files/usr/lib/lua/shared-state.lua
```diff
@@ -148,7 +148,6 @@ function SharedStateBase:sync(urls)
 	self:unlock()
 	self:_sync(urls)
 	self:lock()
-	self:load() -- Take in account changes happened during sync
```
germanferrero, line 256: @spiccinini , thanks for catching this issue!

But together with the change at load method, we are changing the behavior of sync.
In master version, changes to the local storage made by another process during sync are merged with those coming from the sync process. I think that's what "Take in account changes happened during sync" means.
In this branch version, we are ignoring those changes.
If shared-state-publish_dnsmasq_leases gets executed during sync for instance, and the host has given new leases, those will enter shared state storage, then leave when sync finish, and then come back when publish happens again.
I think the most safe option is to keep the current behavior of the master branch.
Thoughts?

G10h4ck:
> Uses a timestamp as an ordered way of modifying the value of a key so that it can be merged "monotonically".

`shared-state` originally used timestamps too, but that was quite problematic because in many situations routers doesn't have the clock synchronized so it because a mess very easily. One quite common case was `shared-state` running at boot before `ntp` managed to sync the clock, so we ended up with never expiring records timestamped to 2024 or whatever date in far future (past dates weren't problematic as they would be overwritten once the node clock synced).

How is the case node clock is out of sync (both in far past and in far future) handled?

packages/lime-system/files/usr/lib/lua/lime/utils.lua
G10h4ck: An early check `if ty1 == ty2 then return true end` would optimize a bunch of cases ;)
Moreover this implementation would overflow the heap in case of a back-reference inside the table

spiccinini:
> > Uses a timestamp as an ordered way of modifying the value of a key so that it can be merged "monotonically".
> 
> `shared-state` originally used timestamps too, but that was quite problematic because in many situations routers doesn't have the clock synchronized so it because a mess very easily. One quite common case was `shared-state` running at boot before `ntp` managed to sync the clock, so we ended up with never expiring records timestamped to 2024 or whatever date in far future (past dates weren't problematic as they would be overwritten once the node clock synced).
> 
> How is the case node clock is out of sync (both in far past and in far future) handled?

Right, nodes with dates in the future are very problematic. If that case was seen in the field I propose to change the method using other merge strategy like the specified here https://github.com/libremesh/lime-packages/issues/868 , what do you think about the other strategy? (quoting)

. Another scheme is to use a counter and a random string nounce. For example the key starts at with counter: 1, nonce: "s6qgHkMT", and then other node changes it to counter: 2, nonce: "mEVfaCb7". If the counters are equal the lexicographic order is used is used to select the winner for the merge.

packages/shared-state/files/usr/lib/lua/shared-state.lua
spiccinini: This is fixed now.

spiccinini:
I've pushed a new merge strategy using a counter and a random number and one fix for a bug that german spoted.
I think that now this PR can be merged and the following improvements to shared-state to be worked in a new PR.

germanferrero:
Oh, the comments I made at https://github.com/libremesh/lime-packages/pull/873 should have been here!

packages/lime-system/files/usr/lib/lua/lime/utils.lua
spiccinini: German already addressed this issues. In my opinion is ok to have a recursive strategy.

germanferrero (approved):
I think we are now ready to merge :)

spiccinini:
As per the discused roadmap with gio (https://github.com/libremesh/lime-packages/issues/884) this is ready for merging.

nicopace (approved):
I skimmed through it ... it looks fine, no big issues.
thanks for the contributions!

PR #873, opened 2021-04-14
Add shared state network nodes

### LINKED ISSUE #867: List network nodes service

spiccinini:
This PR adds an rpcd/ubus service that provides the list of the nodes of the network and the ability to remove one node from the list. This uses the shared-state multiwriter as main storage so it should be easiear to first review and merge the shared-state changes of PR #872 .

Most of the code was done by @germanferrero , great work! 

Fixes #867

Implementation details (see rationale in the related issue #867):

* Use the new shared-state-multiwriter:
    * The keys are the hostnames of the nodes
    * Each node inserts its own key in the database
    * Any node can modify the membership of other node (because the community should be able to mark a node as not participating anymore, for example people moved, or the name of the node changed). This is provided by ubus in this PR and its frontend in the Lime-App PR https://github.com/libremesh/lime-app/pull/301
    * This database is persistent
    * attributes stored in the multiwriter database: `hostname, ipv4, ipv6, board, fw_version, member`
    * attributes computed from the `nodes_and_links` database plus the `membership` attribute: status. It can be: `recently_connected`, `disconnected`, `gone`. A node is considered disconnected if its member attribute is True but its key is not in the `nodes_and_links` database. As its key can be present but the node may be off or disconnected for a couple of minutes (until the TTL drops to 0 and it is removed from the nodes_and_links db) the status is `recently_connected` and not just `connected` so that it does not give "false information".
    * A cli to show the network nodes list

This is the CLI provided, it only shows and can't be used to remove nodes (use the ubus interface for that):
image


nicopace:
haven't reviewed yet, but looks very nice! thanks for the work put in this!

germanferrero:
The motivation for `recently_connected ` also applies for `recently_disconnected` as the "disconnected" node could recently come back to the network but synchronization of it's nodes_and_links key between that node and the current node didn't happened yet.
Should we named it `recently_disconnected` too?

germanferrero:
> The motivation for `recently_connected ` also applies for `recently_disconnected` as the "disconnected" node could recently come back to the network but synchronization of it's nodes_and_links key between that node and the current node didn't happened yet.
> Should we named it `recently_disconnected` too?

After chatting with @spiccinini, `recently_disconnected` wouldn't be accurate neither, since the node not in sync may be the current node. `disconnected_or_not_sync` or  `very_likely_disconnected` would, as the time needed to spread nodes_and_links key after boot should be very short.

germanferrero (approved):
Looks great @spiccinini good job :)

packages/shared-state/tests/test_shared_state.lua
```diff
@@ -126,13 +126,50 @@ describe('Shared State MultiWriter tests #sharedstatemultiwriter', function()
     after_each('', function()
         test_utils.teardown_test_dir()
     end)
-    
-    it('test merge data from remote db with newer timestamp', function()
+
+    it('test insert new data', function()
+        local ss = shared_state.SharedStateMultiWriter:new('foo')
+        local sharedStateB = shared_state.SharedStateMultiWriter:new('B')
```
germanferrero, line 132: non used :)

packages/shared-state/files/usr/lib/lua/shared-state.lua
```diff
@@ -324,11 +326,18 @@ function SharedStateMultiWriter:insert(data)
...
+function shared_state._getFortune()
+	return math.random(1, 100000)
+end
+
 function SharedStateMultiWriter:_insert(key, data)
 	local lv = self.storage[key]
 	if (lv == nil or not utils.deepcompare(lv.data, data)) then
+		local changes = lv and lv.changes + 1 or 0
 		self.storage[key] = {
 			lastModified=os.time(),
```
germanferrero, line 338: Are we keeping lastModified for some reason? debugging, it could be misleading though...

germanferrero (changes requested):
I left minor changes request in the comments :) Looks great!

packages/shared-state/files/usr/lib/lua/shared-state.lua
spiccinini: My idea is that having the last modf time may be useful in the future for example for housekeeping.  mean, to have the date already in the db now will allow to know later if we can prune a value or not. But yes it is not useful right know so I am also not very convinced that having it is a good idea.

germanferrero: Yeah, I guess it may be helpfull in the future and it doesn't harm no

germanferrero (approved):

packages/shared-state-network_nodes/files/etc/uci-defaults/shared-state-network_nodes-cron
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+unique_append()
```
nicopace, line 3: this is not the first time I see this function.
should it become part of a library?

packages/shared-state-network_nodes/files/usr/lib/lua/network-nodes.lua
```diff
@@ -0,0 +1,100 @@
+local shared_state = require("shared-state")
+local utils = require("lime.utils")
+local config = require("lime.config")
+
+local network_nodes = {}
+
+function network_nodes.node(hostname, member, fw_version, board, ipv4, ipv6)
```
nicopace, line 7: this function feels like a private function within the module... could it be the case?

packages/shared-state-network_nodes/files/usr/lib/lua/network-nodes.lua
```diff
@@ -0,0 +1,100 @@
...
+    return {hostname=hostname, member=member, fw_version=fw_version, board=board, ipv4=ipv4, ipv6=ipv6}
+end
+
+function network_nodes.serialize_for_network_nodes(node)
```
nicopace, line 11: same as for some of the following, these look like private methods belonging to the module.

nicopace (approved):

packages/shared-state-network_nodes/files/etc/uci-defaults/shared-state-network_nodes-cron
spiccinini: Yes, when we create the lime-utils package :)

packages/shared-state-network_nodes/files/usr/lib/lua/network-nodes.lua
spiccinini: Thanks for pointing it. I've pushed a refactor for this and all the non public functions to be accesed with a leading underscore (as they need to be reachable for the tests).

PR #881, opened 2021-05-13
lime-location: keep the location settings

spiccinini:
Add missing location settings file to the keep list.


germanferrero (approved):
Thanks

nicopace (approved):

PR #882, opened 2021-05-24
lime-system: flush autogen before modifying it

spiccinini:
This fixes a bug that prevented configs that were autogenerated to be present in the lime-autogen file, they were used "in memory"  during the lime-config and then lost. This behaviour is very confusing as reading the lime-autogen was measlieading.


nicopace (approved):

germanferrero (approved):
I do understand this fix, and it's correct. An alternative, as we discussed with @spiccinini , would have been to change  `config.end_batch()` to use `uci::commit` instead of `uci::save`, which would persist all current changes at autogen, without the need of flushing at the end.
But the current fix is ok too.

What I do not understand is why lime-hwd-* modules need to write sections with `autogenerated` option. Is this for debugging purposes only? I'll ask this again in the mailing list.

ilario:
@germanferrero  Setting the autogenerated option to false is the only way to stop the annoying lime-hwd-* scripts from overwriting user's manual edits to those sections.

PR #883, opened 2021-06-05
Add feature mac-based config file

itec78:
Can set specific configuration for every device based on mac address.
You can create a configuration file /etc/config/lime-[mac address] containing the setting for the single node.
Running lime-config the configuration is read from that file (if available), and has priority between lime-community and lime-node.
So you can compile a single firmware including multiple config files, one for every node.


ilario:
Amazing! Thanks @itec78!
More discussion on this is available on the mailing list [here](https://lists.libremesh.org/pipermail/lime-users/2021-June/thread.html).
Do you have a convenient way to obtain the primary MAC for a device?

itec78:
> Do you have a convenient way to obtain the primary MAC for a device?
If you run lime-config from console the config file name is prompted.
Mac-config: lime-14cc20cd41c1
On my node it corresponds to the wan mac

PR #885, opened 2021-06-22
Fix tmate black screen when joining

germanferrero:
Tmate shows an initial screen like this:
Screenshot_20210621_234039

For a recently created session, but only for ssh/terminal clients, in the web client (as in the LimeApp) a black screen is shown instead.

By sending CTRL + c right after opening the session, that screen is not shown anymore, and both ssh/terminal and web clients see the Libremesh Banner and the prompt when joining.


germanferrero:
Thanks @nicopace for finding the bug :)

spiccinini (approved):

nicopace (approved):

PR #889, opened 2021-07-13
Fix the packages URL

### LINKED ISSUE #888: OPKG does not find our Packages.gz for releases
Just tested an image downloaded from 
http://repo.libremesh.org/expansive_emancipation/2020.1/targets/ath79/generic/

```
# opkg update
[...]
Downloading http://feed.libremesh.org/2020.1/Packages.gz
*** Failed to download the package list from http://feed.libremesh.org/2020.1/Packages.gz

Downloading http://feed.libremesh.org/profiles/Packages.gz
Updated list of available packages in /var/opkg-lists/profiles
Downloading http://feed.libremesh.org/profiles/Packages.sig
Signature check passed.
Collected errors:
 * opkg_download: Failed to download http://feed.libremesh.org/2020.1/Packages.gz, wget returned 8.
```

OPKG looks for `.../2020.1/...` but, as the directory in the feed is `v2020.1`: https://github.com/libremesh/lime-feed/tree/gh-pages/v2020.1, it should look for `.../v2020.1/...` instead.

I suppose that the fix is as easy as add a `v` here (?):

https://github.com/libremesh/lime-packages/blob/06399e5de7fb7b9d3cf13c19f7bf70accb11299a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos#L18

RhnSharma:
Fixes #888
Hi @ilario, how does this look?
Let me know if this needs any changes.
Thanks

### DIFF
--- a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
+++ b/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos
@@ -15,7 +15,7 @@ feeds_file="/etc/opkg/limefeeds.conf"
 [ "$LIME_CODENAME" == "development" ] && {
 	packages_url="http://feed.libremesh.org/master";
 } || {
-	packages_url="http://feed.libremesh.org/$LIME_RELEASE"
+	packages_url="http://feed.libremesh.org/v$LIME_RELEASE"
 }
 
 profiles_url="http://feed.libremesh.org/profiles"


ilario:
I did not test but looks right! :)
Thanks!!!

ilario (approved):

spiccinini:
I think that it may be best to remove the `v` from the feed itself (instead of adding it to the feed config), what do you think?

ilario:
Mh so that would mean to modify the Github action file instead?

spiccinini:
> Mh so that would mean to modify the Github action file instead?

Yes, and  i believe a we should do a manual move of the last release files so it will start working.

ilario:
Replaced by merged #891

PR #890, opened 2021-07-16
Add client hotspot wwan connection handling

spiccinini:
For multiple things it is useful to connect temporarily the device to an access point. To provide remote access/support through tmate or other means it is helpful that someone close to the device share its mobile phone connection to the router. The lime-app will use this functionality.

Also the check-internet package now provides an optional ubus interface to check.

TODO:
- [ ] add tests


ilario:
Nice!!
Sounds like similar to uto-usb-wwan, correct?
 https://github.com/libremesh/lime-packages/tree/master/packages/auto-usb-wwan

How will the user enable the wwan on the LibreMesh node?

spiccinini:
> Nice!!
> Sounds like similar to auto-usb-wwan but directly with the router's antennas, correct?
> https://github.com/libremesh/lime-packages/tree/master/packages/auto-usb-wwan

> 
> How will the user enable the wwan on the LibreMesh node?

It can be done with the following (using radio1 instead of the default radio0)
```
root@LiMe-1efa02:~# echo | /usr/libexec/rpcd/lime-utils-admin call hostpot_wwan_is_connected
{"status":"ok","connected":false}

root@LiMe-1efa02:~# echo '{"radio": "radio1"}'  | /usr/libexec/rpcd/lime-utils-admin call hostpot_wwan_enable
{"status":"ok"}

root@LiMe-1efa02:~# echo | /usr/libexec/rpcd/lime-utils-admin call hostpot_wwan_is_connected
{"status":"ok","connected":true,"signal":-47}

root@LiMe-1efa02:~# echo | /usr/libexec/rpcd/lime-utils-admin call hostpot_wwan_disable
{"status":"ok"}

root@LiMe-1efa02:~# echo | /usr/libexec/rpcd/lime-utils-admin call hostpot_wwan_is_connected
{"status":"ok","connected":false}
```

packages/ubus-lime-utils/files/usr/libexec/rpcd/lime-utils-admin
```diff
@@ -106,6 +161,9 @@ if arg[1] == 'call' then
     elseif  arg[2] == 'firmware_upgrade' then firmware_upgrade(msg)
     elseif  arg[2] == 'last_upgrade_metadata' then last_upgrade_metadata(msg)
     elseif  arg[2] == 'firmware_confirm' then firmware_confirm(msg)
+    elseif  arg[2] == 'hotspot_wwan_enable' then hotspot_wwan_enable(msg)
```
germanferrero, line 164: Needs to be hotspot_wwan_enable(unpack(msg)), and the same for hotspot_wwan_disable

spiccinini: Thanks! That is why tests were needed :) I did not use unpack to maintain consistency.

packages/ubus-lime-utils/tests/test_hotspot_wwan.lua
```diff
@@ -71,6 +88,67 @@ describe('hotspot_wwan tests #hotspot_wwan', function()
         assert.is.equal(-66, status.signal)
     end)
 
+    it('test is_safe is false when no mesh ifaces configured', function()
```
germanferrero, line 91: Shouldn't be "test is_safe is false when mesh iface configured in the same radio"?

packages/ubus-lime-utils/tests/test_hotspot_wwan.lua
```diff
@@ -71,6 +88,67 @@ describe('hotspot_wwan tests #hotspot_wwan', function()
...
+        uci:set('wireless', 'radio0', 'hwmode', '11a')
+        uci:set('wireless', 'radio0', 'macaddr', '01:23:45:67:89:AB')
+        uci:set('wireless', 'radio0', 'disabled', '0')
+        uci:commit('wireless')
+
+        local is_safe = hotspot_wwan._is_safe('internet', 'psk2', 'radio0')
+        assert.is_true(is_safe)
+    end)
+
+    it('test is_safe is false when mesh ifaces configured', function()
```
germanferrero, line 104: It doesn't seem to me that we are setting up a "mesh iface" over radio 0 in this test

packages/ubus-lime-utils/tests/test_hotspot_wwan.lua
spiccinini: yes

packages/ubus-lime-utils/tests/test_hotspot_wwan.lua
spiccinini: right.... did a mistake when refactored the code to remove duplication in the tests

germanferrero: fixed

germanferrero (approved):

luandro (approved):
Well tested and simple logic. Great work ;)

PR #891, opened 2021-07-31
add tail command to remove first character of the string

### LINKED ISSUE #888: OPKG does not find our Packages.gz for releases

RhnSharma:
Fixes #888
Hi @ilario, I have made the changes as discussed in https://github.com/libremesh/lime-packages/issues/888#issuecomment-889988221. 
Let me know if this needs any changes.
Thanks

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Set package destination
         run: |
-          export TAG=$(echo "${{ github.ref }}" | cut -d '/' -f 3-)
+          export TAG=$(echo "${{ github.ref }}" | cut -d '/' -f 3- | tail -c +2)
           echo "$TAG"
           echo "DEST_DIR=$TAG" >> $GITHUB_ENV
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Set package destination
         run: |
-          export TAG=$(echo "${{ github.ref }}" | cut -d '/' -f 3- | tail -c +2)
+          export TAG=$(echo "${{ github.ref }}" | cut -d '/' -f 3- | perl -pe 's/v([0-9])/$1/')
           echo "$TAG"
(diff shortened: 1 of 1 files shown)


ilario:
I fear I proposed the solution in a rush and it's wrong.
That "$TAG" is going to be used also for "master", right? @spiccinini @aparcar 
In that case we have to remove the first character ooonly if it's a tag like `v2020.1`.
So the `| tail -c +2` could be replaced by `| perl -pe 's/v([0-9])/$1/'` (hoping that that environment has perl).
Sorry @RhnSharma for the mess, could you change the content of this pull request or open another?
Thanks for your patience

RhnSharma:
@ilario I have made the change you suggested. Please take a look at it and let me know if this needs any changes.
Thanks

ilario (approved):
Looks good to me (hoping that perl is available there). We should merge this and then rename the directories in lime-feed.

spiccinini:
Looks good to me! thanks for doing it

spiccinini (approved):

PR #896, opened 2021-09-22
Migrate frequency band suffix options to uci sections for each band.

germanferrero:
While working on a service that would expose ubus enpoints for changing AP settings (such as disabling ap mode), I found out that it is really difficult to work with `list mode` option when the `frequency suffix` is used.
A lot of thinking and code is needed to unpack the configuration scenario, handle all corner cases and pack it again when doing some change programmatically (or even manually).
This PR proposal is to migrate from configs like these:

```
config lime 'wifi'
    list modes 'ap'
    list modes 'apname'
    list modes 'ieee80211s_5ghz'
    option htmode_5ghz '40'
```

To

```
config lime-wifi-band '5ghz'
    list modes 'ap'
    list modes 'apname'
    list modes 'ieee80211s'
    option htmode '40'

config lime-wifi-band '2ghz'
    list modes 'ap'
    list modes 'apname'
```
 
In this way, disabling ap mode from 2ghz requires only to remove ap mode from `2ghz.modes` list
Otherwise we need to think of replacing 'ap' with 'ap_5ghz' in `wifi.modes` to avoid modifying 5ghz radios behavior too.

The code for wireless.configure gets simplified aswell.
Now all options for a radio would be taken from three levels, in order of precedence: radio (config wifi), band (config lime-wifi-band), general (config lime wifi).

A migration script is added as a uci-default so lime-community and lime-node configurations out there are transformed to the new syntaxis **before** running lime-config on first boot.

**Breaking changes**
- Drops support for "<option_name>_<frequency_suffix>" on specific radio configs. This is a no sense config, as one radio always corresponds to one frequency band. (I'll like to hear from you if you consider that this should still be supported).


germanferrero:
> just wanted to point out, this is not _always_ the case. I remember using dualband radios where the same radio has both frequency bands (they can't be used concurrently). I remember one USB dongle that worked like this. In the beginning, routers with two independent radios (one for each band) where marketed as "simultaneous dual-band" to differentiate them from single-radio dualband (which meant "non-simultaneous dual-band") I guess such single-radio dualband are rare nowadays and can be left unsupported.

Oh, I didn't know, that's curious.. I'd say that we can go forward with this config schema and address the support for these kind of radios in the future if anyone claim the need for it.

PR #900, opened 2021-10-11
Relax switch vlan filter

G10h4ck:
Address recent bug discovered in Quintana Libre with eth1.2 LibreRouter interface being ignored when specific configuration section for it is needed.

### DIFF
--- a/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
+++ b/packages/lime-proto-babeld/files/usr/lib/lua/lime/proto/babeld.lua
@@ -92,11 +92,11 @@ end
 
 function babeld.setup_interface(ifname, args)
 	if not args["specific"] and ifname:match("^wlan%d+.ap") then
-		utils.log("lime.proto.babeld.setup_interface(...)", ifname, "ignored")
+		utils.log("lime.proto.babeld.setup_interface(%s, ...) ignored", ifname)
 		return
 	end
 
-	utils.log("lime.proto.babeld.setup_interface(...)", ifname)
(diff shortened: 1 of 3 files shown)


ilario:
I did not test, but looks like the second %s is missing in the log line "dev_parser found Wifi device".

Neither this I tested: can the CPU port be discovered with something like "swconfig dev switch0 help"?

What will happen with DSA? The CPU port is not there anymore?

ilario:
So I understand that we can merge this, right?

spiccinini (approved):

ilario:
Info just got at Battlemesh from @arinc9:

the CPU ports are the one which have such a file:
```
/sys/firmware/devicetree/base/ethernet@1e100000/mdio-bus/switch@1f/ports/port@6/ethernet
```

Ref.
https://lkml.kernel.org/netdev/f7d5d961-8f66-e744-ab59-0077445d5873@arinc9.com/T/

G10h4ck:
that is a really interesting info! Thanks @ilario ! If you can ask also if it is possible to get more informations about the switch in a "standard" way would be great.
Does this method works also with pre-DSA versions?
we just need to check if it is well supported on all kernel versions used by OWRT and then improve our code!

arinc9:
> that is a really interesting info! Thanks @ilario ! If you can ask also if it is possible to get more informations about the switch in a "standard" way would be great.

We have the network interfaces DSA creates. Other than that, you can check the devicetree bindings for DSA.

> Does this method works also with pre-DSA versions? we just need to check if it is well supported on all kernel versions used by OWRT and then improve our code!

No, the binding on the referred link is only for DSA. Keep in mind that not all targets on OpenWrt use DSA.

PR #906, opened 2021-12-22
lime-proto-anygw: use the configured domain as a hostrecord

meskio:
If a system domain is being configured use it as a domain for this
router by configuring the DNS hostrecord to use it together with
thisnode.info and minodo.info.

### DIFF
--- a/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
+++ b/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
@@ -18,6 +18,10 @@ function anygw.configure(args)
 	anygw.configured = true
 
 	local cloudDomain = config.get("system", "domain")
+	if not utils.has_value(anygw.FQDN, cloudDomain) then
+		table.insert(anygw.FQDN, cloudDomain)
+	end
+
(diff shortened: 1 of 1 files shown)


ilario (approved):
Welcome Meskio!!!!
I did not test but the changes make sense and the code should work for that.
Thanks!

meskio:
Thanks ilario. I did test in a librerouter setting my own domain or leaving thisnode.info and I believe it does work fine. But I don't have much experience with lua. I'm happy to hear that the code makes sense :)

spiccinini (approved):
Hi @meskio thanks for contributing with a fix! The code looks good! I was not aware (or I forgot) about this bug.

I would use an utility helper that we have to simplify the code:
```
if not utils.has_value(anygw.FQDS, cloudDomain) then
    table.insert(...)
```

Usually we are trying to write unittests for each PR but as anygw.lua does not have any test yet and you are not used to lua then I am not asking you to add a test. Anyway have manually tested and it works as expected.

meskio:
@spiccinini thanks for the review. I updated it to use `util.has_value`.

spiccinini:
Thanks Meskio!

meskio:
Thank you for merging it :)

PR #920, opened 2022-03-08
WIP: Add services for service discovery

luandro:


ilario:
Hi Luandro!
I would suggest to rename the package to something more telling than just `services`.
Thanks!

ilario:
Also, is Network > Services a new submenu?
I see only this package there.
Did you consider including this package in an already existing menu?

PR #928, opened 2022-07-27
Moved udhcpc.user file to installer script

### LINKED ISSUE #927: ubus-lime-utils file conflict when compiling with OpenWrt 22.03
There's a file conflict when compiling on top of OpenWrt 22.03.0-rc4 between:
https://github.com/libremesh/lime-packages/blob/400ac95448a1f1d5a83496488e40c302cde766f0/packages/ubus-lime-utils/files/etc/udhcpc.user
and:
https://github.com/openwrt/openwrt/blob/130118f7aa3191a2afd244d49e3057d4c5688676/package/network/config/netifd/files/etc/udhcpc.user

This is what the error looks like:

```
Collected errors:
 * check_data_file_clashes: Package ubus-lime-utils wants to install file /home/ilario/software/openwrt/build_dir/target-mips_24kc_musl/root-ath79/etc/udhcpc.user
	But that file is already provided by package  * netifd
 * opkg_install_cmd: Cannot install package lime-app.
 * check_data_file_clashes: Package ubus-lime-utils wants to install file /home/ilario/software/openwrt/build_dir/target-mips_24kc_musl/root-ath79/etc/udhcpc.user
	But that file is already provided by package  * netifd
 * opkg_install_cmd: Cannot install package ubus-lime-utils.
make[2]: *** [package/Makefile:70: package/install] Error 255
```

ilario:
Fix #927
Taken inspiration from these other uci-defaults files:
https://github.com/libremesh/lime-packages/blob/2c4b783f68860f2542b3910fc6c46f9566b1b205/packages/lime-proto-bmx6/files/etc/uci-defaults/85-add-bmx6-addresses-to-hosts
and
https://github.com/libremesh/lime-packages/blob/f4f8d004ff8564452ab705b02687eff9cc063334/packages/lime-app/files/etc/uci-defaults/95-lime-app-rpc-acl
Should work, but I still have to test it on a router.

### DIFF
--- a/packages/ubus-lime-utils/files/etc/uci-defaults/ubus-lime-utils-udhcpc.user-setup_hotspot_watchping
+++ b/packages/ubus-lime-utils/files/etc/uci-defaults/ubus-lime-utils-udhcpc.user-setup_hotspot_watchping
@@ -1,7 +1,7 @@
 #!/bin/sh
 grep setup_hotspot_watchping /etc/udhcpc.user || {
 	echo "ubus-lime-utils: adding setup_hotspot_watchping to /etc/udhcpc.user"
-	cat > /etc/udhcpc.user << EOF
+	cat >> /etc/udhcpc.user << EOF
 setup_hotspot_watchping() {
     ifname="client-wwan"
     gw=$(ip r show default dev $ifname | while read default via ip rest; do [[ $via == "via" ]] && echo $ip && break; done)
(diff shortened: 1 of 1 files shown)


ilario:
To me it does not look like there is any need to update the tests, is this correct @spiccinini?

G10h4ck:
@ilario see my coments at https://github.com/libremesh/lime-packages/issues/927

ilario:
Just tested this PR and it is broken.
@G10h4ck pointed out that we could use the /etc/udhcpc.user.d/ directory and place stuff in there. This way is much better so I am going to close this PR.
The doubt is whether that directory will work also on OpenWrt 19.07. To be tested.

PR #930, opened 2022-09-01
Fixes in the ansible script and tools

irina556:
1. Remove host interface
        It remove an interface that was created on the host after starting a node with qemu_dev_start
2. Put the message in ansible to access the cloud nodes
        By raising a cloud of nodes we were able to access each one by clusterssh. When you wanted to do it with ssh, it was not 
        possible because there was a problem with the keys, so a message is added indicating how you can access any cloud 
        node through ssh.
3. Add "files" in ansible with the images
4. Port change
        When I wanted to launch a node with internet access using --enable-wan for qemu_dev_start, a port collision occurred as 
        dnsmasq uses port 53 by default, which is usually occupied by the systemd-resolved service. To fix this, another port is 
        allocated so that this collision does not occur.
5. Change ifconfig to IP
        In the linux_bridge.py module, there were calls to ifconfig, and since modern Linux distributions don't have this command 
        to view the TCP/IP configuration, it wasn't possible to run the node cloud (unless you installed the net- tools for the use of 
        ifconfig). To avoid this inconvenience for any other user, calls that were previously made with ifconfig are now made with IP.

### DIFF
--- a/tools/ansible/modules/linux_bridge.py
+++ b/tools/ansible/modules/linux_bridge.py
@@ -56,11 +56,6 @@ class LinuxBridge (object) :
         self.state = module.params['state']
         return
 
-    def brctl (self, cmd) :
-
-       	return self.module.run_command (['brctl'] + cmd)
-
-
     def ip(self, cmd) : 
 
         return self.module.run_command (['ip'] + cmd) 
--- a/tools/ansible/modules/linux_bridge.py
+++ b/tools/ansible/modules/linux_bridge.py
@@ -54,17 +54,16 @@ class LinuxBridge (object) :
         self.module = module
         self.bridge = module.params['bridge']
         self.state = module.params['state']
-
         return
 
     def brctl (self, cmd) :
 
-        return self.module.run_command (['brctl'] + cmd)
+       	return self.module.run_command (['brctl'] + cmd)
 
 
-    def ifconfig (self, cmd) :
+    def ip(self, cmd) : 
 
-        return self.module.run_command (['ifconfig'] + cmd)
+        return self.module.run_command (['ip'] + cmd) 
 
 
     def br_exists (self) :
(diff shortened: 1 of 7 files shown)


germanferrero:
Please do remove librerouteros files from the commit history

tools/ansible/modules/linux_bridge.py
```diff
@@ -54,17 +54,18 @@ def __init__ (self, module) :
         self.module = module
         self.bridge = module.params['bridge']
         self.state = module.params['state']
+        #self.type = module.params['bridge']
```
germanferrero, line 57: remove commented line :)

tools/ansible/qemu_cloud_start.yml
```diff
@@ -104,5 +104,8 @@
           now to manage the cloud you can do:
 
           clusterssh {{ linklocals | join(' ') }}
+          
+          to access to node you can do:
```
germanferrero, line 108: This ends up printing:
```
to access to node you can do:
ssh -o StrictHostKeyChecking=no -o"HostkeyAlgorithms +ssh-rsa" fe80::5000:ff:feab:c000%lm_cloudC fe80::5000:ff:feab:c001%lm_cloudC fe80::5000:ff:feab:c002%lm_cloudA fe80::5000:ff:feab:c003%lm_cloudA fe80::5000:ff:feab:c004%lm_cloudA fe80::5000:ff:feab:c005%lm_cloudB fe80::5000:ff:feab:c006%lm_cloudB fe80::5000:ff:feab:c007%lm_cloudB fe80::5000:ff:feab:c008%lm_cloudB fe80::5000:ff:feab:c009%lm_cloudD fe80::5000:ff:feab:c010%lm_cloudD fe80::5000:ff:feab:c011%lm_cloudD
```

which has two issues:
- user root should be used "root@fe80::5000:ff:feab:c000%lm_cloudC"
- ssh takes only one host as parameter not multiple hosts.

germanferrero: Also, we could print  -o options for clusterssh too

Issue #771, opened 2020-10-03
Package lime-utils independently from lime-system


ilario:
I copy and paste here the [original discussion](https://github.com/libremesh/lime-packages/issues/766#issuecomment-702742621) started on #766:

@spiccinini 
> Some packages depend only in the lime.utils part of lime-system package. In my opinion we should refactor out this lua module into its own package.

@ilario 

> > Some packages depend only in the lime.utils part of lime-system package. In my opinion we should refactor out this lua module into its own package.
> 
> To add some data, here you are a table of the required lua files from lime-system by each package:
> 
> | Package | Required lua file |
> |-------------|------------------------|
> | deferable-reboot | lime.config lime.utils |
> | first-boot-wizard | lime.config lime.utils lime.wireless |
> | lime-hwd-ground-routing | lime.config lime.hardware_detection lime.utils |
> | lime-hwd-openwrt-wan | lime.config lime.hardware_detection lime.network lime.hwd.openwrt_wan lime.utils |
> | lime-hwd-usbradio | lime.config lime.hardware_detection lime.utils |
> | lime-proto-anygw | lime.config lime.network lime.system |
> | lime-proto-babeld | lime.config lime.network |
> | lime-proto-batadv | lime.config lime.network lime.proto lime.utils lime.wireless |
> | lime-proto-bgp | lime.config lime.network lime.utils |
> | lime-proto-bmx6 | lime.config lime.network lime.utils lime.wireless |
> | lime-proto-bmx7 | lime.config lime.network lime.utils lime.wireless |
> | lime-proto-olsr | lime.config lime.network lime.utils lime.wireless |
> | lime-proto-olsr2 | lime.config lime.network lime.utils lime.wireless |
> | lime-proto-olsr6 | lime.config ime.network lime.utils lime.wireless |
> | lime-proto-wan | lime.utils |
> | lime-smart-wifi | lime.config lime.utils lime.wireless |
> | lime-system | lime.config lime.generic_config lime.mode lime.network lime.system lime.utils lime.wireless |
> | lime-webui | lime.config |
> | safe-upgrade | lime.utils |
> | shared-state | lime.utils lime.wireless |
> | ubus-lime-batman-adv  | lime.utils |
> | ubus-lime-bmx6 | lime.utils |
> | ubus-lime-groundrouting | lime.config lime.network |
> | ubus-lime-location | lime.config lime.network lime.utils lime.wireless |
> | ubus-lime-metrics | lime.utils |
[...]

Issue #940, opened 2022-10-19
Missing dependencies from luci-lib-jsonc


ilario:
jsonc is widely used across LibreMesh packages.
Some of the packages include it in their dependencies, but some do not.
These are the packages in which I think the dependency from `luci-lib-jsonc` is missing:
```
check-internet
eupgrade
lime-proto-bmx6
lime-system
pirania
shared-state-network_nodes
ubus-lime-batman-adv
ubus-lime-bmx6
ubus-lime-groundrouting
ubus-lime-location
ubus-lime-metrics
ubus-lime-openairview
ubus-lime-utils
ubus-tmate
```

Issue #929, opened 2022-08-23
Remove lime-map-agent and luci-app-lime-location packages


ilario:
The presence of the [lime-map-agent](https://github.com/libremesh/lime-packages/tree/master/packages/lime-map-agent) and [luci-app-lime-location](https://github.com/libremesh/lime-packages/tree/master/packages/luci-app-lime-location) packages in this repository causes plenty of warnings that confuse users.
The messages look like this:
```
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'libremap-agent', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-location', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-system', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-wireless', which does not exist
WARNING: Makefile 'package/feeds/libremesh/luci-app-lime-location/Makefile' has a dependency on 'libremap-agent', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'libremap-agent', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-location', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-system', which does not exist
WARNING: Makefile 'package/feeds/libremesh/lime-map-agent/Makefile' has a dependency on 'luci-lib-libremap-wireless', which does not exist
WARNING: Makefile 'package/feeds/libremesh/luci-app-lime-location/Makefile' has a dependency on 'libremap-agent', which does not exist
```

PR #941, opened 2022-10-19
Fixing some dependencies

ilario:
As indicated in #940, I just realized that some dependencies are missing.
Here go some fixes, more will be added to this PR, please do not merge yet.

### DIFF
--- a/packages/babeld-auto-gw-mode/Makefile
+++ b/packages/babeld-auto-gw-mode/Makefile
@@ -10,7 +10,7 @@ define Package/$(PKG_NAME)
   CATEGORY:=Utilities
   TITLE:=Babel auto Internet gateway module.
   MAINTAINER:=Santiago Piccinini <spiccinini@altermundi.net>
-  DEPENDS:=+libubus-lua +watchping +lime-proto-babeld
+  DEPENDS:=+libubus-lua +watchping +lime-proto-babeld +lime-system
   PKGARCH:=all
 endef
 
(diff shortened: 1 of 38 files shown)


ilario:
Many of these dependencies will have to be changed when we fix #771, but #771 is had its second birthday a few weeks ago, so it does not look like we're going to fix it soon (hope is still there).

ilario:
My personal dependencies crusade is over, I am not sure I fixed everything (e.g. luci.tools stuff I didn't check where it is).
Just checked with the buildroot of OpenWrt 22.03 and no additional warning (the ones from libremap are still there, see https://github.com/libremesh/lime-packages/issues/929) appears, so it should be ok to merge.
I didn't check with OpenWrt 19.07 but none of the packages mentioned as dependencies should have been renamed since 2019.

ilario:
One doubt: the `luci-lib-json` dependency has not been updated to `luci-lib-jsonc` for `luci-mod-status` and `luci-app-openairview`. I don't know if these two packages would work also with jsonC.

ilario:
> Just checked with the buildroot of OpenWrt 22.03 and no additional warning (the ones from libremap are still there, see #929) appears, so it should be ok to merge. I didn't check with OpenWrt 19.07 but none of the packages mentioned as dependencies should have been renamed since 2019.

Correction: there IS a new warning.

```
tmp/.config-package.in:101823:error: recursive dependency detected!
tmp/.config-package.in:101823:	symbol PACKAGE_pirania is selected by PACKAGE_shared-state-pirania
tmp/.config-package.in:60861:	symbol PACKAGE_shared-state-pirania is selected by PACKAGE_pirania
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"
```

I confirmed with OpenWrt 19.07 and this is the only new warning that appears with both 22.03 and 19.07.

Reported as a separate issue on #942

G10h4ck (approved):

PR #947, opened 2022-11-16
migrate-wifi-bands-cfg check for conf files being existing

### LINKED ISSUE #945: Error in first run of migrate-wifi-bands-cfg
When installing lime-system via `opkg` on a clean OpenWrt system, this error is printed:

```
Configuring lime-system.
/usr/bin/lua: /usr/bin/migrate-wifi-bands-cfg:38: bad argument #1 to 'pairs' (table expected, got nil)
stack traceback:
	[C]: in function 'pairs'
	/usr/bin/migrate-wifi-bands-cfg:38: in function 'move_restof_to_specific_band'
	/usr/bin/migrate-wifi-bands-cfg:51: in function 'migrate_file'
	/usr/bin/migrate-wifi-bands-cfg:59: in main chunk
	[C]: ?
```

this happens as at the first install there is no `lime-community` nor `lime-node` files and this list is empty:

https://github.com/libremesh/lime-packages/blob/de7f38f6e555a18684728fd4c02ec866e3abc753/packages/lime-system/files/usr/bin/migrate-wifi-bands-cfg#L7

So there should a check for the `options` variable to not be `nil` between these lines:

https://github.com/libremesh/lime-packages/blob/de7f38f6e555a18684728fd4c02ec866e3abc753/packages/lime-system/files/usr/bin/migrate-wifi-bands-cfg#L37-L38

ilario:
Fix #945

### DIFF
--- a/packages/lime-system/files/usr/bin/migrate-wifi-bands-cfg
+++ b/packages/lime-system/files/usr/bin/migrate-wifi-bands-cfg
@@ -35,6 +35,9 @@ end
 
 local function move_restof_to_specific_band(file_name)
     local options = uci:get_all(file_name, 'wifi')
+    if options == nil then
+        return
+    end
     for key, value in pairs(options) do
         local band_name = utils.split(key, '_')[2]
         if band_name == '2ghz' or band_name == '5ghz' then


ilario:
Tested, see the output of the `opkg install lime-system` command here:
https://github.com/libremesh/lime-packages/pull/948#issuecomment-1326726898

and when running it manually after the installation, the output is:

```
root@LiMe-eb7eac:~# migrate-wifi-bands-cfg 
migrate-wifi-modes: migration finished, now run lime-config
```

and the section is actually run (checked adding some print).

G10h4ck (approved):

Issue #958, opened 2022-12-10
DSA config: bat0 does not get into the bridge


ilario:
In the mailing list, [Gothos reported](https://lists.autistici.org/message/20221210.180941.ac807e16.en.html) that adding bat0 in the br-lan bridge partially fixed the issues found when using the current LibreMesh code on top of OpenWrt 21.02 and 22.03.

With swconfig (approx OpenWrt 19.07), the interfaces included in br-lan were defined like this in `/etc/config/network` (see full file in https://github.com/libremesh/lime-packages/issues/951#issue-1469264556):

```
config interface 'lan'
	option type 'bridge'
	option ifname 'eth0.1'
[...]
```

but with DSA (approx OpenWrt 21.02 and 22.03), now this section only contains the IP configuration, while the bridged interfaces list are defined in another section of the same file:
```
config device
	option name 'br-lan'
	option type 'bridge'
	list ports 'lan1'
	list ports 'lan2'
	list ports 'lan3'
	list ports 'lan4'
```

This causes that, on recent OpenWrt releases, the `bat0` interface gets added to the bridge in the old way by `proto/lan.lua` from `lime-system`, that does not work with DSA-style configuration:
https://github.com/libremesh/lime-packages/blob/06399e5de7fb7b9d3cf13c19f7bf70accb11299a/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua#L59

ilario:
Maybe this is already fixed. To be checked.

Issue #951, opened 2022-11-30
Adapting LibreMesh to OpenWrt 23.05


ilario:
Up to now, LibreMesh works with OpenWrt 19.07 and 18.06.
The tests on OpenWrt 21.02 and 22.03 were not very successful yet.
Some comments from the mailing list can be read here: (mailing list server is down, adding once it gets up again)

Very likely, we will have to adapt to support DSA way to configuring switches rather than the traditional swconfig, see the last part of this comment https://github.com/libremesh/lime-packages/issues/944#issue-1437483339

Here I report the outcome of lime-config configurations when running them on OpenWrt 19.07 and on OpenWrt 22.03.
At a first sight, some differences in /etc/config/network can be spotted and are likely the source of our problems.

**On OpenWrt 19.07**

Output of lime-report on clean OpenWrt 19.07:

[OpenWrt19-plain.txt](https://github.com/libremesh/lime-packages/files/10121271/OpenWrt19-plain.txt)

Console output during the lime-system installation:

[install-lime-system.txt](https://github.com/libremesh/lime-packages/files/10121277/install-lime-system.txt)

Output of lime-report on OpenWrt 19.07 after installing lime-system and rebooting:

[OpenWrt19-with_lime-system.txt](https://github.com/libremesh/lime-packages/files/10121282/OpenWrt19-with_lime-system.txt)

**On OpenWrt 22.03**

Output of lime-report on clean OpenWrt 22.03:

[OpenWrt22-plain.txt](https://github.com/libremesh/lime-packages/files/10121287/OpenWrt22-plain.txt)

Console output during the lime-system installation:

[install-lime-system.txt](https://github.com/libremesh/lime-packages/files/10121286/install-lime-system.txt)

Output of lime-report on OpenWrt 22.03 after installing lime-system and rebooting:

[OpenWrt22-with_lime-system.txt](https://github.com/libremesh/lime-packages/files/10121289/OpenWrt22-with_lime-system.txt)

ilario:
Here go more outputs from lime-report when installing more packages:

**On OpenWrt 19.07**

Output of lime-report on OpenWrt 19.07 after installing lime-system and lime-proto-anygw:

[OpenWrt19-with_lime-system_lime-proto-anygw.txt](https://github.com/libremesh/lime-packages/files/10206154/OpenWrt19-with_lime-system_lime-proto-anygw.txt)

[...]

PR #959, opened 2022-12-11
Replace OpenWrt 19.07 switch config style with OpenWrt 21.02 one in proto-lan and network.lua's device parser

ilario:
See #951 

Specifically, here I tried to address two big differences that can be spotted in the reports attached in #951:

* the fact that with current code and when DSA is present, bat0 interface was not being added to br-lan bridge, see #958
* the fact that with current code and when DSA is present, eth0 was selected for addition to the bridge
* the fact that with current code and when DSA is present, the VLAN interfaces from the ethernet ports were not added inside bat0

On the last point, there are two big doubts:

* with DSA, eth0.1 does not exist anymore and has been replaced by lan1, lan2, lan3... So which interfaces should be created and added to bat0? I decided to create a VLAN for each of these, for example lan1_29, lan2_29, lan3_29...
* why the WAN interface was added to bat0? With the current code, there are interfaces like eth0-2_29 (a VLAN created on top of the WAN) being added to bat0, but this sounds useless and unsafe to me. So I did not add WAN interface to bat0, but only LAN ones.

I coarsely tested on a single router, so this PR is not ready to merge at all.

Indeed, what I observed is that only lan1_29 gets added to bat0, while lan2_29... do not get added, even if in `/etc/config/network` they are all indicated to be part of bat0 in the same way. No idea what is happening here.

### DIFF
--- a/packages/lime-hwd-openwrt-wan/files/usr/lib/lua/lime/hwd/openwrt_wan.lua
+++ b/packages/lime-hwd-openwrt-wan/files/usr/lib/lua/lime/hwd/openwrt_wan.lua
@@ -1,5 +1,13 @@
 #!/usr/bin/lua
 
+--! LibreMesh community mesh networks meta-firmware
+--!
+--! Copyright (C) 2014-2023  Gioacchino Mazzurco <gio@eigenlab.org>
+--! Copyright (C) 2023  Asociación Civil Altermundi <info@altermundi.net>
+--!
+--! SPDX-License-Identifier: AGPL-3.0-only
+
+
 local hardware_detection = require("lime.hardware_detection")
 local config = require("lime.config")
 local utils = require("lime.utils")
@@ -22,7 +30,7 @@ function openwrt_wan.detect_hardware()
 		if networkTable then
 			local wanTable = networkTable['wan']
 			if wanTable then
-				ifname = wanTable['ifname']
+				ifname = wanTable['device']
 			end
 		end
 		if ifname and ifname ~= "" then
@@ -47,8 +55,10 @@ function openwrt_wan.detect_hardware()
 			config.set(openwrt_wan.sectionName, "protocols", protos)
 			config.set(openwrt_wan.sectionName, "linux_name", ifname)
 			config.end_batch()
+
+			utils.dbg("WAN interface:", ifname)
 		else
-			utils.log("No wan interface detected")
+			utils.dbg("WAN interface not found")
 		end
 	end
 end
--- a/packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua
+++ b/packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua
@@ -1,5 +1,12 @@
 #!/usr/bin/lua
 
+--! LibreMesh community mesh networks meta-firmware
+--!
+--! Copyright (C) 2014-2023  Gioacchino Mazzurco <gio@eigenlab.org>
+--! Copyright (C) 2023  Asociación Civil Altermundi <info@altermundi.net>
+--!
+--! SPDX-License-Identifier: AGPL-3.0-only
+
 local libuci = require("uci")
 local fs = require("nixio.fs")
 local utils = require("lime.utils")
@@ -20,7 +27,7 @@ end
 
 function wan.setup_interface(ifname, args)
 	local uci = libuci:cursor()
-	uci:set("network", "wan", "ifname", ifname)
+	uci:set("network", "wan", "device", ifname)
 	uci:save("network")
 
 	if utils.is_installed('firewall') then
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -228,6 +228,16 @@ function network.scandevices()
 			utils.log( "network.scandevices.dev_parser found vlan device %s " ..
 			           "and marking %s as nobridge", dev, rawif )
 		end
+		--! With DSA, the LAN ports are not anymore eth0.1 but lan1, lan2...
+		if dev:match("^lan%d+$") then
+			local lower_if_path = utils.unsafe_shell("ls -d /sys/class/net/" .. dev .. "/lower*")
+			local lower_if_table = utils.split(lower_if_path, "_")
+			local lower_if = lower_if_table[#lower_if_table]:gsub("\n", "")
+			devices[lower_if] = { nobridge = true }
+			devices[dev] = {}
+			utils.log( "network.scandevices.dev_parser found LAN port %s " ..
+			           "and marking %s as nobridge", dev, lower_if )
+		end
 
 		if dev:match("^wlan%d+"..wireless.wifiModeSeparator.."%w+$") then
 			devices[dev] = {}
@@ -260,6 +270,14 @@ function network.scandevices()
 		           created_device or "not_found")
 		dev_parser(created_device)
 		dev_parser(base_interface)
+		--! With DSA switch config, lan* ports are included in br-lan as "ports"
+		local ports = section["ports"]
+		for _,port in pairs(ports) do
+		utils.log( "network.scandevices.owrt_device_parser found "..
+			   "interface %s with port %s",
+			   created_device or "not_found", port or "not_found")
+			   dev_parser(port)
+		end
 	end
 
 	function owrt_switch_vlan_parser(section)
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -1,5 +1,12 @@
 #!/usr/bin/lua
 
+--! LibreMesh community mesh networks meta-firmware
+--!
+--! Copyright (C) 2013-2023  Gioacchino Mazzurco <gio@eigenlab.org>
+--! Copyright (C) 2023  Asociación Civil Altermundi <info@altermundi.net>
+--!
+--! SPDX-License-Identifier: AGPL-3.0-only
+
 network = {}
 
 local ip = require("luci.ip")
@@ -452,6 +459,7 @@ function network.createVlanIface(linuxBaseIfname, vid, openwrtNameSuffix, vlanPr
 		uci:set("network", owrtDeviceName, "device")
 		uci:set("network", owrtDeviceName, "type", vlanProtocol)
 		uci:set("network", owrtDeviceName, "name", linux802adIfName)
+		--! This is ifname also on current OpenWrt
 		uci:set("network", owrtDeviceName, "ifname", linuxBaseIfname)
 		uci:set("network", owrtDeviceName, "vid", vlanId)
 	end
@@ -468,7 +476,7 @@ function network.createVlanIface(linuxBaseIfname, vid, openwrtNameSuffix, vlanPr
 	--! ifname in network because it is already set in wireless, because
 	--! setting ifname on both places cause a netifd race condition
 	if vid ~= 0 or not linux802adIfName:match("^wlan") then
-		uci:set("network", owrtInterfaceName, "ifname", linux802adIfName)
+		uci:set("network", owrtInterfaceName, "device", linux802adIfName)
 	end
 
 	uci:save("network")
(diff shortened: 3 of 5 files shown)


ilario:
> Indeed, what I observed is that only lan1_29 gets added to bat0, while lan2_29... do not get added, even if in `/etc/config/network` they are all indicated to be part of bat0 in the same way.

Just thought (will test ASAP): maybe only lan1_29 was added because it was the only Ethernet port with an active link (to my laptop). So maybe the others would get added as soon as the lower lanX interface gets up.

ilario:
I can confirm that, with the current state of this PR, it works on OpenWrt 19.07.
Here you have the lime-report for a router running OpenWrt 19.07, and some LibreMesh packages from this pull request:
[OpenWrt19-with_lime-system_lime-proto-anygw-batadv-babeld.txt](https://github.com/libremesh/lime-packages/files/10257641/OpenWrt19-with_lime-system_lime-proto-anygw-batadv-babeld.txt)

and you can compare it with the one obtained without this PR here: https://github.com/libremesh/lime-packages/issues/951#issuecomment-1346095002

G10h4ck:
I don't have DSA devices at hand, @ilario can you test it with a few of them and see if everything looks good?

ilario:
I thought I had more of the supported ones, but actually I just have two (I will check if some more, but if any they will be very old):

- [YouHua WR1200JS](https://openwrt.org/toh/youhua/wr1200js)
- [Plasma Cloud PA1200](https://openwrt.org/toh/hwdata/plasma_cloud/plasma_cloud_pa1200)

I tested on the first (both with OpenWrt 22.03 and 19.07) but not yet on the second (I started testing on Ubiquiti NanoStation M5 XM and NS Loco M2 XM and was going to test on TP-Link TL-WDR3600 before realizing they were not ported yet to DSA, maybe I could compile master including the [pull request](https://github.com/openwrt/openwrt/pull/4622) you mentioned in [the meeting](https://lists.autistici.org/message/20230104.145426.57d35257.en.html)).

G10h4ck:
Having a mixed setup of DSA and non-DSA routers is ideal for testing this PR, all must work fine on all of them, at that point we can merge it

ilario:
Just tested on OpenWrt 22.03 on a non-DSA device (Ubiquiti NanoStation M5 XW) and the `/etc/config/network` file is quite similar to the DSA case, not to the OpenWrt 19.07 case.
I didn't expect that, but obviously it makes sense.
So, instead of checking for DSA - not-DSA we should check for pre-22.03 or 22.03.

Considering this mess, and the need to restructure everything related to the firewall (OpenWrt 22.03 replaced iptables and ebtables with nftables, which means that installing a simple package like lime-proto-wan can pulll a ton of useless dependencies) I would suggest to not support anything older than OpenWrt 22.03 for the next release.
This would also mean creating a branch for compiling on OpenWrt 19.07 (for example it can be named `2020.1`), using the master branch for developing the next release and indicating in the documentation that for compiling an image the `2020.1` branch should be used.

ilario:
> So, instead of checking for DSA - not-DSA we should check for pre-22.03 or 22.03.

This is not difficult, eh, I can modify the PR if we consider that is too early for abandoning OpenWrt 19.07...

G10h4ck:
> Just tested on OpenWrt 22.03 on a non-DSA device (Ubiquiti NanoStation M5 XW) and the `/etc/config/network` file is quite similar to the DSA case, not to the OpenWrt 19.07 case. I didn't expect that, but obviously it makes sense. So, instead of checking for DSA - not-DSA we should check for pre-22.03 or 22.03.
> 
> Considering this mess, and the need to restructure everything related to the firewall (OpenWrt 22.03 replaced iptables and ebtables with nftables, which means that installing a simple package like lime-proto-wan can pulll a ton of useless dependencies) I would suggest to not support anything older than OpenWrt 22.03 for the next release. This would also mean creating a branch for compiling on OpenWrt 19.07 (for example it can be named `2020.2` practically creating a new release following the last one `2020.1`), using the master branch for developing the next release and indicating in the documentation that for compiling an image the `2020.2` branch should be used.

With so many changes what IMHO should we do, is decide which OpenWrt version is our development target, if we decide to target a new non-compatible version we should tag/branch the last commit compatible with the previous version as target_OpenWrt_XX.XX, if we use a tag it means that commit is the last supporting that OpenWrt version carved in stone, aka absolutely no more works gets there, if we branch we remain open to the possibility to accept some very very very well tested bug fix commit ending up there.

G10h4ck:
> Ok, anyway the firewall will block it, so ok.

Also because of that it should not be installed :-p

ilario:
> With so many changes what IMHO should we do, is decide which OpenWrt version is our development target

Yep, after a very short discussion on the Element chat, it was decided to go on supporting only OpenWrt 22.03 on the master branch.

> if we decide to target a new non-compatible version we should tag/branch the last commit compatible with the previous version as target_OpenWrt_XX.XX, if we use a tag it means that commit is the last supporting that OpenWrt version carved in stone, aka absolutely no more works gets there, if we branch we remain open to the possibility to accept some very very very well tested bug fix commit ending up there.

Yep, I think that a branch is better.
What do you think about making a minor release like `2020.2` with a tag and a `target_OpenWrt_19.07` branch for including bug fixes? There are plenty of trivial things that as soon as get fixed can go also there as they affect OpenWrt 22.03 and 19.07 in the same way.

>> Ok, anyway the firewall will block it, so ok.
> Also because of that it should not be installed :-p

Let's keep the discussion on https://github.com/libremesh/libremesh.github.io/pull/139 (or https://github.com/libremesh/lime-packages/issues/280).

ilario:
> > ```
> > * why the WAN interface was added to bat0? With the current code, there are interfaces like eth0-2_29 (a VLAN created on top of the WAN) being added to bat0, but this sounds useless and unsafe to me. So I did not add WAN interface to bat0, but only LAN ones.
> > ```
> 
> It is not useless, and not dangerous, please keep this behaviour unchanged, so keep the wan interface used for mesh too by default

Even if I include `wan` there, WAN-WAN connection will be broken anyway as it is currently broken (as currently the WAN port gets included in br-lan, and so I still don't understand why people use that WAN-WAN thing).

The main reason why I need to include `wan` here, is because otherwise the lime-proto-wan and lime-hwd-openwrt-wan packages will not configure the WAN port, I suspect...? I should try... can anyone confirm?

So the possible solutions are 2, that I can think of:

1. we add wan here.
Then, for people willing to do WAN-WAN connections, we document how to manually remove WAN from br-lan and take down the firewall rules blocking incoming connections on WAN (main ticket https://github.com/libremesh/lime-packages/issues/976);
1. we add wan here and we modify lan.lua for recognizing the WAN port and avoiding to add it to the br-lan bridge, something like this https://github.com/libremesh/lime-packages/blob/f82f3a9f477a3f3ff78f94c07d2b8a0647d31d5d/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua#L46
but it would be easy just with DSA as the WAN interface is clearly named `wan`. In swconfig it would be a mess, ideas on clean ways for doing that @G10h4ck?
Then, for people willing to do WAN-WAN connections, we document how to take down the firewall rules blocking incoming connections on WAN  (main ticket https://github.com/libremesh/lime-packages/issues/976).

ilario:
> Even if I include `wan` there, WAN-WAN connection will be broken anyway as it is currently broken (as currently the WAN port gets included in br-lan, and so I still don't understand why people use that WAN-WAN thing).

Maybe I observed that just because I didn't have lime-hwd-openwrt-wan installed, as otherwise it should create specific configuration for the WAN interface and not include it in the lan bridge. If this is the case, sorry, I thought it was a general behavior.

Anyway, the complete absence of users commenting on the wan-wan use-case make me think that we are struggling to support something that nobody uses.

ilario:
What is remaining to do:

* update the tests

And I also kept the DSA checker even if it is not being used anymore in this PR, supposing that is going to be useful in other occasions.

ilario:
In order to be able to update the tests, I would need someone to prepare a Docker image based on OpenWrt 22.03 to be uploaded on Docker Hub https://hub.docker.com/u/libremesh who can do that? @spiccinini @altergui

G10h4ck:
I think we should keep this whole PR for after the retargetting, so please keep de DSA detection code as it will be useful

G10h4ck:
I have been pushing some improvements here https://github.com/G10h4ck/lime-packages/tree/test_PR_959

ilario:
I saw that you implemented the usage of `device` additionally to that of `ifname`.
I think we can directly remove the usage of `ifname` as I believe that since OpenWrt 21.02 it has been renamed to `device`, see https://openwrt.org/releases/21.02/notes-21.02.0#new_network_configuration_syntax_and_boardjson_change

G10h4ck:
Thanks for suggestion and for pointing out documentation, I have fixed it as you suggested

ilario:
Wait, before merging this I think we need to do the following:

* create a branch for compiling on OpenWrt 19.07
* write on the development page of the website that the compilation should be done using that branch
* write the same thing on the mailing list and on the Element group

Do you agree?

aparcar:
So this is ready to merge now?

ilario:
Yes, I think it is ready to merge. I could not test @G10h4ck 's commits yet but he surely did.
Anyway I would leave a couple of grace days for the community to read the announcement email https://lists.autistici.org/message/20230323.212314.230ce707.en.html as this pull request will break their builds otherwise.

G10h4ck:
Yeah I have tested the commits, those are not enough to run properly libremesh on top of newer OpenWrt, and hence as @ilario said the resulting build will be broken, but I am already working on more stuff to push in another PR soon that should fix at least what I have detected so far.

aparcar:
WAN isn't detect on my device...
```
@/usr/lib/lua/lime/hwd/openwrt_wan.lua:53 detect_hardware WAN interface: wan
network.scandevices.owrt_ifname_parser found ifname wlan0-ap
network.scandevices.dev_parser found WiFi device wlan0-ap
network.scandevices.owrt_ifname_parser found ifname wlan0-apname
network.scandevices.dev_parser found WiFi device wlan0-apname
network.scandevices.owrt_ifname_parser found ifname wlan0-mesh
network.scandevices.dev_parser found WiFi device wlan0-mesh
network.scandevices.owrt_ifname_parser found ifname wlan1-ap
network.scandevices.dev_parser found WiFi device wlan1-ap
network.scandevices.owrt_ifname_parser found ifname wlan1-apname
network.scandevices.dev_parser found WiFi device wlan1-apname
network.scandevices.owrt_ifname_parser found ifname wlan1-mesh
network.scandevices.dev_parser found WiFi device wlan1-mesh
network.scandevices.owrt_device_parser found base interface not_found and derived device br-lan
network.scandevices.dev_parser got nil device
network.scandevices.owrt_device_parser found interface br-lan with port bat0
network.scandevices.owrt_device_parser found interface br-lan with port lan1
network.scandevices.dev_parser found LAN port lan1 and marking eth0 as nobridge
network.scandevices.owrt_device_parser found interface br-lan with port lan4
network.scandevices.dev_parser found LAN port lan4 and marking eth0 as nobridge
network.scandevices.owrt_device_parser found interface br-lan with port lan3
network.scandevices.dev_parser found LAN port lan3 and marking eth0 as nobridge
network.scandevices.owrt_device_parser found interface br-lan with port lan2
network.scandevices.dev_parser found LAN port lan2 and marking eth0 as nobridge
network.scandevices.dev_parser found plain Ethernet device eth0
```

Is it possible that the wan interface isn't scanned at all?

ilario:
Seems that lime-hwd-openwrt-wan is working correctly.
But for some reason, the wan.configure() from lime-proto-wan does not get called.

G10h4ck:
> WAN isn't detect on my device...
> 
> ```
> @/usr/lib/lua/lime/hwd/openwrt_wan.lua:53 detect_hardware WAN interface: wan
> ```
> 
> Is it possible that the wan interface isn't scanned at all?

The interface is correctly scanned and it is called `wan` maybe it get filtered out later in the code, I remember we have regexp filter to match ethernet interfaces that before were called ethX, so maybe some of those filters is silently dropping the interface at some point, I don't have a DSA based device at hand so if you can debug where the wan interface get discarded would be good

PR #966, opened 2023-01-03
unstuck-wifi: send SIGTERM to iw-processes still running after 5 minutes

### LINKED ISSUE #964: wifi-unstuck gets stuck
Processes like "iw dev wlan1-ap scan freq 5180 5240" are slowly accumulating on some routers. They're run by wifi-unstuck-wa which is run by cron every 10 min. The iw scan command seems to get stuck occasionally. According to @spiccinini, it is caused by a driver issue that hangs the scan.

pony1k:
Modification of wifi_unstuck_wa.lua so that it sends SIGTERM to all iw-processes still running after 5 minutes.
Fixes #964

### DIFF
--- a/packages/wifi-unstuck-wa/files/usr/lib/lua/wifi_unstuck_wa.lua
+++ b/packages/wifi-unstuck-wa/files/usr/lib/lua/wifi_unstuck_wa.lua
@@ -8,6 +8,9 @@ local wu = {}
 wu.FREQ_2GHZ_LIST = "2412 2462"
 wu.FREQ_5GHZ_LIST = "5180 5240"
 
+-- if iw runs for 5 min, it is likely hanging
+wu.TIMEOUT = 300
+
 function wu.get_stickable_ifaces()
 	local uci = config.get_uci_cursor()
 	local ifaces = {}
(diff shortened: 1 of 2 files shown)


ilario:
Thanks!!!
I did not try to understand how the code works but as far as I could test it works (I just replaced the command with a sleep).
Amazing!!

ilario (approved):

packages/wifi-unstuck-wa/files/usr/lib/lua/wifi_unstuck_wa.lua
```diff
@@ -28,8 +31,48 @@ function wu.get_stickable_ifaces()
...
+	end
+
+	repeat
+		-- wait for 100ms
+		nixio.nanosleep(0,100e6)
+
+		-- see if something changed
+		while true do
+			pid,state,code = nixio.waitpid(nil,"nohang")
+			if not pid then break end
```
ilario, line 48: From the documentation on https://neopallium.github.io/nixio/modules/nixio.html#nixio.waitpid I don't understand when `pid` is going to be `nil`. There it only says that pid can be zero, but in Lua zero in an if sentence is a true, so I don't understand how this works. But it works!

ilario:
For me this is good for merge.
@spiccinini @G10h4ck @nicoechaniz @selankon do you want to check this or can be merged?

packages/wifi-unstuck-wa/files/usr/lib/lua/wifi_unstuck_wa.lua
pony1k: That's because it returns `nil,nil,nil` if no child has changed state, despite documentation says otherwise. Anyway, one shouldn't rely on this. It will improve it.

ilario (approved):

G10h4ck (approved):

PR #971, opened 2023-01-18
dnsmasq move confdir setting for ujail

### LINKED ISSUE #970: Dnsmasq on OpenWrt 22.03 does not like conf-dir line in /etc/dnsmasq.conf
A line gets added in /etc/dnsmasq.conf here:
https://github.com/libremesh/lime-packages/blob/15aa6d2636f9dc506ad7161cd13eca4ea0056a44/packages/lime-system/files/usr/lib/lua/lime/network.lua#L161

And its presence somehow causes that dnsmasq does not start on OpenWrt 22.03.
The problem is caused by the usage of procd-ujail and should be possible to solve like this:
https://github.com/openwrt/openwrt/issues/9726#issuecomment-1198828327

ilario:
Fix #970
Still untested, need to be tested both on OpenWrt 19.07 and 22.03.


ilario:
Tested on OpenWrt 19.07 and it works.

Tested on OpenWrt 22.03 and it works.

BUT setting the configuration directory to be /etc/dnsmasq.d unsets the original configuration directory that was set to be /tmp/dnsmasq.d, so that this is not going to work anymore:
https://github.com/libremesh/lime-packages/blob/63630ba7e2818e2147d3477a8f33aa9d97b7dbd7/packages/batman-adv-auto-gw-mode/files/etc/hotplug.d/net/99-batman-gw#L13-L20

In OpenWrt 19.07 this is easy to fix as we can simply have this one directory as confdir and the other one executed in dnsmasq.conf, but in OpenWrt 22.03 only one confdir is allowed by procd-ujail for accessing.

@altergui do you think it would be ok to move that stuff to /etc/dnsmasq.d rather than /tmp/dnsmasq.d? How often do we expect this thing to be written? Every time the watchping ways that the connection is up or down?
In this case it would be too often for the flash memory for the router.
So we could place the `disable-dhcp-server-in-br-lan` file in /var/run/dnsmasq/ that, even if is not the right place, is included in the list of places accepted by ujail and is in the RAM.
Anyway there are other issues with that package: #972 so I am not sure if it is worth to fix it or if it needs a deep restructuring.

altergui:
> @altergui do you think it would be ok to move that stuff to /etc/dnsmasq.d rather than /tmp/dnsmasq.d? How often do we expect this thing to be written? Every time the watchping ways that the connection is up or down?

every time the gateway node goes up or down, this hotplug action is triggered on the rest of the nodes. gateway may go down because watchping detects internet connection is down, or also because of power loss for example. in stable conditions this happens only once per boot, let's say once per day. but if internet is flaky, it might get triggered every few minutes (or every minute, max, since that's watchping interval)

altergui:
> So we could place the `disable-dhcp-server-in-br-lan` file in /var/run/dnsmasq/ that, even if is not the right place, is included in the list of places accepted by ujail and is in the RAM.

to me, it does sound like the right place, but i'm totally out of context since i haven't been following openwrt nor libremesh development for years.

ilario:
> > So we could place the `disable-dhcp-server-in-br-lan` file in /var/run/dnsmasq/ that, even if is not the right place, is included in the list of places accepted by ujail and is in the RAM.
> 
> to me, it does sound like the right place, but i'm totally out of context since i haven't been following openwrt nor libremesh development for years.

Ok, I can move it there.
Opinions @G10h4ck @spiccinini @selankon ?

ilario:
I moved the dnsmasq configuration file from `batman-adv-auto-gw-mode` to `/var/run/dnsmasq/disable-dhcp-server-in-br-lan`.
Still to be tested.

ilario:
In order to get the `/var/run/dnsmasq/disable-dhcp-server-in-br-lan` file be considered by dnsmasq, I had to include a link to it in the dnsmasq config folder, but then dnsmasq was failing to start if the link's target was missing.
So then I added an init script creating the empty file in `/var/run/dnsmasq` before dnsmasq starts and replaced the removal of the file with the overwriting of it.

I tested it on OpenWrt 19.07 and got blocked by this https://github.com/libremesh/lime-packages/issues/972#issuecomment-1427115440 but apart from that seems working.

ilario:
In https://github.com/libremesh/lime-packages/issues/972#issuecomment-1440513185 I proposed to eliminate [batman-adv-auto-gw-mode](https://github.com/libremesh/lime-packages/tree/06399e5de7fb7b9d3cf13c19f7bf70accb11299a/packages/batman-adv-auto-gw-mode), if this happens, some commits from here will have to be removed.

ilario:
Closing in favor of #979

PR #973, opened 2023-01-23
Add meuno.info to anygw for portuguese acessibility

luandro:
Currently lime-app is only acessible through thisnode.info and minodo.info, neither make sense for portuguese speakers. Since it seems an easy update and the domain is very cheap, I'd like to start this process, which will make our lives in Brasil much easier when supporting CN processes, by making addresses acessible.

What would be the next steps? Buy the domain? Does anything need to be done with it?

### DIFF
--- a/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
+++ b/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
@@ -11,7 +11,7 @@ anygw = {}
 anygw.configured = false
 
 anygw.SAFE_CLIENT_MTU = 1350
-anygw.FQDN = {"thisnode.info", "minodo.info"}
+anygw.FQDN = {"thisnode.info", "minodo.info", "meuno.info"}
 
 function anygw.configure(args)
 	if anygw.configured then return end


ilario:
Hi!
Looking in the code, that seems to be the only important place for meuno.info to be added, so I approve the pull request :)
There is only another place in Pirania that I have no idea if it is relevant also:
https://github.com/libremesh/lime-packages/blob/1bff5dc33002dcdc9205fc624a5ce02a1dbcac85/packages/pirania/files/etc/config/pirania#L4

ilario (approved):
Seems good. Consider buying that domain for hosting guidance information for users who did not configure well their connection to the mesh (e.g. if you have an internet connection via wifi with your laptop and the connection to the router via cable, it is likely that your laptop is going to look for meuno.info online rather than locally. So the content of this webpage on the internet would be something like "You have to disconnect from the internet, leave active only the connection to the router").

luandro:
Simply adding this to my router didn't make `meuno.info` start working. Is there something else missing?

ilario:
Uh!!!! But minodo.info works???
That variable is being used here:
https://github.com/libremesh/lime-packages/blob/8de4eb1f36406ac42132c69cc708d49a72c4193e/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua#L119
Can you copy and paste here the content of `/etc/config/dhcp` after running `lime-config`?

luandro:
I had rebooted and thought that would have been enough, but had to actually do a `lime-config` and now `meuno.info` is showing on `config/dhcp` under `config hostrecord 'anygw_dns'`. Doing a `/etc/init.d/dnsmasq restart` made the dns start working. So confirming this works. Thanks @ilario!

ilario:
Great :D
Thank you indeed!

PR #975, opened 2023-01-27
shared-state-publish_dnsmasq_leases recognize IPv6 when IPv6 leases are present

### LINKED ISSUE #969: shared-state-publish_dnsmasq_leases errors
When I run `shared-state-publish_dnsmasq_leases` it errors with:
```
/usr/bin/lua: /usr/bin/shared-state-publish_dnsmasq_leases:48: table index is nil
stack traceback:
        /usr/bin/shared-state-publish_dnsmasq_leases:48: in function 'add_dhcp_leases_from_files'
        /usr/bin/shared-state-publish_dnsmasq_leases:130: in main chunk
        [C]: ?
```
It is the following lines:
```
_, client_mac, client_ip, client_hostname, client_id = unpack(line:split(" "))
leasesTable[client_ip] = { hostname=client_hostname, mac=client_mac }
```
which have trouble because of this line in the `/tmp/dhcp.leases`-file:
```
duid 00:01:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx:xx`
```
I did a quick research. According to [here](https://dnsmasq-discuss.thekelleys.org.narkive.com/ia9YNLmE/dnsmasq-leases-file-format-specification), this is normal when DHCPv6 is served.

ilario:
Fix #969
I still have to test the modified code.
@pony1k can you test this?

### DIFF
--- a/packages/shared-state-dnsmasq_leases/files/usr/bin/shared-state-publish_dnsmasq_leases
+++ b/packages/shared-state-dnsmasq_leases/files/usr/bin/shared-state-publish_dnsmasq_leases
@@ -71,11 +71,16 @@ function add_dhcp_leases_from_files()
 end
 
 function add_lease(client_mac, client_ip, client_hostname)
-	local leaseData = {}
-	leaseData[client_ip] = { hostname=client_hostname, mac=client_mac }
-	local cStdin = io.popen("shared-state insert dnsmasq-leases", "w")
-	cStdin:write(JSON.stringify(leaseData))
-	cStdin:close()
+	-- avoid adding IPv6 as collisions are very unlikely
+	-- if we were going to include it, we should prepend "id:" to the duid received as first argument
+	-- see https://github.com/libremesh/lime-packages/pull/975#issuecomment-1413984349
+	if client_ip:match('^%d+%.%d+%.%d+%.%d+$') then
+		local leaseData = {}
+		leaseData[client_ip] = { hostname=client_hostname, mac=client_mac }
+		local cStdin = io.popen("shared-state insert dnsmasq-leases", "w")
+		cStdin:write(JSON.stringify(leaseData))
+		cStdin:close()
+	end
 
 	if type(client_hostname) == "string" and client_hostname:len() > 1 then
 		local hostTable = {}
--- a/packages/shared-state-dnsmasq_leases/files/usr/bin/shared-state-publish_dnsmasq_leases
+++ b/packages/shared-state-dnsmasq_leases/files/usr/bin/shared-state-publish_dnsmasq_leases
@@ -53,9 +53,11 @@ function add_dhcp_leases_from_files()
 			if client_hostname:len() > 1 then
 				hostTable[client_ip.." "..client_hostname] = true
 			end
-			-- IPv6 lines have a number instead of the MAC address
+			-- avoid adding IPv6 as collisions are very unlikely
+			-- if we were going to include it, we should remember that
+			-- IPv6 lines have a number here instead of the MAC address
 			-- https://github.com/libremesh/lime-packages/issues/969
-			if not tonumber(client_mac) then
+			if client_ip:match('^%d+%.%d+%.%d+%.%d+$') then
 				leasesTable[client_ip] = { hostname=client_hostname, mac=client_mac }
 			end
 		end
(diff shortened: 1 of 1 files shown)


ilario:
Just realized that the if statements can be rearranged in a more useful way, will improve in a few days.

ilario:
Now the if sentences are more useful.
@pony1k can you test if this works?
Thanks!!!

packages/shared-state-dnsmasq_leases/files/usr/bin/shared-state-publish_dnsmasq_leases
```diff
@@ -45,9 +45,14 @@ function add_dhcp_leases_from_files()
 
 	for line in io.lines("/tmp/dhcp.leases") do
 		_, client_mac, client_ip, client_hostname, client_id = unpack(line:split(" "))
-		leasesTable[client_ip] = { hostname=client_hostname, mac=client_mac }
-
-		if type(client_hostname) == "string" and client_hostname:len() > 1 then
+		-- client_ip will be nil in duid lines
+		-- https://dnsmasq-discuss.thekelleys.org.narkive.com/ia9YNLmE/dnsmasq-leases-file-format-specification
+		if client_ip and type(client_hostname) == "string" and client_hostname:len() > 1 then
```
pony1k, line 50: The dhcp.leases file contains lines with hostname `*`, i.e.
```
1675107518 12:34:65:3d:ce:2b 10.51.147.128 * 01:12:34:65:3d:ce:2b
1675106752 ab:cd:15:8a:62:93 10.51.147.146 * 01:ab:cd:15:8a:62:93
```
It is correct that we don't want to add `*` into the `hostTable`. But I think we still want to add them into the `leasesTable`, which will not happen with this arrangement of if-statements.

ilario: Ah ok, it makes sense, thanks! I am going to correct this today.

G10h4ck:
looks ok, please test it both with `anygw` and without, if it works well we can merge it

pony1k:
I have tested it on different routers with and without anygw. It works fine except one thing: As stated in the man-page, when dnsmasq calls the script with `add` for an ipv6-lease, it will pass the duid of the host instead of the mac-address. The duid then appears in the `/tmp/dhcp.hosts_remote`-file in other nodes, in the place where the mac-address would be. The man-page of dnsmasq states that in this case, the duid should be prefixed with `id:`. I.e.
`
id:00:01:00:01:29:9d:9c:12:00:4c:e0:ab:cd:ef,[fd43:1508:3393:fe00::d312]
`
instead of
`
00:01:00:01:29:9d:9c:12:00:4c:e0:ab:cd:ef,[fd43:1508:3393:fe00::d312]
`

ilario:
@pony1k with the current code of this PR, that should not appear at all: a line like 
```
1674755842 1276192316 fd70:d3e4:df63:184d::ccbf katze 00:04:eb:cf:8f:90:ce:2c:e3:20:12:55:c6:fe:9b:c7:16:ee
```
does not get added to `leasesTable` which is then used for generating `/tmp/dhcp.hosts_remote` as it will match the `not tonumber(second field of the string that usually is the MAC)`.

But I think I got what you mean: that also this line should be added but using the last field prefixed with `id:`.
I can add that over the next days :)
Thanks!

pony1k:
> does not get added to `leasesTable` which is then used for generating `/tmp/dhcp.hosts_remote` as it will match the `not tonumber(second field of the string that usually is the MAC)`.

Yes it does. Look in the `add_lease`-function. This function gets called when the script is run with `add` as first argument. 

> But I think I got what you mean: that also this line should be added but using the last field prefixed with `id:`. I can add that over the next days :)

I'm not sure if it should. Maybe it's fine to not share ipv6-leases, since collisions are very unlikely. I haven't reflected on this.

ilario:
> > does not get added to `leasesTable` which is then used for generating `/tmp/dhcp.hosts_remote` as it will match the `not tonumber(second field of the string that usually is the MAC)`.
> 
> It does not get added to `leasesTable`, but it still ends up in the `dnsmasq-leases.json`. Look in the `add_lease`-function. This function gets called when the script is run with `add` as first argument.

Ooh, ok!
I will modify that also there then!
 
> > But I think I got what you mean: that also this line should be added but using the last field prefixed with `id:`. I can add that over the next days :)
> 
> I'm not sure if it should. Maybe it's fine to not share ipv6-leases, since collisions are very unlikely. I haven't reflected on this.

Meh, it shouldn't harm neither.

G10h4ck:
> I'm not sure if it should. Maybe it's fine to not share ipv6-leases, since collisions are very unlikely. I haven't reflected on this.

Yeah unless a very conflict prone assignation algorithm is used (never saw anything like that, plus it make no sense at all in a mesh network), no need to share IPv6 leases as the collision is so unlikely that also the RFC standards take that for granted. Most probably we didn't share those also with this motivation. As it is not needed, avoid sharing them is better, so the global state is lighter, less RAM is used, less CPU is used, less traffic is generated, everything works better.

ilario:
Ok, I added the filter also to the `add_lease` function, for leaving out the IPv6.
@pony1k @G10h4ck do you think is ok now?

G10h4ck:
Looks good to me, @pony1k can you test this and report back so we can merge it before re-targeting?

pony1k:
I have tested it. It works fine. Thank you!

G10h4ck (approved):

ilario:
Thank you both!

PR #978, opened 2023-03-07
Expose get_loss

selankon:
It expose the get_loss function for IPv4 ips. It will be used to do diagnosis of where the last internet path is broken, from the frontend on the new landing page. It just return ping loose to specific ip, example:

```bash
ubus call lime-metrics get_loss '{"target" : "1.1.1.4"}'
{
	"status": "ok",
	"loss": "0"
}
```

### DIFF
--- a/packages/ubus-lime-metrics/files/usr/lib/lua/lime-metrics.lua
+++ b/packages/ubus-lime-metrics/files/usr/lib/lua/lime-metrics.lua
@@ -35,6 +35,22 @@ function metrics.get_metrics(target)
     return result
 end
 
+function metrics.get_loss(target)
+    local result = {}
+    local node = target
+    local loss = nil
+
+    if lutils.is_installed("lime-proto-babeld") then
+        loss = utils.get_loss(node, 4)
+    else
+        return {status="error", error={msg="No lime-proto-babeld found", code="1"}}
+    end
+    result.loss = loss
+    result.status = "ok"
+    return result
+end
+
+
 function metrics.get_gateway()
     local result = {}
     local gw = nil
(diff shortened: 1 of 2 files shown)


ilario:
This works also without Babeld, no?
It is a simple ping, why does it look for lime-proto-babeld?
I tested it with IPv4 on LibreMesh built on OpenWrt 19.07 and it works :)

ilario:
Tested also on OpenWrt 22.03 and it also works there.
I still have to understand why it checks for the presence of Babeld...

selankon:
@G10h4ck probably is not needed. I just copy pasted the code on lime-metrics `metrics.get_metrics`: https://github.com/libremesh/lime-packages/blob/master/packages/ubus-lime-metrics/files/usr/lib/lua/lime-metrics.lua#L19

```lua
    if lutils.is_installed("lime-proto-bmx6") then
        loss = utils.get_loss(node..".mesh", 6)
        shell_output = lutils.unsafe_shell("netperf -6 -l 10 -H "..node..".mesh| tail -n1| awk '{ print $5 }'")
    elseif lutils.is_installed("lime-proto-babeld") then
        loss = utils.get_loss(node, 4)
        shell_output = lutils.unsafe_shell("netperf -l 10 -H "..node.."| tail -n1| awk '{ print $5 }'")
    else
        return {status="error", error={msg="No lime-proto-bmx6 or lime-proto-babeld found", code="1"}}
    end
```

Probably, it can work without this. And also, I'm thinking that maybe ipv6 has to be supported?

G10h4ck:
just remove the check then, let's keep the IPv6 stuff for another PR

selankon:
@G10h4ck  this should fix it https://github.com/libremesh/lime-packages/pull/978/commits/97646ddad63e87f5804099025f29a59ed55a98a2

selankon:
I don't know why the tests are failing and I can't re run the jobs

Issue #977, opened 2023-02-24
Archive abandoned packages


ilario:
We have many packages whose maintainer is not maintaining anymore.
Some are still working but some are outdated and broken and unused, and can be archived.

The proposal is to create a `packages-archived` folder and moving them there.

For example we can start from:

- lime-web https://github.com/libremesh/lime-packages/issues/897
- batman-adv-auto-gw-mode  https://github.com/libremesh/lime-packages/issues/972
- everything related to BMX6 and BMX7
- ...

ilario:
* adhoc from lime-system #855

ilario:
The previous cleaning was done in #576 and already there (in 2019) there were suspects also on:

- [ ]   [lime-ap-watchping](https://github.com/libremesh/lime-packages/tree/master/packages/lime-ap-watchping)
- [ ]   [lime-hwd-usbradio](https://github.com/libremesh/lime-packages/tree/master/packages/lime-hwd-usbradio)
- [ ]   [luci-app-batman-adv](https://github.com/libremesh/lime-packages/tree/master/packages/luci-app-batman-adv)
- [ ]   [luci-app-openairview](https://github.com/libremesh/lime-packages/tree/master/packages/luci-app-openairview)
- [ ]   [luci-lib-jquery-flot](https://github.com/libremesh/lime-packages/tree/master/packages/luci-lib-jquery-flot)

ilario:
Also related, removing stuff mentioned here #809

ilario:
- [ ] cotonete https://github.com/libremesh/lime-packages/tree/master/packages/cotonete

ilario:
#1097 by @G10h4ck is removing some of the obsolete packages.

Additionally to the aforementioned ones, another package to remove is:

- [ ] luci-mod-status (it uses swconfig that is not going to be present on some DSA router)

ilario:
> * everything related to BMX6 and BMX7

If I am not mistaken, at least @pony1k is using BMX7, so we should not remove that as it works.

On the contrary, BMX6 is not present anymore in the [openwrt-routing repository](https://github.com/openwrt/routing), so mentions to BMX6 should be deleted from other packages  (e.g. inside lime-system, lime-docs, lime-report) and from network-profiles (e.g. [canvies/temb-5GHz-BMX6](https://github.com/libremesh/network-profiles/blob/master/canvies/temba-5GHz-BMX6/Makefile)).

PR #979, opened 2023-03-10
dnsmasq move confdir setting for ujail, avoiding to fix batman-adv-auto-gw-mode

### LINKED ISSUE #970: Dnsmasq on OpenWrt 22.03 does not like conf-dir line in /etc/dnsmasq.conf

ilario:
Fix #970

This is identical to #971 but without the commits that would be needed for fixing batman-adv-auto-gw-mode.

batman-adv-auto-gw-mode is severely outdated and will likely be archived, see #977

Please consider merging.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -154,11 +154,11 @@ function network.setup_dns()
 			uci:set("dhcp", s[".name"], "expandhosts", "1")
 			uci:set("dhcp", s[".name"], "domainneeded", "1")
 			uci:set("dhcp", s[".name"], "server", resolvers)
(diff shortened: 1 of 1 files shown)


G10h4ck:
I suppose you have tested this again on both openwrt 19 and 22, did I understand correctly?

ilario:
Yes, see the comments on #971

G10h4ck (approved):

PR #980, opened 2023-03-14
Avoid random from busybox

### LINKED ISSUE #800: No value in $RANDOM
As stated here: [lime-users-maillinglist](https://lists.libremesh.org/pipermail/lime-users/2020-November/001802.html)
I found out, that in OpenWRT snapshot there is no value initialized for the Variable $RANDOM which is used in crontab for LiMe.

### LINKED ISSUE #949: unique_append should check if /etc/crontabs/root file exist before appending
The onliner function used often in the code for adding a line to the root crontab does not check for its existence before looking into it with grep. For example here:

https://github.com/libremesh/lime-packages/blob/5398a91f50abeb5988c56de41aec7bf20da361f8/packages/shared-state/files/etc/uci-defaults/shared-state-publishers-cron#L3-L10

This causes an error on the very first execution of such function:

```
Configuring shared-state.
grep: /etc/crontabs/root: No such file or directory
```

Nevertheless, the content gets added to the crontab anyway, so everything should work even when this error message appears.

ilario:
Here I joined the fix for 2 issues:
Fix #800 
Fix #949 
As both fixed happened mostly on the same files.
If you prefer so, I can split the PR in two different ones.
Also, this goes toward the possibility to use the OpenWrt ImageBuilder instead of having to make one specific for LibreMesh


ilario:
Just realized that the "unique" feature of the unique_append will stop working with this PR as the string will already include the random number embedded.

The alternative is to embed the whole command for generating the random number instead of defining the $randomNumber variable and then using it.

Opinions?

ilario:
Modified, now the random generator is included in the string to be appended, so that the unique_append can again check if the string is already present.

G10h4ck:
please split this in two PR

regarding $RANDOM removal, the proposed solution seems unpractical to me, favoring copy pasting of a not that readable snippet of code, better create a simple script in the path as an example `/usr/bin/randnum` containing that snippet and call it when usefull as `randnum`. Then randnum should be factored into it's own package

G10h4ck (changes requested):
See my comments in the discussion

G10h4ck:
Now that you tell it, AFAIR that was my conclusion a bunch of years ago too

pony1k:
We could also produce random numbers with awk. I.e. `awk "BEGIN{srand(); print int(rand()*100)}"` generates numbers in [0,100). It is easier to read than the /dev/urandom-hexdump-thing. Busybox-awk with math functions is included by default, I think.

ilario:
> We could also produce random numbers with awk. I.e. `awk "BEGIN{srand(); print int(rand()*100)}"` generates numbers in [0,100). It is easier to read than the /dev/urandom-hexdump-thing. Busybox-awk with math functions is included by default, I think.

Just tested. I can confirm that it is present in default OpenWrt and that it is more understandable (even for me that I know nothing of awk) than the hexdump command.
But it returns the same number if you run it more than once during the same second.
This would be a problem considering that this PR mainly edits files in `/etc/uci-defaults`, that likely gets executed one next to the other on the first boot (if the packages were compiled in or if you install more than one).

pony1k:
>This would be a problem considering that this PR mainly edits files in /etc/uci-defaults, that likely gets executed one next to the other on the first boot (if the packages were compiled in or if you install more than one).

Before, the random numbers were generated at the cron jobs run time. Why not keep it like this? This would also solve the problem with unique append. I was thinking this was addressed in the last commit. But now that I look at it more closely, it appears that still the random numbers are generated only once upon creation. So unique_append could add the cron job mutliple times with different random values.
So, my proposal would be, for example:
```
unique_append \
	'*/20 * * * * ((sleep $(awk "BEGIN{srand(); print int(rand()*600)}"); check-date-http &> /dev/null)&)'\
	/etc/crontabs/root
```
This litearally adds
```
*/20 * * * * ((sleep $(awk "BEGIN{srand(); print int(rand()*600)}"); check-date-http &> /dev/null)&)
```
to crontab. It is still not super readable though.

ilario:
> Before, the random numbers were generated at the cron jobs run time. Why not keep it like this? This would also solve the problem with unique append. I was thinking this was addressed in the last commit. But now that I look at it more closely, it appears that still the random numbers are generated only once upon creation. So unique_append could add the cron job mutliple times with different random values.

Oooooops, I thought I corrected that but you are right, the issue is still there.
Just force pushed fixing that.
Thanks @pony1k for noticing that!!!

> So, my proposal would be, for example:
> 
> ```
> unique_append \
> 	'*/20 * * * * ((sleep $(awk "BEGIN{srand(); print int(rand()*600)}"); check-date-http &> /dev/null)&)'\
> 	/etc/crontabs/root
> ```
> 
> This litearally adds
> 
> ```
> */20 * * * * ((sleep $(awk "BEGIN{srand(); print int(rand()*600)}"); check-date-http &> /dev/null)&)
> ```
> 
> to crontab. It is still not super readable though.

It is more readable than the hexdump version...

You are right that the problem I mentioned above (having all the same random sleep number) will not happen if the random command is embedded in the crontab line AND we differentiate the times at which they run (instead than many running at */5 we should have them running at */4, */5, */6...).
Can you prepare a pull request with that?
Otherwise I will do that.

In parallel, I am also preparing a pull request to propose to OpenWrt packages a command that does this number generation, as requested above from @G10h4ck, so that at some point in the future we will stop needing this.

pony1k:
> You are right that the problem I mentioned above (having all the same random sleep number) will not happen if the random command is embedded in the crontab line AND we differentiate the times at which they run (instead than many running at */5 we should have them running at */4, */5, */6...).
> Can you prepare a pull request with that?
> Otherwise I will do that.

Even if we put */4, */5, */6 and so on, at every full hour, it generates all random numbers at the same time. Maybe awk isn't a good solution after all.

ilario:
> Even if we put */4, */5, */6 and so on, at every full hour, it generates all random numbers at the same time. Maybe awk isn't a good solution after all.

Meh, good enough. I don't think that it is a problem if sometimes it happens at the same time... @G10h4ck, is it?

ilario:
More options:

* Add a "+123" in the command calculating the sleep time, with a different number for each script
* Replace the whole random thing with some number written by hand, different for each file, which is going to have few conflicts.

Do we really need it to be random?
For example, if we have:
*/4 sleep 12
*/5 sleep 23
*/6 sleep 45
There will be reasonably few times when the things happen at the same time, would this be enough?

ilario:
> Do we really need it to be random? For example, if we have: */4 sleep 12 */5 sleep 23 */6 sleep 45 There will be reasonably few times when the things happen at the same time, would this be enough?

I answer myself:
if we write a set number of seconds, there is the drawback that all the routers will run the same command at the same time. But I don't know if this is a problem.

G10h4ck:
> > Do we really need it to be random? For example, if we have: */4 sleep 12 */5 sleep 23 */6 sleep 45 There will be reasonably few times when the things happen at the same time, would this be enough?
> 
> I answer myself: if we write a set number of seconds, there is the drawback that all the routers will run the same command at the same time. But I don't know if this is a problem.

Yeah it is definitely a problem, and we experienced that in quintana-libre, imagine all routers attempting to sync at same time.

Please, as requested before, split the two utils in two different packages, and be careful that the random generator generate real random not same value if called ad similar time

Then we add those to libremesh repository, so we can use the already, meanwhile you propose them for openwrt inclusion

ilario:
> Please, as requested before, split the two utils in two different packages, 

@aparcar is dealing with the unique-append part in #988 
The random part is in #991

> and be careful that the random generator generate real random not same value if called ad similar time

Using /dev/urandom this problem should not happen

> Then we add those to libremesh repository, so we can use the already, meanwhile you propose them for openwrt inclusion

Proposed it in #991

PR #981, opened 2023-03-17
check-date-http improve error handling

### LINKED ISSUE #723: Error handling in check-date-http
Running check-date-http on a two nodes mesh with no internet connection results in this error:

```
/usr/bin/lua: /usr/bin/check-date-http:49: attempt to index local 'algo' (a number value)
stack traceback:
	/usr/bin/check-date-http:49: in main chunk
	[C]: ?
```

this means that checking for the `success` variable here 

https://github.com/libremesh/lime-packages/blob/50656f611171bde588ebe79cdb57a55777cedc3a/packages/check-date-http/files/usr/bin/check-date-http#L44-L45

is not enough to assure that the connection was good.

ilario:
Fix #723 
Added a check for the response of the HTTP request.
It was failing in the case of non-existing servers or non-reachable servers (e.g. `ip6only.me` on a non-IPv6 network).

I also took advantage of this PR for removing 
`http://stackoverflow.com/` from the servers list as it gave a `success` false.

In my opinion, we should remove also `ip6only.me` as I don't see any advantage of having a server that does not answer to IPv4 as we already have servers answering both IPv4 and IPv6.


ilario:
Tested all the servers on the list from a router with the LibreMesh packages installed on OpenWrt 19.07 and on an internet connection with only IPv4 (directly connected to the wan port). Tested adding a non-existing server name and tested in absence of internet connection (unplugging the wan cable).

G10h4ck:
> In my opinion, we should remove also `ip6only.me` as I don't see any advantage of having a server that does not answer to IPv4 as we already have servers answering both IPv4 and IPv6.

If I remember well IPv4-only and IPv6-only address have been explicitly to avoid problems with quirky networks setups which might hinder one or the other but not both, the default behaviour of DNS resolver and then upstream libraries should be to prefer IPv6 over IPv4 when available, but how the availability of one or the other is tested is a quirky businesses per se, so we put also those addresses which are inherently IPv4-only or IPv6-only to have a predictable consistent behaviour when all those difficult to predict and quirky situations happens.

G10h4ck (approved):

PR #982, opened 2023-03-18
Check for /etc/init.d/odhcpd existence before executing

### LINKED ISSUE #954: Check if odhcpd is present before enabling/disabling
odhcpd is suggested for deselection in the compilation guide on the website.
So, every user should see an error message like this (thanks @hiurequeiroz for reporting):
sh: /etc/init.d/odhcpd: not found

caused by these lines:
https://github.com/libremesh/lime-packages/blob/2a702d3aeeb8d7dab9e3181c2af2715a509584a3/packages/lime-system/files/usr/lib/lua/lime/network.lua#L189-L197

so the presence of odhcpd should be ensured before its configuration.

ilario:
Fix #954

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -189,11 +189,11 @@ function network.clean()
 	if config.get_bool("network", "use_odhcpd", false) then
 		utils.log("Use odhcpd as dhcp server")
 		uci:set("dhcp", "odchpd", "maindhcp", 1)
-		os.execute("/etc/init.d/odhcpd enable")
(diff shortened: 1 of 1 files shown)


G10h4ck:
did you tested it? If so I can merge it

ilario:
Yes, tested on a router with OpenWrt 22.03 and odhcpd installed. For checking what happems when the file is not present, I renamed that file.

G10h4ck:
I am not sure the behaviour of `os.execute(...)` is the same on OpenWrt 19.07, either test it or we keep this for after the retargetting

ilario:
I connected to my router with 19.07 installed and realized that I already tested there.
Now I am not sure if I tested it also on 22.03 or only on 19.07, sorry for this.
As soon as I manage to test it again on 22.03 I will let you know, for now I can confirm this works on 19.07.

PR #983, opened 2023-03-18
shared-state check for babeld file existence before reading it

ilario:
When installing shared-state without Babeld, the following error appears every time shared-state is used:

```
# shared-state sync dnsmasq-hosts
grep: /var/etc/babeld.conf: No such file or directory
```

I tested the proposed change and executes the code only if that file is present.

Tell me if you prefer I open also an issue on this.


ilario:
Tested this on a router running OpenWrt 22.03

G10h4ck (approved):

PR #986, opened 2023-03-26
lime-system: fix 5Ghz detection for 22.03

aparcar:


aparcar:
Actually never mind, this isn't required. A lack of drivers caused lime-config to fail, not a change in software.

PR #987, opened 2023-03-26
lime-proto-babeld: enable ubus bindings

aparcar:
Makes it easier to debug and allows the usage of luci-app-babeld


ilario:
https://openwrt.org/docs/guide-user/services/babeld#ubus_bindings

Looks very useful!

ilario (approved):

G10h4ck (approved):

PR #988, opened 2023-03-26
improve unique_append and exit 0 cleanup

aparcar:


ilario:
Looks good to me!
I was going to propose `unique-append` to OpenWrt's packages repository as an executable, do you think it makes sense anyway?

I agree that the exit 0 command should not be needed.
Is this correct @spiccinini @G10h4ck ?

@aparcar maybe you missed an exit 0 also in `packages/pirania/files/etc/uci-defaults/90-captive-portal-cron`?

aparcar:
> I was going to propose unique-append to OpenWrt's packages repository as an executable, do you think it makes sense anyway?

You could add it to base-files/lib/functions.sh

ilario:
Fix #949 replacing the solution from #980

G10h4ck:
no please just make a command, one probably don't want to include a bigger library for a small guy like that

aparcar:
You mean like a file in /bin/ or something?

G10h4ck:
> You mean like a file in /bin/ or something?

yes

G10h4ck:
> Isn't it easier to make that part of the lime-system package instead of adding yet another custom package? The metadata alone for OPKG is literally 10 times more lines than the script itself.

Bffff that is a problem per se... not related to make it a command itself (which I still insist)
What do you suggests to keep things as modular as possible without bloating metadata?

Just for this two things we could fit them into `lime-system`, but definitely doesn't seems the way to go for more stuff...

ilario:
Will fix #949

G10h4ck:
Apparently this stagnated in favor of other discussed solutions

Thanks anyway ;)

ilario:
> Apparently this stagnated in favor of other discussed solutions
> 
> Thanks anyway ;)

Wait... I don't get this... Which other discussed solutions?

Issue #909, opened 2022-01-11
safe-reboot can be triggered by unauthenticated user 


germanferrero:
See usr/libexec/rpcd/lime-utils, there is a method for calling safe-reboot. This allow anyone reaching the router ip to reboot it.

ilario:
I had no idea, verrrry interesting!!!!
I don't know how to use this, so I managed to get only to this point:
```
$ curl --location --request POST 'http://thisnode.info/ubus' \
--header 'Content-Type: application/json' \
--data-raw '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "call",
    "params": [
        "00000000000000000000000000000000",
        "lime-utils",
        "safe_reboot",
        {
            "action": "status"
        }
    ]
}'
```

Resulting in

```
{"jsonrpc":"2.0","id":1,"error":{"code":-32002,"message":"Access denied"}}
```

And [reading here](https://openwrt.org/docs/techref/ubus#actually_making_calls) seems that I should get an ID from a previous login step, so it does not look like completely open to non-users to me, but I really never touched this stuff before.
Thanks for finding this interesting issue!

ilario:
@selankon @spiccinini @germanferrero can anyone give instructions on how to trigger this? Or can you tell if this is already safe?

ilario:
Asked to Ger via email and got details, still to be tested:

> Fíjate de entrar a limeapp con el navegador y la pestaña de network abierta.
> Verás un "ajax" request de login inicial que esto crea un session id para el usuario lime-app.
> 
> Incluyendo ese session id en el request de safe-reboot ya podrás reproducirlo!

selankon:
Is there any reason to unprotect this method? I mean, a use case where an admin that can't reach the ip tells to a random person without knowledge to access and reboot? Or is a bug at all?

Either, I cannot find how `lime-utils-admin` methods are protected... I suspect that could be enough to move the `safe_reboot` functions to lime-utils-admin files.

I will try ASAP

ilario:
The issue has been opened by @germanferrero, whho implemented a lot of this stuff (I believe), so I would bet it is a bug, not a feature.

selankon:
FTR, to get session just do:

```bash
curl -X POST http://thisnode.info/ubus/ -d '{"id":1,"jsonrpc":"2.0","method":"call","params":["00000000000000000000000000000000","session","login",{"username":"lime-app","password":"generic","timeout":5000}]}'
```

[...]

PR #989, opened 2023-03-27
Move safe reboot to admin protected function

selankon:
It fixes https://github.com/libremesh/lime-packages/issues/909


ilario:
Heloo, from the conversation on #909 I get that this is ready to merge, right? Can you change it from draft to ready for review? Thanks for your work!!!!!!

ilario (approved):

PR #990, opened 2023-03-27
Port libremesh to fw4 and nftables

G10h4ck:


G10h4ck:
Merging this PR LiMe should work reasonably against OpenWrt main branch

nicoechaniz (commented):
ok

nicoechaniz (approved):

ilario:
Wait, maybe it was some problem with the bat0 interface, I have to double check that.

pony1k:
I think, the dependency on `kmod-ebtables-ipv6` should stay. It was removed in this commit.

ilario:
@pony1k can you share us why you think so? I am completely ignorant on this, so maybe it is obvious, but it would be useful if you could tell more. Do you refer to the error message I reported above?

@G10h4ck for reproducing the error message I reported above, you can simply run
```
/etc/init.d/lime-anygw-ebtables restart
```

pony1k:
> @pony1k can you share us why you think so? I am completely ignorant on this, so maybe it is obvious, but it would be useful if you could tell more. Do you refer to the error message I reported above?

Sure! The package description of `ebtables-kmod-ipv6` reads
```This option adds the IPv6 support to ebtables, which allows basic IPv6 header field filtering and target support.```
and the lines
https://github.com/libremesh/lime-packages/blob/8aa007e456d3cc98fd45997689624dbbeade3c7b/packages/lime-proto-anygw/files/etc/init.d/lime-anygw-ebtables#L11-L12
are defining ebtables-rules that use the ip6 match extention of ebtables. I haven't reproduced the error, but I think this is very likely the cause.

**Comment by pony1k** [comment]
@G10h4ck Are you sure?
First of all, `ebtables-nft` selects the package `kmod-ebtables`. Thats makes me believe that it must be somehow using that. I did a lot of reading. [Here](https://www.redhat.com/en/blog/using-iptables-nft-hybrid-linux-firewall) they say about `iptables-nft`:

> Extensions (called in iptables via -m or, with a few exceptions, -j parameter) are still parsed by the extension modules themselves, so iptables-nft reuses that code. This is necessary anyway since compat expressions expect a payload exactly as created by those extension parsers.

Maybe this is also true for `ebtables-nft`? The manual page `ebtables-nft.8` states, regarding matching extensions:

> These  extensions deal with functionality supported by kernel modules supplemental to the core ebtables code.

I believe they mean kernel modules like `ebtables-ipv6`. Also interesting I find [this](https://developers.redhat.com/blog/2020/08/18/iptables-the-two-variants-and-their-relationship-with-nftables) article from redhat blog, about the relationship between `ip-tables-legacy`, `ip-tables-nft` and `nftables`.

Because this is all still a bit confusing for me, I did some testing. This is on a router without `kmod-ebtables-ipv6` with OpenWrt 22 and libremesh master:
```
root@wtw-teilchen-mast-mitte:~# ebtables -t nat -L POSTROUTING
Bridge table: nat

Bridge chain: POSTROUTING, entries: 1, policy: ACCEPT
-s aa:aa:aa:00:00:00/ff:ff:ff:00:00:00 -o bat0 -j DROP
```
You see, the rules with ipv6 are missing. 
```
root@wtw-teilchen-mast-mitte:~# nft list chain bridge nat POSTROUTING
table bridge nat {
        chain POSTROUTING {
                type filter hook postrouting priority srcnat; policy accept;
                oifname "bat0" ether saddr aa:aa:aa:00:00:00/24 counter packets 126 bytes 11304 drop
        }
}
```
Also when you list the rules with nft.
```
root@wtw-teilchen-mast-mitte:~# /etc/init.d/lime-anygw-ebtables start
ebtables v1.8.7 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain POSTROUTING
ebtables v1.8.7 (nf_tables):  RULE_APPEND failed (No such file or directory): rule in chain POSTROUTING
```
This is the error message @ilario saw.
Then, after I installed `kmod-ebtables-ipv6` on the device:
```
root@wtw-teilchen-mast-mitte:~# /etc/init.d/lime-anygw-ebtables start
```
No error, cool.
```
root@wtw-teilchen-mast-mitte:~# ebtables -t nat -L POSTROUTING
Bridge table: nat

Bridge chain: POSTROUTING, entries: 5, policy: ACCEPT
-s aa:aa:aa:00:00:00/ff:ff:ff:00:00:00 -o bat0 -j DROP
-s aa:aa:aa:00:00:00/ff:ff:ff:00:00:00 -o bat0 -j DROP
-s aa:aa:aa:00:00:00/ff:ff:ff:00:00:00 -o bat0 -j DROP
-p 0x86dd -o bat0 --ip6-proto ipv6-icmp --ip6-icmp-type router-solicitation -j DROP
-p 0x86dd -o bat0 --ip6-proto ipv6-icmp --ip6-icmp-type router-advertisement -j DROP
```
```
root@wtw-teilchen-mast-mitte:~# nft list chain bridge nat POSTROUTING
table bridge nat {
        chain POSTROUTING {
                type filter hook postrouting priority srcnat; policy accept;
                oifname "bat0" ether saddr aa:aa:aa:00:00:00/24 counter packets 160 bytes 14256 drop
                oifname "bat0" ether saddr aa:aa:aa:00:00:00/24 counter packets 0 bytes 0 drop
                oifname "bat0" ether saddr aa:aa:aa:00:00:00/24 counter packets 0 bytes 0 drop
                oifname "bat0" ether type ip6 # xt_ip6 counter packets 0 bytes 0 drop
                oifname "bat0" ether type ip6 # xt_ip6 counter packets 0 bytes 0 drop
        }
}
```
Noteworthy is the comment `# xt_ip6 counter packets 0 bytes 0 drop` in the end. This is nft telling us that it uses xtables matching, as opposed to nftables.

PR #991, opened 2023-03-27
Adding the random-numgen command and use it for removing usage of $RANDOM

### LINKED ISSUE #800: No value in $RANDOM

ilario:
Fix #800 
Replaces the (other half of) #980 

When https://github.com/openwrt/packages/pull/20728 will be included in openwrt-22.03, the package will need to be removed from lime-packages repository.

### DIFF
--- a/packages/check-date-http/Makefile
+++ b/packages/check-date-http/Makefile
@@ -19,7 +19,7 @@ define Package/$(PKG_NAME)
 	CATEGORY:=LibreMesh
 	MAINTAINER:=Gioacchino Mazzurco <gio@altermundi.net>
 	URL:=http://libremesh.org
-	DEPENDS:=+libuci-lua +lua \
+	DEPENDS:=+libuci-lua +lua +random-numgen \
 		+luci-lib-httpclient
 	PKGARCH:=all
 endef
--- a/packages/check-date-http/Makefile
+++ b/packages/check-date-http/Makefile
@@ -19,7 +19,7 @@ define Package/$(PKG_NAME)
 	CATEGORY:=LibreMesh
 	MAINTAINER:=Gioacchino Mazzurco <gio@altermundi.net>
 	URL:=http://libremesh.org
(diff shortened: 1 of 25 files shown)


G10h4ck:
You forgot to add the dependency on `random-numgen` for packages that used `$RANDOM` before

G10h4ck (changes requested):
see discussion

ilario:
Right! Thanks!
Fixing ASAP.

ilario:
Now it should be good.

pony1k:
Why do we not want to use the busybox-`$RANDOM`? We would only need to add a dependency on `BUSYBOX_CUSTOM`.

ilario:
> Why do we not want to use the busybox-`$RANDOM`? We would only need to add a dependency on `BUSYBOX_CUSTOM`.

For going toward the solution of #1000

G10h4ck (approved):

G10h4ck:
Bufff, so we should just revert this PR -_-

G10h4ck:
@ilario can you revert it cleanly ?

ilario:
I would wait for reverting it.
This can coexist with $RANDOM.
I would revert it once we abandon the support for OpenWrt 22.03 (that will happen in a long time, as we just dropped the support for OpenWrt 19.07...).

G10h4ck:
My modifications are targetting openwrt main branch not 2020 release

ilario:
2020.3 release does not include this.
2023.1 will be based on OpenWrt 22.03, for which the $RANDOM-inclusion commit will, very likely, not be backported. So we need random-numgen.

Anyway, there is very little harm (maybe the image size increases by 1 kB? I did not measure) in keeping random-numgen until when we are sure that everybody is compiling on an OpenWrt version that includes $RANDOM.

PR #992, opened 2023-03-28
lime-system: enable dhcp for wan by default

aparcar:
Instead of setting it to none.


ilario:
In the previous releases this was set here:

https://github.com/libremesh/lime-packages/blob/8aa007e456d3cc98fd45997689624dbbeade3c7b/packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua#L22

But for me it stopped working at some point. I suppose that #990 fixed this...?
Did you try if #990 (installing both `lime-proto-wan` and `lime-hwd-openwrt-wan`) works?

If that works, do you think that we should set it to `dhcp` by default anyway?

@G10h4ck @spiccinini is there any reason for not having that on `dhcp` by default?

aparcar:
I have both installed and currently it does not work, the `wan` interface is skipped

ilario:
Ok, so it is the same that happened also to me on a YouHua WR1200JS.
The lime-hwd-openwrt-wan was detecting the WAN interface, but then this was not processed in lime.network's `owrt_ifname_parser` nor in `owrt_device_parser`.
Maybe is because they still look for the pre-openwrt21.02 `ifname` strings?

ilario:
> Ok, so it is the same that happened also to me on a YouHua WR1200JS. The lime-hwd-openwrt-wan was detecting the WAN interface, but then this was not processed in lime.network's `owrt_ifname_parser` nor in `owrt_device_parser`. Maybe is because they still look for the pre-openwrt21.02 `ifname` strings?

Confirming:
It misses the wan interface because owrt_ifname_parser does not look for the "device" parameter content. But I am not sure at all that replacing "ifname" with "device" in owrt_ifname_parser is a good idea... Seems that there is useful stuff also in ifname. To be confirmed.

G10h4ck:
Should be fixed by https://github.com/libremesh/lime-packages/pull/998

Issue #997, opened 2023-03-30
Wireless band hwmode changed from 11g and 11a to 2g and 5g 


ilario:
Comparing the `/etc/config/wireless` file from OpenWrt 19.07 and 22.03 there is a clear difference:

`hwmode` for the 2.4 GHz radio changed from `11g` to `2g`
`hwmode` for the 5 GHz radio changed from `11a` to `5g`

this affects some tests and the code of only one abandoned package:

https://github.com/libremesh/lime-packages/blob/06399e5de7fb7b9d3cf13c19f7bf70accb11299a/packages/lime-hwd-usbradio/src/usbradio.lua#L69

ilario:
> Hello maintainers, I'm new to freifunk.net open source contributions & excited to get started. I would like to work on this issue, & I'll follow the contribution guidelines carefully. Could you please assign it to me if it's available? Thank you!

Please go ahead, it is available and nobody is working on this. Thanks for your help!

PR #998, opened 2023-03-30
Split network.lua's owrt_ifname_parser

ilario:
From OpenWrt 19.07 to 21.02, the content of `/etc/config/wireless` did not change (except #997) but the content of interfaces and devices in `/etc/config/network` changed significantly.

Inside lime-system's network.lua, the owrt_ifname_parser was used both for parsing `wifi-iface` from  `/etc/config/wireless` and for `interface` from `/etc/config/network`.

In this yet-to-be-tested code, I added a new section for probing only the interfaces from `/etc/config/networks`.

Also, I removed the "compound interface" code from owrt_ifname_parser as I suppose this was needed for br-lan, that now passed from being an `interface` to a `device` and so is taken care by owrt_device_parser.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -272,12 +272,14 @@ function network.scandevices()
 			           ifn )
 			dev_parser(ifn)
 		end
-		if ( type(ifn) == "table" ) then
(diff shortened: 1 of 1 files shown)


ilario:
Tested, works for me for getting `dhcp` activated on `wan` interface.

G10h4ck (approved):

PR #999, opened 2023-04-03
Remove iputils-ping retrocompatibility with OpenWrt 19.07

### LINKED ISSUE #794: Package lime-debug dependency to iputils-ping6
when i tried to compile libremesh i failed with openwrt-trunk.
the package "iputils-ping6" seems to be missing, so i got this error:
`u@PC:~/git2/openwrt$ time make V=99`
`WARNING: Makefile 'package/feeds/libremesh/lime-debug/Makefile' has a dependency on 'iputils-ping6', which does not exist`
`WARNING: Makefile 'package/feeds/libremesh/lime-proto-bgp/Makefile' has a dependency on 'bird4', which does not exist`
`WARNING: Makefile 'package/feeds/libremesh/lime-proto-bgp/Makefile' has a dependency on 'bird6', which does not exist `
`...`
`Collected errors:`
` * satisfy_dependencies_for: Cannot satisfy the following dependencies for lime-debug:`
` *     iputils-ping6 `

when trying it with openwrt brunch 19.07 i got only these bird-errors. also on openwrt packages-download dirs there is no iputils-ping6 any more. (f.i. https://downloads.openwrt.org/snapshots/packages/i386_pentium/base/)

so, do we have to change https://github.com/libremesh/lime-packages/blob/master/packages/lime-debug/Makefile ?

ilario:
Fix #794 

The `ping6retrocompat.mk` file was used for compiling with OpenWrt releases having split `iputils-ping` and `iputils-ping6` or single package `iputils-ping`.

For some reason, it did not work for me, and it is not needed anymore as we are not supporting OpenWrt 19.07 on the master branch (the OpenWrt_19.07_compatible branch has been created for that).


G10h4ck:
Removing `$(PING6_SYMLINK)` dependency will cause `ping4` and `ping6` not being provided by `iputils-ping` anymore so `shared-state` will be broken because it uses `ping6` you should also change the occurrences of `ping6` into `ping -6` at least into `shared-state-get_candidates_neigh`

ilario:
> Removing `$(PING6_SYMLINK)` dependency will cause `ping4` and `ping6` not being provided by `iputils-ping` anymore so `shared-state` will be broken because it uses `ping6` you should also change the occurrences of `ping6` into `ping -6` at least into `shared-state-get_candidates_neigh`

Right, will do. Thanks!

ilario:
> you should also change the occurrences of `ping6` into `ping -6`

Isn't it enough to use `ping` without the `ping -6`?
I did not test yet if they are equivalent, do you have insight on this?

ilario:
@selankon please check if the modification of `get_loss` can have any impact on lime-app

I added some commits where I replaced all the occurrences of `ping6` with `ping`.
I did not include the `-6` option to `ping` for decreasing the complexity of the code (you will see that I got rid of some code that was checking if an IP is v4 or v6).
But it would be easy to just revert the commits and replace `ping6` with `ping -6`, if needed.

G10h4ck:
should be fine, please test with a few of routers if shared-state keeps discovering neighbors and then we can merge this

G10h4ck (approved):

Issue #647, opened 2019-11-24
HT40 not accepted as channel width


ilario:
Checking `logread` I noticed an error related to the 5 GHz channel width used: HT40 is not an accepted value (already reported in #630).

```
daemon.notice netifd: radio1 (2614): Usage:	iw [options] dev <devname> set channel <channel> [NOHT|HT20|HT40+|HT40-|5MHz|10MHz|80MHz]
daemon.notice netifd: radio1 (2614): Options:
daemon.notice netifd: radio1 (2614): 	--debug		enable netlink debugging
user.notice mac80211: Failed command: iw dev wlan1-mesh set channel 48 HT40
daemon.err hostapd: Configuration file: /var/run/hostapd-phy1.conf
```

This is in contrast with the [documentation on OpenWrt website here](https://openwrt.org/docs/guide-user/network/wifi/basic#common_options).

I thought that replacing `HT40` with `HT40-` was a good idea (see [the attempt here](https://github.com/libremesh/lime-packages/compare/330ae4ca17d5f0b490f12dd280990d94a3ecca2c...ilario:ht40)) but actually this would work for channel 48 but not for channel 157.
Check out the allowed HT40+ or HT40- [on this website](https://www.digi.com/resources/documentation/digidocs/90001394/references/r_5ghz_h20_and_ht40_chanl.htm).

One possible solution is to copy the system employed [in lime-smart-wifi here](https://github.com/libremesh/lime-packages/blob/ede4b03a4df3af5c9b3be1545e98099c728e6ba9/packages/lime-smart-wifi/src/lime-smart-wifi.lua#L61-L85) and use it in lime-system's wireless.lua?
Otherwise we could accept a list of HT* related to the list of 5 GHz channels???

I tried replacing the default iw package with the iw-full package but HT40 (without +/-) is still not accepted.

dangowrt:
I'm trying to understand how this call to `iw` is happening. In OpenWrt, we are automatically replacing `HT40` with `HT40-` or `HT40+` depending on band and channel, see
https://github.com/openwrt/openwrt/blob/master/package/kernel/mac80211/files/lib/netifd/wireless/mac80211.sh#L726
and
https://github.com/openwrt/openwrt/blob/master/package/kernel/mac80211/files/lib/netifd/wireless/mac80211.sh#L631

However, this is apparently not the case here:
[...]

PR #1001, opened 2023-04-04
Updated lime-example to follow lime-defaults

ilario:
Modified lime-example.txt for being closer to https://github.com/libremesh/lime-packages/blob/master/packages/lime-system/files/etc/config/lime-defaults


ilario:
Added also some info about the `htmode` parameter, inspired by this comment: https://github.com/libremesh/lime-packages/issues/647#issuecomment-1503968192

G10h4ck (approved):

PR #1004, opened 2023-04-06
shared-state-dnsmasq_servers correct serversfile option setting

ilario:
Fix one of the two still existing errors from this issue: https://github.com/libremesh/lime-packages/issues/970#issuecomment-1490091329

If the option gets set out of UCI, then ujail forbids the access to the `/var/shared-state/dnsmasq_servers` file, resulting in a `No such file or directory` error.


ilario:
I had to advance in init order the shared-state service, for having it create the dnsmasq_servers file before that dnsmasq tries to read it.
I think that it would work even if the file is created later, but this way we also avoid some "No such file" errors when dnsmasq starts.

ilario:
So now it is tested on a YouHua WR1200JS with OpenWrt 22.03 and works.

G10h4ck (approved):

PR #1012, opened 2023-04-13
lime-proto-batadv remove retrocompatibility

ilario:
The support for the new configuration style for Batman-adv has been introduced in https://github.com/libremesh/lime-packages/commit/fed6a468257ac461a4c81b16ac3361a36e7d3bde

Here I remove the new configuration style.
I did not test yet, but the edit is simple enough that I hope I didn't break anything.

### DIFF
--- a/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
+++ b/packages/lime-proto-batadv/files/usr/lib/lua/lime/proto/batadv.lua
@@ -9,49 +9,28 @@ local config = require("lime.config")
 batadv = {}
 
 batadv.configured = false
-batadv.old_cfg_api = false
-batadv.type_option = 'master'
-batadv.ifc_proto = 'batadv_hardif'
-
-function batadv.detect_old_cfg_api()
-    return not fs.lstat("/lib/netifd/proto/batadv_hardif.sh")
-end
 
 function batadv.configure(args)
 	if batadv.configured then return end
 	batadv.configured = true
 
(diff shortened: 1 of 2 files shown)


spiccinini (approved):

spiccinini:
The code looks good but it still does not pass the tests

ilario:
> The code looks good but it still does not pass the tests

Yesss, someone should prepare a Docker image for testing on top of OpenWrt 22.03.

spiccinini:
It is not clear to me why tests stop working do you know which merge broke them,  and I dont see why it is related to 22.03 as the tests are unit tests of lua files mostly, some libs like uci are from openwrt but if we are not using new features then it should continue working. So it seems that some merged code was merged and broke the tests. Finding which one will help fix it

ilario:
It was expected to happen with #959 which is when we abandoned the support on OpenWrt 19.07 in the master branch.

ilario:
Indeed, the fails in these tests refer to the names of the ethernet interfaces https://github.com/libremesh/lime-packages/pull/959/checks

ilario:
For example, this commit stops looking for interfaces using the pre-OpenWrt 21.02 naming.

spiccinini:
Indeed but the issue is that the tests need to be updated, not the test image

ilario:
> Indeed but the issue is that the tests need to be updated, not the test image

Ooooohhh, ok, I did not expect that.
I am slowly fixing the failing tests. But I can already tell you that I have no LibreRouter with me for updating these files: 
https://github.com/libremesh/lime-packages/tree/master/tests/devices/librerouter-v1

PR #1013, opened 2023-04-13
Batman-adv add the orig_interval to the lime-* config files and set a larger default value

### LINKED ISSUE #1010: Increase Batman-adv's orig_interval for decreasing overhead traffic
We could increase the time between each packet sent by the Batman-adv routing protocol for probing the packet loss.

Using tcpdump you can see the amount of packets set from Batman-adv, and it is a quite large and unneeded flow of packets.
Up to now it is set to 1000 = 1 per second, but the actual packets flowing are many more:  in LibreMesh we have a huge number of interfaces (since OpenWrt 21.02 we have even more: one for each LAN port!) and Batman-adv sends packets from each router to each interface. The result is that we have a quite large flow of OGM, much more than needed.

Also, in LibreMesh networks, usually nodes do not move much (clients move, but nodes usually don't), so there is no need for a such-frequent probing of the link quality.

Documented here: https://openwrt.org/docs/guide-user/network/wifi/mesh/batman#batman-adv_options_for_bat0_the_main_mesh_interface

ilario:
Fix #1010


ilario:
Tested on a YouHua WR1200JS with OpenWrt 22.03.
Works, also confirmed with tcpdump.

G10h4ck:
In my experience even with the default value of 1000 batman-adv is quite slow to detect when a node goes down, sometimes even if other paths are available, the connection is broken until the fallen node comes back, I wonder how changing this setting will impact on that situation, hopefully it doesn't make things too much worse, let's merge this for now and keep an eye on how this setting alter the behaviour in larger networks

G10h4ck (approved):

Issue #1009, opened 2023-04-12
Move from BATMAN_IV to BATMAN_V


ilario:
For evaluating the links, the version 5 of Batman-adv uses estimated bandwidth (asking to the wireless drivers) rather than packet loss.
It should be just better and stable enough to use (from what I understood from @lindnermarek).

We can try what happens when setting this in the `routing_algo` https://openwrt.org/docs/guide-user/network/wifi/mesh/batman#batman-adv_options_for_bat0_the_main_mesh_interface

ilario:
During Wireless BattleMesh we realized that BATMAN_V is not correctly detecting the throughput of the wifi interfaces, indicating all of them as zero throughput.

PR #1014, opened 2023-04-13
Batman-adv allow the user to set the routing_algo

ilario:
This is a first step for #1009 
By default, here BATMAN_IV is still selected.


ilario:
Tested on a YouHua WR1200JS with OpenWrt 22.03.
Works, also confirmed with the first line of the output of `batctl o`.

G10h4ck (approved):

Issue #1000, opened 2023-04-03
Get compatible with OpenWrt ImageBuilder


ilario:
Following the discussion from https://github.com/libremesh/lime-packages/issues/794

Advantages would be many.
For example, being able to compile LibreMesh images through https://firmware-selector.openwrt.org/ or allowing the community to compile images without needing the full BuildRoot and without the need to create and host our own ImageBuilder.

Related issues:
#794 (and #999 would help also for this issue)
#800 (and #991 is specifically written for this issue)

Some problematic code:

https://github.com/libremesh/lime-packages/blob/39a57a75622aeb51672a00ab3c1987238395248a/packages/lime-debug/Makefile#L38-L42

ilario:
Likely, also the dependency from `dnsmasq-dhcpv6` is going to be problematic, as it conflicts with the `dnsmasq` package which is included by default.
https://github.com/libremesh/lime-packages/blob/8aa007e456d3cc98fd45997689624dbbeade3c7b/packages/lime-proto-anygw/Makefile#L24

For example, when compiling with both packages selected, errors such this appear:

```
 * check_data_file_clashes: Package dnsmasq-dhcpv6 wants to install file openwrt/build_dir/target-arm_cortex-a7+neon-vfpv4_musl_eabi/root-ipq40xx/etc/hotplug.d/ntp/25-dnsmasqsec
	But that file is already provided by package  * dnsmasq
```

ilario:
> Likely, also the dependency from `dnsmasq-dhcpv6` is going to be problematic, as it conflicts with the `dnsmasq` package which is included by default.

The solution has to be: using dnsmasq for IPv4 and odhcpd-ipv6only for IPv6. See #294

ilario:
We should also get rid of the `@(PACKAGE_iw||PACKAGE_iw-full)` in the Makefiles of prometheus-node-exporter-lua-location-latlon, prometheus-node-exporter-lua-wifi-params, prometheus-node-exporter-lua-wifi-stations-extra and prometheus-node-exporter-lua-wifi-survey. But these packages are not suggested on the website, so it is not important.

ilario:
An important step forward has been made in #1015.

The created x86 images does not yet boot on qemu (reported by @a-gave ) and the other images are yet to be tested.

PR #1015, opened 2023-04-14
various readme improvements

aparcar:
Let's clean things up a little bit and let people use use the ImageBuilder.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,8 @@
 
 [![LibreMesh logo](https://raw.githubusercontent.com/libremesh/lime-web/master/logo/logo.png)](https://libremesh.org)
 
-[LibreMesh project][5] includes the development of several tools used for deploying libre/free mesh networks.
+[LibreMesh project][5] includes the development of several tools used for
+deploying libre/free mesh networks.
 
 The firmware (the main piece) will allow simple deployment of auto-configurable,
 yet versatile, multi-radio mesh networks. Check the [Network Architecture][4] to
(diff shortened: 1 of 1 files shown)


G10h4ck:
For new WiFi 6 radios, we might end up needing a custom build for hostapd, is this possible with imagebuilder?

aparcar:
uhm... why do we need custom builds with the imagebuiler? I'm running LibreMesh successfully on WiFi 6 (and 7) devices with images based on the ImageBuilder.

ilario:
Don't you have any problem with dnsmasq-ipv6? https://github.com/libremesh/lime-packages/issues/1000#issuecomment-1497391930

I think that apart from that, we are already compatible with the image builder for the core packages (see #1000 for problems with non-core packages).

pony1k:
@ilario Why should there be a problem with `dnsmasq-ipv6`? When I specify `PACKAGES=-dnsmasq [...]`, imagebuilder builds without dnsmasq.

ilario:
> When I specify `PACKAGES=-dnsmasq [...]`, imagebuilder builds without dnsmasq.

Ooh, I didn't know that was possible, nice!

ilario:
Confirmed, you need to add "-dnsmasq", otherwise you have file conflicts with "dnsmasq-dhcpv6".
So I would merge this and then:
1) add the "-dnsmasq"
2) add a minimum amount of recommended packages (maybe babeld alone works, but I feel is not tested enough for suggesting that in the readme). So adding "lime-proto-batadv lime-proto-anygw lime-hwd-openwrt-wan" (last one is for having the WAN port configured as WAN)

@aparcar should I merge and do that edits or can you add this to the pull request? Thanks for this amazing work!

aparcar:
Kindly proceed.

ilario (approved):

Issue #1008, opened 2023-04-12
All routers answering to anygw IP when connected via LAN-LAN


ilario:
I first reported this issue here (first spotted by @pedro-nonfree) https://github.com/libremesh/lime-packages/issues/1007#issuecomment-1503684283 but @pony1k correctly identified that the issues are idependent.

What happens is that when one use the anygw IP (e.g. connecting to thisnode.info) gets answers from both routers, if they are connected in this way:

Laptop-ethernet port____lan1-router1-lan2____lan1-router2

This was observed using DSA-supported routers (PlasmaCloud PA1200 with both ports configured as LAN; Xiaomi MiRouter 4A gigabit edition) using OpenWrt 22.03 branch.

Seems that this should be avoided by these rules: 
https://github.com/libremesh/lime-packages/blob/8aa007e456d3cc98fd45997689624dbbeade3c7b/packages/lime-proto-anygw/files/etc/init.d/lime-anygw-ebtables#L9-L10

G10h4ck:
In case of non-DSA hardware switches, those rules does not even see the packets because they are switched directly by the switching chip without pushing them to the CPU and then to the kernel.

In case of DSA hardware switch it probably depends on the implementation, reading a bunch of kernel and driver code should help deciphering this.

G10h4ck:
anyway one should not use AnyGW IP for SSH or similar things, having the lime-app responding on that IP is already an usability compromise

LaneaLucy:
Shouldn't there be ground routing set, if you connect 2 lime systems over cable, which by default uses different vlans, which again solves the problem, that the CPU don't see the packets?

G10h4ck:
> Shouldn't there be ground routing set, if you connect 2 lime systems over cable, which by default uses different vlans, which again solves the problem, that the CPU don't see the packets?

That is a possibility but it doesn't happen automatically

ilario:
> anyway one should not use AnyGW IP for SSH or similar things, having the lime-app responding on that IP is already an usability compromise

@G10h4ck Also lime-app will have a very erratic behavior in this setup, no? Users are encouraged to access lime-app via names like thisnode.info, that gets resolved with the AnyGW IP. Then the browser will randomly receive answers from both router1 and router2, no?

[...]

PR #1016, opened 2023-04-15
lime-proto-anygw: set thisnode.info to local node IP

ilario:
Seems that the possibility to use the Anygw IP for connecting to the web or the SSH interface is not going to be restored, see https://github.com/libremesh/lime-packages/issues/1008#issuecomment-1508445716

So, in order to have http://thisnode.info working, we can have it pointing to the local node IP, rather to the Anygw IP.

Up to now I checked that the gateway IP used by the clients is still being the same, and that thisnode.info is resolved as the directly-connected node IP.
What I still have to check is, from the point of view of the clients, whether the DNS resolver is the anygw IP or the local node IP.

### DIFF
--- a/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
+++ b/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua
@@ -102,7 +102,7 @@ function anygw.configure(args)
 	--! it wildcards subdomains)
 	uci:set("dhcp", "anygw_dns", "hostrecord")
 	uci:set("dhcp", "anygw_dns", "name", {"anygw", unpack(anygw.FQDN)})
-	uci:set("dhcp", "anygw_dns", "ip", anygw_ipv4:host():string() .. "," .. anygw_ipv6:host():string())
+	uci:set("dhcp", "anygw_dns", "ip", ipv4:host():string() .. "," .. ipv6:host():string())
 
 	uci:save("dhcp")
 


ilario:
> > web or the SSH interface is not going to be restored
> 
> What do you mean?

I mean that if #1008 is left as it is now (and this comment seems to indicate that it is not going to be fixed https://github.com/libremesh/lime-packages/issues/1008#issuecomment-1508445716 but maybe I misunderstood) the Anygw IP (e.g. with the default configuration 10.13.0.1) will not work for connecting via SSH nor via HTTPS. I would suppose it will not work with HTTP neither, but I did not check.

Obviously, this will happen in the circumstances described in #1008, but they are not unusual.

> Lime-app seems working fine,

This deserves another issue, for me it usually does not work, I usually see only a white page with the upper menu (like this: https://github.com/libremesh/lime-app/issues/378#issuecomment-1511266962), but I did not check into that for making sure that it happens 100% of the times.

> while anygw IP was never meant to be used for SSH connections, one can use the we interface to determine the node specific IP, or `ping ff02::1%youriface` I won't fiddle with this kind of things that may break other stuff to avoid a minor nuisance in a corner case

Please notice that if the Anygw IP does not work, neither the "thisnode.info" will suffer the same issues.
So are you suggesting that users should run the broadcast ping and use the link-local IPv6 for connecting to the lime-app on their smartphones?

G10h4ck:
> Please notice that if the Anygw IP does not work, neither the "thisnode.info" will suffer the same issues. So are you suggesting that users should run the broadcast ping and use the link-local IPv6 for connecting to the lime-app on their ethernet connected laptops?

The most common use-case of `thisnode.info` usage is connecting to the node from a WiFi client, usually a not so smart phone, changing it to the node specific IP will break this use case in unpredictable ways, one of them is because the client will cache the DNS response and attempt to use that IP even when roaming to another node, so basically the user will see the old node webui and not the current one, which is what happening now.

Connecting more routers via cable (also this happens only when LAN-LAN cable connection happens, that we are attempting to deprecate in favor of ethernet MESH-MESH without lan) is something more common for hackers but not "usual" users, hackers can easily use other means, for example using directly the node name, or discover the IP with ping6 as suggested before.

I would not break "usual" user experience, to make life too easier in a more cornerish case which we are attempting to deprecate.

PR #1028, opened 2023-05-28
readme: expanded instructions on ImageBuilder

ilario:
In #1015, @aparcar created an excellent documentation about using the ImageBuilder for building LibreMesh on OpenWrt 22.03 (not yet supported!!).
In https://github.com/libremesh/lime-packages/pull/1015#issuecomment-1540906636 I suggested a few edits, here I implemented those and many others.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,21 +21,43 @@ We encourage each network community to create its firmware profile on
 
 [In this page][1] we provide a list of requirements that ensure you to have a
 working LibreMesh node on your router. This list comes with no warranties: read
-carefully the [model-specific instructions on OpenWrt wiki][2] and be
+carefully the [model-specific instructions on OpenWrt wiki][OpenWrt-ToH] and be
 extra-careful when flashing your routers!
 
 ## Building a Firmware Image on Your PC
 
-The LibreMesh firmware can be compiled by following [these instructions][13].
+### Building the stable release LibreMesh 2020.1
 
-### Using the ImageBuilder
+#### Using the BuildRoot
 
-Start an ImageBuilder of your choice, use containers for an easier setup:
+The BuildRoot **will cross-compile the whole OpenWrt and the LibreMesh packages** on your computer, so it will take approx 10 GB of disk space and a few hours of compilation time.
+
+For compiling LibreMesh firmware with this method, you can follow [these instructions][development_page].
+
+#### Using the ImageBuilder
+
+The ImageBuilder method is not available for the stable release.
+
+### Building the experimental firmware
+
+The experimental code still has serious issues that have to be solved, use it only for developing or debugging.
+
+#### Using the BuildRoot
+
+As explained above, in the instuctions on the website you will find where to specify the version of the code to compile.
+
+#### Using the ImageBuilder
+
+The ImageBuiler **will download pre-compiled parts of the OpenWrt releases**, and add the pre-compiled LibreMesh packages, so it is **much faster** than the BuildRoot method (but less practical if you want to develop some new features modifying LibreMesh source code).
+
+Start an ImageBuilder of your choice, for example ath79-generic if your device is supported within it, use containers for an easier setup:
 
 ```shell
 mkdir ./images/
-docker run -it -v $(pwd)/images:/images/ ghcr.io/openwrt/imagebuilder:ath79-generic-v22.03.4
+docker run -it -v $(pwd)/images:/images/ ghcr.io/openwrt/imagebuilder:ath79-generic-v22.03.5
 ```
+
+If your device is not part of ath79-generic profiles, you can replace it with another <target>-<subtarget> combination. For knowing which target and subtarget is best suited for your router, check out the page about it in the [OpenWrt's Table of Hardware][OpenWrt-ToH].
 
 Within the container, add the `lime-packages` feed:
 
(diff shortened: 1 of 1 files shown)


ilario:
Pls, this is just an expansion of @aparcar's #1015... Can anyone review?
@spiccinini @G10h4ck @selankon

a-gave:
I encountered a small issue with imagebuilder about batman-adv. I report it here in the meantime:
If i build libremesh selecting `lime-proto-batadv`, `batctl-tiny` is selected instead of `batctl-default`.

In this situation, the batctl commands listed under 'debug tables' are not available [0], and instead return the help message.
I noticed it by trying to run `batctl n` and `batctl o`.

<details>
<summary>Output of `batctl n`</summary>
```
root@falena:~# batctl n
Error - no valid command or debug table/JSON specified: n
Usage: batctl [options] command|debug table|debug json [parameters]
options:
 	-h print this help (or 'batctl <command|debug table|debug json> -h' for the parameter help)
 	-v print version

commands:
 	meshif <netdev> aggregation|ag             [0|1]             	display or modify aggregation setting
 	meshif <netdev> ap_isolation|ap            [0|1]             	display or modify ap_isolation setting
 	vlan <vdev> ap_isolation|ap                [0|1]             	display or modify ap_isolation setting for vlan device or id
 	meshif <netdev> vid <vid> ap_isolation|ap  [0|1]             	display or modify ap_isolation setting for vlan device or id
 	meshif <netdev> bonding|b                  [0|1]             	display or modify bonding setting
 	meshif <netdev> bridge_loop_avoidance|bl   [0|1]             	display or modify bridge_loop_avoidance setting
 	meshif <netdev> distributed_arp_table|dat  [0|1]             	display or modify distributed_arp_table setting
 	hardif <netdev> elp_interval|et            [interval]        	display or modify elp_interval setting
 	meshif <netdev> fragmentation|f            [0|1]             	display or modify fragmentation setting
 	meshif <netdev> gw_mode|gw                 [mode]            	display or modify the gateway mode
 	meshif <netdev> hop_penalty|hp             [penalty]         	display or modify hop_penalty setting
 	hardif <netdev> hop_penalty|hp             [penalty]         	display or modify hop_penalty setting
 	meshif <netdev> interface|if               [add|del iface(s)]	display or modify the interface settings
 	meshif <netdev> isolation_mark|mark        [mark]            	display or modify isolation_mark setting
 	meshif <netdev> loglevel|ll                [level]           	display or modify the log level
 	meshif <netdev> multicast_fanout|mo        [fanout]        	display or modify multicast_fanout setting
 	meshif <netdev> multicast_forceflood|mff   [0|1]             	display or modify multicast_forceflood setting
 	meshif <netdev> network_coding|nc          [0|1]             	display or modify network_coding setting
 	meshif <netdev> orig_interval|it           [interval]        	display or modify orig_interval setting
 	routing_algo|ra                            [mode]            	display or modify the routing algorithm
 	hardif <netdev> throughput_override|to     [mbit]        	display or modify throughput_override setting

debug tables:                                   	display the corresponding debug table

JSON queries:                                   	display results of netlink query as JSON
```
</details>

ilario:
Oooh, right!
Maybe it never happened to me as I was selecting ubus-lime-batman-adv manually, but this package is not needed, so we should not assume that people select it.

I found an old comment pointing out that this problem was likely going to appear:
https://github.com/libremesh/libremesh.github.io/issues/103#issuecomment-570812066

Do you think that we should include it by default (including `+batctl` as a dependency in the Makefile of lime-proto-batadv)? I mean, it is a debug tool, but a very useful one. Otherwise we can suggest it between the optional packages in the compilation guide of the website.

a-gave:
> I found an old comment pointing out that this problem was likely going to appear:
https://github.com/libremesh/libremesh.github.io/issues/103#issuecomment-570812066

About this comment I would add that the batctl package is still splitted into `tiny` and `default` with openwrt 22.03.5, at least with ath79_generic. They are visible here for instance https://downloads.openwrt.org/releases/22.03.5/packages/mips_24kc/routing/

> Do you think that we should include it by default (including +batctl as a dependency in the Makefile of lime-proto-batadv)?

As in the previous link from openwrt the difference in size is just 31KB. 
It seems to me that the devices work fine with batctl-tiny, but I think for those installing lime for the first, not finding the debug tables can be confusing.
A solution could be to add `batctl-default` in the compilation guide of the site, and in the Makefile of suggested-packages https://github.com/libremesh/network-profiles/tree/master/libremesh/suggested-packages. Maybe creating another suggested-packages-small-flash-size-whatever without it, or clearly specifying the `tiny` version.
I would prefere this solution, but only because it is reversible ..instead of adding it in the Makefile `lime-proto-batadv`, also if the v22.03.5 and so the current libremesh master branch are intended for device with at least 8MB of flash so 31KB should could not be a problem..  

In both cases I think it should be defined as the full '+batctl-default' or '+batctl-tiny', instead of '+batctl' , but I haven't tested it!
from https://openwrt.org/docs/guide-user/network/wifi/mesh/batman#installation
> if batctl isn't available, install batctl-default.

ilario:
> > Do you think that we should include it by default (including +batctl as a dependency in the Makefile of lime-proto-batadv)?
> 
> As in the previous link from openwrt the difference in size is just 31KB. It seems to me that the devices work fine with batctl-tiny, but I think for those installing lime for the first, not finding the debug tables can be confusing. A solution could be to add `batctl-default` in the compilation guide of the site, and in the Makefile of suggested-packages https://github.com/libremesh/network-profiles/tree/master/libremesh/suggested-packages. Maybe creating another suggested-packages-small-flash-size-whatever without it, or clearly specifying the `tiny` version. I would prefere this solution, but only because it is reversible ..instead of adding it in the Makefile `lime-proto-batadv`, also if the v22.03.5 and so the current libremesh master branch are intended for device with at least 8MB of flash so 31KB should could not be a problem..

I agree, let's add `batctl-default` this to the compilation guide (between the optional packages, where also lime-debug is) and to the "suggested-packages".
Can you make a pull request on that?

For the mini version, it would be great also.
Whatever we add as a network profile, could also be used for expanding this paragraph:

> In both cases I think it should be defined as the full '+batctl-default' or '+batctl-tiny', instead of '+batctl' , but I haven't tested it! from https://openwrt.org/docs/guide-user/network/wifi/mesh/batman#installation
> 
> > if batctl isn't available, install batctl-default.

G10h4ck (approved):

PR #1030, opened 2023-07-04
Add shared state async node  

javierbrk:
This is a proposal to implement an improvement on current shared state base infrastructure.


packages/ubus-lime-location/files/usr/lib/lua/lime/location.lua
G10h4ck, line 1: I think this change would silently break other components using that data like lime app, and anyway is unrelated to the PR, let's keep it out.

packages/shared-state-async-node/Makefile
```diff
@@ -0,0 +1,81 @@
...
+PKG_SOURCE_VERSION:= 691c4412a84bb8142c73b4f5ff319f025b3973f7
+PKG_MAINTAINER:=Javier Jorge <javierbrk@gmail.com>
+PKG_LICENSE:=AGPL-3.0
+
+HOST_BUILD_PREFIX:=$(STAGING_DIR_HOST)
+
+# Source settings (i.e. where to find the source codes)
+# This is a custom variable, used below 
+# disable git package source and enable Build/Prepare line 29-30
+# SOURCE_DIR:=
```
G10h4ck, line 40: This block should be removed as source tree override should be used for that purpose as explained by Jow here https://forum.archive.openwrt.org/viewtopic.php?id=46916

G10h4ck (changes requested):

G10h4ck (approved):

PR #1033, opened 2023-07-16
lime.network.scandevices: fix finding intefaces on dsa devices

pony1k:
On devices that have only a single lan port, like `D-Link DAP-X1860 A1`, with DSA, the lan port is just called `lan`, not `lan1` as one might expect. This changes a regex `^lan%d+$` to `^lan%d*$`, so that the `lan` interface is correctly recognized and configured.
Instead of just scraping `eth0`, `eth1`, ... from `/sys/class/net/`, the second commit adds scraping of `lan`, `lan1`, `lan2`, ... aswell. Without this, `lanx`-interaces are not configured if they do not appear in the configuration.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -243,16 +243,18 @@ function network.scandevices()
 			utils.log( "network.scandevices.dev_parser found vlan device %s " ..
 			           "and marking %s as nobridge", dev, rawif )
 		end
+
 		--! With DSA, the LAN ports are not anymore eth0.1 but lan1, lan2...
-		if dev:match("^lan%d+$") then
+		--! or just lan.
+		if dev:match("^lan%d*$") then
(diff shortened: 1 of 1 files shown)


ilario:
Looks good for me, both commits!
@spiccinini @G10h4ck

ilario:
Did you check on another device, one that has more than one LAN port, if the generated /etc/config/network is different before and after these edits?

pony1k:
It is used in production on devices with multiple ports with and without dsa (`Linksys E8450 (UBI)`, `TP-LINK TD-W8970`, `AVM FRITZ!Box 4040`, `TP-Link TL-WDR3600 v1`). I did not do a direct comparison before and after, but the configuration looks good and it is working normally. I checked that there is nothing called `lan` in the devices that have `lan1`, `lan2`, ... in `/sys/class/net/` and `lan` is the only additional string that the regex matches, so it should be fine.

ilario (approved):
Looks good. I did not test but @pony1k did.

PR #1039, opened 2023-08-08
wifi-unstuck-wa: allow parametrizable values for interval and timeout

### LINKED ISSUE #1034: Allow customization of checking frequency in wifi-unstuck-wa
Currently, [wifi-unstuck-wa](https://github.com/libremesh/lime-packages/tree/ae6b33b6d874b473f8a4c6c087d88bdd4866cc7d/packages/wifi-unstuck-wa) is set to run (if installed, it is not in the recommended packages) every 10 minutes:

https://github.com/libremesh/lime-packages/blob/ae6b33b6d874b473f8a4c6c087d88bdd4866cc7d/packages/wifi-unstuck-wa/files/etc/uci-defaults/wifi-unstuck-cron#L9

And has an internal maximum limit to the execution time set to 5 minutes:

https://github.com/libremesh/lime-packages/blob/ae6b33b6d874b473f8a4c6c087d88bdd4866cc7d/packages/wifi-unstuck-wa/files/usr/lib/lua/wifi_unstuck_wa.lua#L12

It would be useful to be able to customize the 10 minutes time. For example for reducing it to 5 minutes and assuring that the link downtime will not be longer than 5 min.
Reported by @a-gave

a-gave:
This close #1034


pony1k:
Instead of modifying the source code with sed inside an uci-defaults script, maybe it would be better to put the content of the [hook_wireless.lua.txt](https://github.com/libremesh/lime-packages/pull/1039/files#diff-a6b2fdd250b5025b2318d4c34b682a770143bcd82834e31d1373f92679aa0958) file directly in `wireless.lua` and sourround it with something like
```
if utils.is_installed("wifi-unstuck-wa") then
    ...
end
```
? Though it moves something that conceptionally belongs to the package wifi-unstuck-wa into the lime-system package. Alternatively you could put just after the line https://github.com/libremesh/lime-packages/blob/861f7bd1f3c785b71faf50267374854859cc0a1c/packages/lime-system/files/usr/lib/lua/lime/config.lua#L204-L205
something like
```
if utils.isModuleAvailable("wifi_unstuck_wa") then
    table.insert(modules_name, "wifi_unstuck_wa")
end
```
and equip the [wifi_unstuck_wa.lua](https://github.com/libremesh/lime-packages/pull/1039/files#diff-bd718ffbac990b537c059b209857a72c826633583635ed7ef08480eeb9042a0e) file with a `wu.configure()` and a `wu.cleanup()` function and put the content of [hook_wireless.lua.txt](https://github.com/libremesh/lime-packages/pull/1039/files#diff-a6b2fdd250b5025b2318d4c34b682a770143bcd82834e31d1373f92679aa0958) inside the `wu.configure()` function. Then move the file into `/usr/lib/lua/lime`. That is the cleanest solution I can think of. Just my two cents. Hope this helps!

ilario (approved):
Looks good to me, but I did not find the time to test. If you tested this, please feel free to merge it :D Thanks and amazing work @a-gave and @pony1k !!!

a-gave:
Tested on openwrt-22.03.5-ath79-generic-tplink_cpe510-v3

PR #1040, opened 2023-08-10
Shrared state ubus

javierbrk:
This pull request provides access to shared-state infrastructure for the UI. 
It is intended to be used to retrieve actual state and record global network state into shared state infrastructure.

The upgrade provides three methods witch can be used to get information from existing shared state modules like "wifi_links_info". 
```
ubus -S call shared-state get_from_sharedState '{"data_type": "wifi_links_info" }'
```
It can also be used to record and retrieve information across the whole network 
```
ubus -S call shared-state insert_into_sharedState_muti_writer '{"data_type": "ref_state_wifi_links", "json": {"primero":{"bleachTTL":27}}}'
ubus -S call shared-state get_from_sharedState_muti_writer '{"data_type": "ref_state_wifi_links"}'
```
to get the info in another node the others need to sync on the same topic
```
shared-state-multiwriter sync ref_state_wifi_links
ubus -S call shared-state get_from_sharedState '{"data_type": "ref_state_wifi_links" }'
```
in conclusion the ui cant create new data types. We need to add them first as a module in every node.


javierajorge:
Output data should not include shared state info, I'll remove extra information

packages/shared-state/files/usr/libexec/rpcd/shared-state
```diff
@@ -0,0 +1,70 @@
...
+require("nixio.util")
+
+local function get_from_sharedState(msg)
+    local sharedState = shared_state.SharedState:new(msg.data_type, 
+		nixio.syslog)
+    local resultTable = sharedState:get()
+    utils.printJson(resultTable)
+end
+
+local function get_from_sharedState_muti_writer(msg)
```
G10h4ck, line 25: `sed -i s/muti_writer/multi_writer/g` ;-)
also better not mixing up underscore (python like) and camel case, either keep _ o camel case but not both ;)

G10h4ck (changes requested):
See inline comments

G10h4ck (approved):

PR #1042, opened 2023-08-25
Add new reference state data types to shared state

javierbrk:
Network reference state will be defined by the UI. The data types must be supported and sync using shared state. This package provides two new data types one for WiFi links information and the other for nodes available in the network


packages/shared-state-reference-state/Makefile
```diff
@@ -0,0 +1,38 @@
...
+PKG_NAME:=shared-state-wifi_links_info
+PKG_VERSION=$(GIT_COMMIT_DATE)-$(GIT_COMMIT_TSTAMP)
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+	TITLE:=Reference state datatypes module for shared-state
+	CATEGORY:=LibreMesh
+	URL:=http://libremesh.org
+	DEPENDS:= shared-state
```
G10h4ck, line 22: are you sure it is ok with the space after `:=` anyway better get rid of it

packages/shared-state-reference-state/Makefile
```diff
@@ -0,0 +1,38 @@
...
+#
+# This is free software, licensed under the GNU Affero General Public License v3.
+#
+
+include $(TOPDIR)/rules.mk
+
+GIT_COMMIT_DATE:=$(shell git log -n 1 --pretty=%ad --date=short . )
+GIT_COMMIT_TSTAMP:=$(shell git log -n 1 --pretty=%at . )
+
+PKG_NAME:=shared-state-wifi_links_info
```
G10h4ck, line 13: Is the package name still appropriate?

G10h4ck (changes requested):
Maybe this should be integrated directly into lime app package?

Plus see inline comments

ilario:
Hi!
Two questions:
* would it be possible to use https://github.com/libremesh/lime-packages/blob/master/libremesh.mk for the Makefile? For example like here: https://github.com/libremesh/lime-packages/blob/e64db16c4e80f29b4687ad346c0e8d7833f7b62d/packages/lime-proto-bmx7/Makefile 
* this code is not specific for LibreMesh, so it should be on OpenWrt's repositories, not here. Are there plans for migrating this growing amount of shared-state code to OpenWrt?

G10h4ck:
Yeah using [libremesh.mk](https://github.com/libremesh/lime-packages/blob/master/libremesh.mk) is in usually a good idea, anyway I have talked with @javierajorge and we concluded the two uci-defaults files that would be the only real addition from this package should instead be included in lime-app as the reference state stuff is all lime-app specific, so this PR will be closed once @javierbrk and @selankon include the contributed code in the proper package (lime-app os some of it's modules).

about upstreaming shared-state code to OpenWrt it is a good idea but not for now, ATM some parts of it like the neighbour candidate discovery does assumptions that are quite tied to libremesh network model even if it doesn't depends on it directly, so we would risk to upstream something that works badly on a different setup, I am working now on a new version of shared-stare that will be a more reasonable candidate for upstreaming to OpenWrt.

Cheers!

packages/shared-state-reference-state/Makefile
javierbrk: ups copy paste, this make file is no longer nedded

javierbrk:
Here is an attempt to move the scripts into lime app package, what do you think ?

G10h4ck:
> Here is an attempt to move the scripts into lime app package, what do you think ?

Seems fine to me, let's wait to hear @selankon if he thinks this is the proper place where to ship those uci-defaults

G10h4ck (approved):

PR #1047, opened 2023-09-02
Default distance setting: increase 10x

### LINKED ISSUE #201: Set a better default distance parameter on the WiFi conifguration

ilario:
Fix #201

Goes against the decisions implemented in #202 and #324

See the discussion on #201 and on this email thread: https://lists.autistici.org/thread/20230815.142621.071b5fc0.en.html#i20230815.142621.071b5fc0

Here, the default distance setting for 5 GHz radios is being increased from 1 km to 10 km.
And the default distance setting for 2.4 GHz from 100 m to 1 km.


spiccinini (approved):

ilario:
Gathered no opinion against this since the start of the discussion via mail, merging.

PR #1048, opened 2023-09-09
add wifi interface name in shared state wifi information module 

javierbrk:
This pull request is a very small improvement witch adds the interface name for each link in shared state's wifi_information module


ilario:
Did not test but looks harmless, merging.

ilario (approved):

PR #1052, opened 2023-09-19
ci: update openwrt/gh-action-sdk

a-gave:
This is an attempt to enable the build of lime-packages via ci also for versions based on openwrt-22.03.5.

This create a distinct workflow file for older release , linking them with the sdk related to openwrt-19.07.10.

This update also the default workflow build.yml to fix the error `/entrypoint.sh: line 8: cd: /home/build/openwrt/: No such file or directory`. This seems due to an update in the directory structure in https://github.com/openwrt/gh-action-sdk

### DIFF
--- a/.github/workflows/build_2020.yml
+++ b/.github/workflows/build_2020.yml
@@ -1,13 +1,9 @@
 name: Build packages
 
 on:
-  release:
-    types: [published]
   push:
-    branches: [master]
     tags:
-      - v*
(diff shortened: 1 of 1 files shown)


ilario (approved):
Seems amazing! Thanks for looking into all of this! The `/entrypoint.sh: line 8: cd: /home/build/openwrt/: No such file or directory` error was breaking all of our CI unit testing!

PR #1056, opened 2023-09-28
Links information module for shared state according to babeld 

javierbrk:
This module incorporates links information extracted from babel using ubus infrastructure. 
The module has been tested in two librerouters and the output of the ` shared-state-async get babel_links_info` in my  setup is 

```JSON{
    "lrprimero": [
        {
            "dst_ip": "fe80::aa40:41ff:fe1f:73ab",
            "iface": "eth1-2_17",
            "src_ip": "fe80::aa40:41ff:fe1f:723b"
        },
        {
            "dst_ip": "fe80::c24a:ff:fefc:3abd",
            "iface": "eth1-2_17",
            "src_ip": "fe80::aa40:41ff:fe1f:723b"
        }
    ],
    "lrsegundo": [
        {
            "dst_ip": "fe80::aa40:41ff:fe1f:723b",
            "iface": "eth1-2_17",
            "src_ip": "fe80::aa40:41ff:fe1f:73ab"
        },
        {
            "dst_ip": "fe80::c24a:ff:fefc:3abd",
            "iface": "eth1-2_17",
            "src_ip": "fe80::aa40:41ff:fe1f:73ab"
        }
    ]
}
```


packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
```diff
@@ -0,0 +1,53 @@
...
+-- ! License, or (at your option) any later version.
+-- !
+-- ! This program is distributed in the hope that it will be useful,
+-- ! but WITHOUT ANY WARRANTY; without even the implied warranty of
+-- ! MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+-- ! GNU Affero General Public License for more details.
+-- !
+-- ! You should have received a copy of the GNU Affero General Public License
+-- ! along with this program.  If not, see <http://www.gnu.org/licenses/>.
+local JSON = require("luci.jsonc")
```
G10h4ck, line 20: Better add an empty line here, between license and includes

packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
```diff
@@ -0,0 +1,53 @@
...
+local utils = require('lime.utils')
+local ubus = require "ubus"
+
+local ifaceip = {}
```
G10h4ck, line 24: Why is this allocated outside the function?

packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
```diff
@@ -0,0 +1,53 @@
...
+
+function get_interface_ip(ifname)
```
G10h4ck, line 26: this function may fail, is it relevant to deal with the error conditions (for example an interface that miss IPv67 link local) ?

G10h4ck (changes requested):
See my inline comments

packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
javierbrk: done

packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
javierbrk: The idea is that each time the script is run this structure contains an empty list witch will be filled with each call to the get_ip method. Subsequent calls will use the information contained in the list.

packages/shared-state-babel_links_info/files/usr/bin/shared-state-publish_babel_links_info
javierbrk: I have added a new test, for ci and also tested locally, there is no problem with that kind of error, the result will be an empty string (non existing interfaces and no ipv6).  Do you think the function should report error explicitly ? If

G10h4ck:
should be ported to shared-state-async before merging

G10h4ck (approved):

PR #1027, opened 2023-05-27
Unit testing update

ilario:
On https://github.com/libremesh/lime-packages/pull/1012#issuecomment-1509146128 @spiccinini wrote that we should update the tests for working with the new LibreMesh code, as the tests were broken with the change from supporting OpenWrt 19.07 to OpenWrt 21 and 22.

I managed to fix some issues, but there are two that are still there:

1) The new OpenWrt style moves the list of the ports in the br-lan from a named interface to an unnamed device. In the tests, this stupid @device[0] thing does not get found:

```
Error → packages/lime-system/tests/test_lime_network.lua @ 77
LiMe Network tests test network.configure() with only lime.proto.lan
packages/lime-system/tests/test_lime_network.lua:100: attempt to index a nil value

stack traceback:
	packages/lime-system/tests/test_lime_network.lua:100: in function <packages/lime-system/tests/test_lime_network.lua:77>
```

The error comes from this line:
https://github.com/ilario/lime-packages/blob/c3e1b44764094b893e07b3a2bd62c964c7cbe5ee/packages/lime-system/tests/test_lime_network.lua#L100

On the hardware seems that there is no problem, as that @device[0] interface gets created by OpenWrt. In the tests, maaaaaaybe, the interface does not get added because there is no device yet? So that we should add a `uci add network device` somewhere? Where is the right place?

2) There is some issue with the LibreRouter testing. I don't think it is related to the LibreRouter, but I tried to update its files to recent OpenWrt style and something didn't work. No idea what is happening here.

```
Failure → tests/test_lime_config_device.lua @ 17
LiMe Config tests test lime-config for a LibreRouter device #librerouter
tests/test_lime_config_device.lua:44: Expected objects to be equal.
Passed in:
(nil)
Expected:
(string) 'lan1'

stack traceback:
	tests/test_lime_config_device.lua:44: in function <tests/test_lime_config_device.lua:17>
```

The error comes from this line:
https://github.com/ilario/lime-packages/blob/c3e1b44764094b893e07b3a2bd62c964c7cbe5ee/tests/test_lime_config_device.lua#L44

PR #1061, opened 2023-10-18
draft: fix bridge device section confusion

### LINKED ISSUE #1060: lime-proto-lan confuses bridge device section
After installing OpenWrt 23.05.0 with current master branch (https://github.com/libremesh/lime-packages/commit/4c51c7e062c3dff6c5218ee797d5a79c2ad6bc3d) onto an [avm fritz4020](https://openwrt.org/toh/avm/fritz.box.4020#avm_fritz_box_4020) I found the following in `/etc/config/network`:
```
config device                               
        option name 'eth1'                                
        option macaddr '44:4E:6D:11:87:D0'                
        list ports 'bat0'                   
        list ports 'eth0.1'                               
        list ports 'eth1'      
```
The ports should be added to the the bridge device further down. It probably has something to do with this line:
https://github.com/libremesh/lime-packages/blob/4c51c7e062c3dff6c5218ee797d5a79c2ad6bc3d/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua#L51-L53

pony1k:
Bridge ports are added to the first device section, which may be the wrong one. This changes the behaviour follwing @ilario s [suggestion](https://github.com/libremesh/lime-packages/issues/1060#issuecomment-1767832137). Now, the first device section that has type `bridge` and the name `br-lan` will be used.
After reading the `/bin/config_generate` script in OpenWrt base-files I am confident that there will always be such a device section on first boot.

fixes #1060

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/proto/lan.lua
@@ -48,9 +48,21 @@ function lan.setup_interface(ifname, args)
 
 	local uci = config.get_uci_cursor()
 	local bridgedIfs = {}
-	-- here we bet that the first device section is the bridge one,
-	-- as it does not have a name for addressing it
-	local oldIfs = uci:get("network", "@device[0]", "ports") or {}
+	-- here we bet that there is a device section of type bridge named 
+	-- br-lan
+	local bridge_section = nil
+	uci:foreach("network", "device",
+		function(s)
+			if bridge_section then return end
+			local dev_type = uci:get("network", s[".name"], "type")
+			local dev_name = uci:get("network", s[".name"], "name")
+			if not (dev_type == 'bridge') then return end
+			if not (dev_name == 'br-lan') then return end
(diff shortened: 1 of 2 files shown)


ilario:
Wow this looks veeerrrry nice!!!!!

And it also solves, in a more elegant way, the problem patched in https://github.com/libremesh/lime-packages/pull/1027/commits/6b72b3b385ae1989388d24ddf297058e4c218bce

Did you test on OpenWrt 22.03 also?

@spiccinini @G10h4ck check this out

G10h4ck (approved):

G10h4ck:
Seems good to me, thanks @pony1k

pony1k:
I tested it on
- fritzbox 4020, OpenWrt 23.05.0, non-dsa
- fritzbox 4040, OpenWrt 22.03.5, non-dsa
- TP-Link Archer C6 v3, OpenWrt 22.03.5, dsa
By tested I mean I removed all ports from the bridge device section and see if `lime-config` brings them back. For me is fine to be merged.

PR #1065, opened 2023-11-17
Update TESTING.md

selankon:
Add a warning advising that latest lime packages breaks the interfaces configurations

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -180,7 +180,7 @@ Install the package `qemu-system-x86_64` if you don't have already installed.
 > [!WARNING]
 > Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages break the qemu image and the interfaces cannot be setted up!
 > This is due to the migration to a newer version of OpenWRT. 
-> To use latest lime packages use a OpenWRT 23 image .
+> To use the latest lime packages, use a OpenWRT 23 image.
 
 ### Build a mesh network
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -178,7 +178,7 @@ Install the package `qemu-system-x86_64` if you don't have already installed.
 
 
 > [!WARNING]
-> Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages breaks the qemu image and the interfaces cannot be setted up!
+> Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages break the qemu image and the interfaces cannot be setted up!
 > This is due the migration to newer version of OpenWRT.
 > To use latest lime packages use a OpenWRT 23 image .
 
--- a/TESTING.md
+++ b/TESTING.md
@@ -179,7 +179,7 @@ Install the package `qemu-system-x86_64` if you don't have already installed.
 
 > [!WARNING]
 > Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages break the qemu image and the interfaces cannot be setted up!
-> This is due the migration to newer version of OpenWRT.
+> This is due to the migration to a newer version of OpenWRT. 
 > To use latest lime packages use a OpenWRT 23 image .
 
 ### Build a mesh network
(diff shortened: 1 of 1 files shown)


TESTING.md
```diff
@@ -176,6 +176,12 @@ Prebuilt development images can be downloaded from here:
 
 Install the package `qemu-system-x86_64` if you don't have already installed.
 
+
+> [!WARNING]
+> Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages breaks the qemu image and the interfaces cannot be setted up!
```
javierbrk, line 181:
```suggestion
> Since the commit https://github.com/libremesh/lime-packages/commit/64b7c01961c6e7bc776f83523aea9264724b3b23 lime packages break the qemu image and the interfaces cannot be setted up!
```

TESTING.md
```diff
@@ -176,6 +176,12 @@ Prebuilt development images can be downloaded from here:
...
+> This is due the migration to newer version of OpenWRT.
```
javierbrk, line 182:
```suggestion
> This is due to the migration to a newer version of OpenWRT. 
```

TESTING.md
```diff
@@ -176,6 +176,12 @@ Prebuilt development images can be downloaded from here:
...
+> To use latest lime packages use a OpenWRT 23 image .
```
javierbrk, line 183:
```suggestion
> To use the latest lime packages, use a OpenWRT 23 image.
```

TESTING.md
```diff
@@ -176,6 +176,12 @@ Prebuilt development images can be downloaded from here:
...
+
```
javierbrk, line 184: Maybe we can also add a link to a prebuilt OpenWRT 23 image.
```suggestion
> 
> New prebuilt images can be downloaded from here:
> * https://firmware-libremesh.antennine.campiinrete.org/releases/master-ow23/targets/x86/64/libremesh-master-ow23-default-x86-64-generic-rootfs.tar.gz
> * https://firmware-libremesh.antennine.campiinrete.org/releases/master-ow23/targets/x86/64/libremesh-master-ow23-default-x86-64-generic-initramfs-kernel.bin
```

javierbrk (commented):
Added some grammar changes and a suggestion to include a link to a prebuilt OpenWRT 23 image.

TESTING.md
selankon: Not sure if we should link non official repos here..

javierbrk: > Not sure if we should link non official repos here..

also further testing has to be done... Let's leave this text behind

javierbrk (approved):
I have confirmed the problem and this warning is very useful. 
These images seem to fix the problem   https://firmware-libremesh.antennine.campiinrete.org/releases/master-ow23/targets/x86/64/libremesh-master-ow23-default-x86-64-generic-rootfs.tar.gz
 https://firmware-libremesh.antennine.campiinrete.org/releases/master-ow23/targets/x86/64/libremesh-master-ow23-default-x86-64-generic-initramfs-kernel.bin

javierbrk (approved):

ilario:
In my opinion, the solution is not to add a warning, but to fix the issue.
If you don't want to put the link to an external website, you can either:

* ask @a-gave to copy those files on the official repo
* link these ones instead https://repo.libremesh.org/releases/2023.1-rc2-ow22/targets/x86/64/

TESTING.md
javierbrk: Added reference to official repo images

G10h4ck:
@selankon @javierbrk @ilario Is there a consensus on how this PR should be dealt with ?

javierbrk:
I think that we can put two links from official repos, the best we have so far are the mini that gotos built and mentioned in the [malinkg list ](https://lists.autistici.org/message/20240908.085937.479f1ea2.en.html) and then the defaults that still have a bug in shared state #1105 so I suggest: 
```suggestion
Prebuilt development images can be downloaded from https://downloads.libremesh.org/, just select a version and go get the files in targets/x86/64/ or just download: 

* https://downloads.libremesh.org/releases/2024.1-rc1-ow23.05.3-default/targets/x86/64/libremesh-2024.1-rc1-ow23.05.3-default-x86-64-generic-rootfs.tar.gz
* https://downloads.libremesh.org/releases/2024.1-rc1-ow23.05.3-default/targets/x86/64/libremesh-2024.1-rc1-ow23.05.3-default-x86-64-generic-initramfs-kernel.bin

Please note that this links are hard-coded and may contain firmware with known bugs like #1105.
```
What do you think @selankon ?

ilario:
I think this has been fixed by @VGDSpehar in #1171, please re-open if this PR is still valid.

PR #1067, opened 2024-01-20
re-implement shared-state from scratch in modern C++ 

G10h4ck:
Add shared-state-async package with almost complete shared-state modern C++ re-implementation, peer discovery is the only things that haven't been implemented yet in C++ and rely still on the shell script copied directly at compile time from the Lua shared-state package.

To show how to port the shared-state plugins to the new implementation also bat-hosts have been ported to the new implementation.


packages/shared-state-bat_hosts/tests/test_bat_hosts.lua
```diff
@@ -1,36 +1,39 @@
...
+--         self:insert({['52:00:00:ab:cd:00'] = "Lime_abcd04_wlan1_mesh"})
+--     end
+--     stub(shared_state.SharedState, "sync", syncMock)
+--     local bathost = bat_hosts.get_bathost('52:00:00:ab:cd:00', 'wlan1-mesh')
+--     assert.is.same(urlsOfSync, { ipv6ll })
+--     assert.is.equal(dataTypeOfSync, 'bat-hosts')
+--     assert.is.same({hostname='Lime-abcd04', iface='wlan1-mesh'}, bathost)
+--     assert.is_nil(bat_hosts.get_bathost('00:aa:bb:cc:dd:00', 'wlan1'))
+--     assert.is_nil(bat_hosts.get_bathost('00:aa:bb:cc:dd:00'))
+-- end)
```
selankon, line 39: Commented code. Should this tests work?

selankon (changes requested):

packages/shared-state-bat_hosts/tests/test_bat_hosts.lua
G10h4ck: nope, because it relies on Lua shared-state

javierbrk (approved):
Excellent, this is the first step, with the example provided porting of other modules should be trivial.

nicoechaniz (approved):
...

PR #1070, opened 2024-01-25
eupgrade: set cached_only to false

### LINKED ISSUE #1069: Eupgrade is_new_version_available always will return false
I noticed that if I use the RPC call:

```
ubus call eupgrade is_new_version_available '{}'
```

It will always return false (actually something like `{"status":"ok"})`. This is because of the RPC function will always ask for cached latest versions: 

```lua
local function is_new_version_available(msg)
    local cached_only = true
    local ret = {}
    if eupgrade.is_enabled() then
        local latest_data = eupgrade.is_new_version_available(cached_only)
        if latest_data then
            ret['version'] = latest_data.version
            ret["release_info_url"] = latest_data['release-info-url']
        end
    end
    ret.status = 'ok'
    utils.printJson(ret)
end
```
https://github.com/libremesh/lime-packages/blob/master/packages/eupgrade/files/usr/libexec/rpcd/eupgrade#L19-L30

The relevant code here is `local cached_only = true`, which mean, that on the function that check if new version is available, it check the cahced latest version, and i none, return false.  https://github.com/libremesh/lime-packages/blob/master/packages/eupgrade/files/usr/lib/lua/eupgrade.lua#L68-L70

I didn't find any place on the packages where actually look the non cached `eupgrade.is_new_version_available`, but maybe I am missing something. How can i force to download new information? 

@spiccinini is this really an error? 

I'll propose to set the `local cached_only = true` as default, but with the ability to call the `is_new_version_available` with a parameter to override it:

```
ubus call eupgrade is_new_version_available '{ "cached_only" : false }'
```

selankon:
Fixes #1069

### DIFF
--- a/packages/eupgrade/files/usr/libexec/rpcd/eupgrade
+++ b/packages/eupgrade/files/usr/libexec/rpcd/eupgrade
@@ -17,7 +17,7 @@ if not conn then
 end
 
 local function is_new_version_available(msg)
-    local cached_only = true
+    local cached_only = false
     local ret = {}
     if eupgrade.is_enabled() then
         local latest_data = eupgrade.is_new_version_available(cached_only)


spiccinini:
Did you try with the lime-app? Forcing no cache will slow the load of the app right @germanferrero ?

selankon:
@spiccinini the react query that check for updates should not block the ui. 

BTW, we could use a background process or something like that, to check the updates, and let this just set to false? I am not familiarized with OWRT systems, have they something similar to services?

packages/eupgrade/files/usr/libexec/rpcd/eupgrade
```diff
@@ -70,4 +70,4 @@ if arg[1] == 'call' then
     elseif  arg[2] == 'download_status' then download_status(msg)
     else utils.printJson({ error = "Method not found" })
     end
-end
+end
```
G10h4ck, line 73: missing new line at the end of file

G10h4ck (changes requested):

G10h4ck:
I would expect from a function named like that to not return a cached result.
Do the sub-called function trigger the download?
If this change slow down the UI, I would suggest to split triggering the download into another function

spiccinini:
> Do the sub-called function trigger the download?

AFAIK no. The issue is that until the http GET does not return it will block (and if you don't have internet it will timeout in more than a few seconds).

spiccinini:
> @spiccinini the react query that check for updates should not block the ui.

You mean that the current code does not block or that it should not? In some version of the  lime-app it blocked the main page of the app, if you tested something like even adding a sleep(30) in the lua file does not block then good!.

ilario:
This edit has been included by @selankon in #1101 so I would close this pull request.
If you consider that the discussion was not finished and that the change should be reverted, please re-open this or open a new issue.

PR #1080, opened 2024-02-11
Fix lime-config fail when there is no lower iface

pony1k:
On a mercusys mr70x, lime-config does not work. It throws the following error:
```
network.scandevices.owrt_network_interface_parser found ifname wan
ls: /sys/class/net/wan/lower*: No such file or directory
/usr/lib/lua/lime/network.lua:223: attempt to index field '?' (a nil value)
stack traceback:
        /usr/lib/lua/lime/config.lua:221: in function </usr/lib/lua/lime/config.lua:221>
        [C]: in function 'foreach'
        /usr/lib/lua/lime/network.lua:339: in function 'scandevices'
        /usr/lib/lua/lime/network.lua:373: in function </usr/lib/lua/lime/network.lua:364>
        [C]: in function 'xpcall'
        /usr/lib/lua/lime/config.lua:221: in function 'main'
        /usr/bin/lime-config:55: in main chunk
        [C]: ?
```
It seems that an interface that is called "wan" does not always have link to a lower interface in /sys/class/net. This is how I would like to fix it. I've tested it already on a mercusys mr70x and it is working fine.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -216,7 +216,8 @@ end
 function network._get_lower(dev)
     local lower_if_path = utils.unsafe_shell("ls -d /sys/class/net/" .. dev .. "/lower*")
     local lower_if_table = utils.split(lower_if_path, "_")
-    return lower_if_table[#lower_if_table]:gsub("\n", "")
+    local lower_if = lower_if_table[#lower_if_table]
+    return lower_if and lower_if:gsub("\n", "")
 end
 
 function network.scandevices()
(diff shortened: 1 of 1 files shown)


ilario:
Uh... Unexpected. Do you have idea if this is a new thing of OpenWrt 23?
Anyway, the code looks good to me.
After this change, does the WAN work?

pony1k:
I believe this is because `wan` is not a switch port, but connected directly to the CPU. See here for `mr70x`:
https://github.com/openwrt/openwrt/blob/6497cdba09d6faa04a4f74465ed95ae21e2ff1ff/target/linux/ramips/dts/mt7621_mercusys_mr70x-v1.dts#L134-L171
While for the `archer-c6` as an example:
https://github.com/openwrt/openwrt/blob/6497cdba09d6faa04a4f74465ed95ae21e2ff1ff/target/linux/ramips/dts/mt7621_tplink_archer-x6-v3.dtsi#L182-L218

The `wan` port works. I checked.

ilario:
Anyway it is something new... I didn't see such messages when using OpenWrt 19.07 (cannot remember if I used also OpenWrt 22) on YouHua WR1200JS that has a very similar DTS... Mah...

ilario (approved):

G10h4ck:
Looks good to me too, thanks @pony1k

ilario:
Sorry for not testing before.
In your case, @pony1k, did this work for activating the dhcp for the wan entry in the /etc/config/network file?
I tried with a YouHua WR1200JS and the dhcp is indicated in /etc/config/lime-autogen but not in /etc/config/network.

So I think we should add something like this "else" part:

```
if lower_if then                                                               
  devices[lower_if] = { nobridge = true }                                
  devices[dev] = {}                                                      
  utils.log( "network.scandevices.dev_parser found WAN port %s " ..      
  "and marking %s as nobridge", dev, lower_if )               
else                                                                           
  devices[dev] = {}                                                      
  utils.log( "network.scandevices.dev_parser found WAN port %s", dev)    
end
```

pony1k:
You're right. If I had tried activating dhcp for the wan entry, it would probably have failed.

What you are proposing looks good.

Issue #1081, opened 2024-02-11
shared-state-async: errors in log


pony1k:
This log message is repeated about once per second, with a few seconds break somtimes:
```
Sun Feb 11 21:07:31 2024 daemon.err shared-state-async[9329]: E 1707685651.973 static void RsTypeSerializer::serial_process(RsGenericSerializer::SerializeJob, RsGenericSerializer::SerializeContext&, std::map<T, U>&, const std::string&) [with T = std::__cxx11::basic_string<char>; U = SharedState::DataTypeConf; std::string = std::__cxx11::basic_string<char>] FROM_JSON operation for std::map mTypeConf expect it to be empty, clearing it now please report to developers
Sun Feb 11 21:07:31 2024 daemon.err shared-state-async[9329]: print_stacktrace Not implemented yet for this platform
```
bat_hosts are synced fine. The shared-state-async package was built from current master (0c8a915e95c2d5bbf6a7c9225bd012888bdd72b9).

PR #1082, opened 2024-02-12
shared-state-async fix log pollution

G10h4ck:
Fix https://github.com/libremesh/lime-packages/issues/1081 log pollution reported by @pony1k


pony1k:
This works. Thanks!

ilario (approved):

PR #1085, opened 2024-02-13
Document how to set ethernet interfaces for mesh only or clients only.

pony1k:
This adds some documentation on how to set ethernet interfaces for mesh only or clients only.

See for example here: https://github.com/libremesh/network-profiles/tree/master/calafou#lime-community-configuration-3


ilario (approved):
Very needed! Thanks!

PR #1086, opened 2024-02-19
Shrared state Async ubus interface 

javierbrk:
This pull request adds shrare-state-async ubus interface. Get method returns sharedstate "get"  json data or reported error in case no json is found. Errors are bubbled up as they come from sharedstate inside a json structure with the name "error".  

Since there is no Shared State async binary available in the testing environment, testing possibilities are reduced. Manual testing can be done on a router with bat-hosts package using this commands:

Get bat-hosts
```bash
# ubus -S call shared-state-async get "{'data_type': 'bat-hosts'}"
{"a8:40:41:1f:73:ab":"lrsegundo_eth1_2","02:95:39:1f:73:aa":"lrsegundo_eth0_250","a8:40:41:1c:85:c3":"lrsegundo_wlan2_mesh","a8:40:41:1c:85:16":"lrsegundo_wlan1_mesh","02:bb:ed:1f:73:aa":"lrsegundo_eth1_250","a8:40:41:1f:73:a8":"lrsegundo_wlan0_ap","aa:40:41:1f:73:a8":"lrsegundo_wlan0_apname","02:db:d6:1f:73:aa":"lrsegundo_eth0_1_250","02:29:0f:1f:73:aa":"lrsegundo_eth1_2_250","c2:10:20:5e:7f:b3":"lrsegundo_bat0","a8:40:41:1f:73:aa":"lrsegundo_eth0"}

``` 
Get an invalid data type
```bash
# ubus -S call shared-state-async get "{'data_type': 'bat-hosss'}"
{"error":53248}
```

Sync valid data type 
```bash 
# ubus -S call shared-state-async sync "{'data_type': 'bat-hosts'}"
{"error":0}

```

Sync invalid data type  
```bash 
ubus -S call shared-state-async sync "{'data_type': 'bat-hoss' }"
{"error":53248}
```

Sync valid data type with unreachable ipv4 addresses
```bash 
# ubus -S call shared-state-async sync "{'data_type': 'bat-hosts' ,'peers_ip':['10.0.0.1','10.0.0.2']}'"
{"error":32768}

```
Sync valid data type with invalid ipv4 address  
```bash
# ubus -S call shared-state-async sync "{'data_type': 'bat-hosts' ,'peers_ip':['10.0.0.1','10.0.2']}'"
{"error":61952}
```
Sync invalid data type with specified ipv4 address  
```bash
ubus -S call shared-state-async sync "{'data_type': 'bat-hosdts' ,'peers_ip':['127.0.0.1','127.0.0.1']}'"
{"error":53248}
```

Sync valid data type with reachable ipv4 addresses
```bash 
# ubus -S call shared-state-async sync "{'data_type': 'bat-hosts' ,'peers_ip':['127.0.0.1','127.0.0.1']}'"
{"error":0}
```


packages/shared-state-async/files/usr/share/rpcd/acl.d/shared-state-async.json
```diff
@@ -0,0 +1,15 @@
...
+                    "shared-state-async": [ "*" ]
+            }
+        },
+        "write": {
+            "ubus": {
+                    "shared-state-async": [ "*" ]
+            }
+        }
+    }
+}
```
G10h4ck, line 15: missing  new line at the end of file

packages/shared-state-async/files/usr/share/rpcd/acl.d/shared-state-async.json
```diff
@@ -0,0 +1,15 @@
+{
+    "lime-app": {
+        "description": "lime-app public access",
```
G10h4ck, line 3: why do we explicitely mention lime-app here? other components could benefit to access shared-state-async via ubus too

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
```diff
@@ -0,0 +1,84 @@
...
+
+This is free software, licensed under the GNU AFFERO GENERAL PUBLIC LICENSE Version 3
+]]
+--
+local utils = require('lime.utils')
+local json = require 'luci.jsonc'
+
+local ERROR_LOG_FILE_NAME = "/tmp/shared_state_rpcd_error.log"
+
+local function parseErrors()
```
G10h4ck, line 17: This way of handling errors doesn't seems robust to me :-/, what if a few calls are performed in parallel and some fails and some not?
shared-state-async command return the error code to the calling shell if some error happens, in that case the error should be bubbled up without more parsing

G10h4ck: if this prove too cumbersome to do in Lua, consider rewriting the whole ubus plugin using ucode, or ash which are supported by default by OpenWrt

G10h4ck (changes requested):
See my inline comments

packages/shared-state-async/files/usr/share/rpcd/acl.d/shared-state-async.json
javierbrk: While logged in via ssh, you have direct, full access to ubus. This file is to enable lime app to use it. other users should enable other profiles.

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
javierbrk: commit e891527014394e7c0b06b1554a14fb0727e49810 implements a new way of handling, what do you think?

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
```diff
@@ -0,0 +1,60 @@
...
+This is free software, licensed under the GNU AFFERO GENERAL PUBLIC LICENSE Version 3
+]]
+--
+local utils = require('lime.utils')
+local json = require 'luci.jsonc'
+
+local function get(msg)
+    local error = os.execute("shared-state-async get " ..
+        msg.data_type .. " 2>/dev/null ")
+    utils.printJson({
```
G10h4ck, line 19: That the error get printed only on failure happens implicitely, please add at least a comment documentig this behaviour, so we cna understand what's happening when we look at this code in a few months or years ;)

G10h4ck (changes requested):

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
javierbrk: Done ! it is a weird behavior that gives a lot of headaches. Any output that is printed in the console is used as output. If its a json everything goes fine, if it is not a json you get "invalid input" as a response.

G10h4ck (approved):

G10h4ck (approved):

PR #1087, opened 2024-02-22
Add shared-state-async network statistics sharing

G10h4ck:


ilario:
We should keep compatibility (or at least avoid breaking it explicitly)
with OpenWrt 21 at least until we make a release based on OpenWrt 23, in my
opinion.
Please use random_numgen instead of $RANDOM.
See https://github.com/libremesh/lime-packages/issues/800


-- 
Inviato dal mio smartstone

G10h4ck:
This stuff (and in general what is meant for mesh-wide milestone) is developed on top (and depends) on OpenWrt development branch, as an example `shared-state-async` depends on a GCC version which is not shipped yet in OpenWrt stable releases, otherwise it will not compile properly.
So doing the change you suggests hoping for retro-compatibility would be futile.

G10h4ck:
other even more retro-compatibility breaking stuff is coming to out of my experimentation to have better support for newer radios https://www.youtube.com/live/DOH5FBMTXmE?feature=shared&t=10848

ilario:
Ok, will it work at least on OpenWrt 23?
If not, please do not propose those changes on the master branch, which is being used for preparing a release based on OpenWrt 23 (we could also create a branch for the release...).

G10h4ck:
AFIU it will not work on OpenWrt 23 either, @javierbrk tried to compile and failed because of not enough recent compiler AFAIR

Better creating a branch for the planned release and backport there stuff if needed, while development keep staying on master branch

G10h4ck:
Ain't got time for that last compatible commit should be this https://github.com/libremesh/lime-packages/commit/a5eb7d888b681e4cab81289949d809ee0e968880

pony1k:
> AFAIU it will not work on OpenWrt 23 either, @javierbrk tried to compile and failed because of not enough recent compiler AFAIR

I compiled shared-state-async from a823f4727e14297d1c8d7a07cdee18ef26246fb1 for OpenWrt 23.05.2 and it is working.

javierbrk:
Hi ! My compilation errors were using gcc 11 shipped with openwrt 22.03 !

Saludos
Ing. Javier Alejandro Jorge

PR #1089, opened 2024-02-27
ci: rebuild packages via sdk only when a package changes

a-gave:
This prevents all packages from being recompiled when not needed (changes to readme, changelog, other not related files)


ilario (approved):
Makes a lot of sense!

PR #1091, opened 2024-03-01
ci: add multi-arch-build.yml;

a-gave:
Add a new packages repository at:
- http://github.com/libremesh/lime-feed/tree/gh-pages/master/packages/$ARCH_PACKAGES
- http://feed.libremesh.org/master/packages/$ARCH_PACKAGES 

That includes packages precompiled by openwrt-sdk for each specific architecture

This is currently required only for the package `shared-state-async` 
After this merge https://github.com/libremesh/lime-packages/pull/1067, imagebuilder builds fail, using the default set of packages, unless you are building for x86_64

update the imagebuilder build guide on `README.md`
update `lime-system/files/etc/uci-defaults/92_add-lime-repos` to use also the new repo

### DIFF
--- a/.github/workflows/multi-arch-build.yml
+++ b/.github/workflows/multi-arch-build.yml
@@ -1,0 +1,78 @@
+name: Multi arch build packages
+
+on:
+  release:
+    types: [published]
+  push:
+    branches: [master]
+    tags:
+      - v*
+      - '!v2020*'
+    paths:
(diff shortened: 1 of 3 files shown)


ilario (approved):
Did not test yet, but we need to have the CI to run this in order to test this. So let's merge and let's see if it works as expected. Thanks @a-gave !!!!!!!!!

Issue #1093, opened 2024-03-11
Bat-hosts get_bathost error


selankon:
Trying libremesh with OWRT23, me and other people noticed that bat host doesn't load correctly a bathost data:

```
root@LiMe-eb7f64:~# ubus  call bat-hosts get_bathost '{"mac": "D2:5F:25:EB:63:DA", "outgoing_i
face": "wlan1-mesh" }'
Command failed: No response
root@LiMe-eb7f64:~# echo '{"mac": "D2:5F:25:EB:63:DA", "outgoing_iface": "wlan1-mesh" }' | /us
r/libexec/rpcd/bat-hosts call get_bathost | jq
Failed to allocate uclient context
lua: /usr/libexec/rpcd/bat-hosts:31: attempt to index local 'bathost' (a nil value)
stack traceback:
	/usr/libexec/rpcd/bat-hosts:31: in function 'get_bathost'
	/usr/libexec/rpcd/bat-hosts:52: in main chunk
	[C]: ?
```

It seems that is not retrieving the info or nill values are not handled properly.  

This causes a "Fetching name" loader continuously shown on the lime app landing page.

ilario:
Who can investigate this?
@javierajorge can you have a look? Thanks!

ilario:
According to this comment
https://github.com/libremesh/lime-packages/pull/1094#issuecomment-2000385170 this has been fixed. Please @javierbrk @selankon open it again if I misunderstood.

PR #1094, opened 2024-03-15
bat_hosts: fixes rpcd error

javierbrk:
Rpcd get bat hosts call was not working causing lime app UI to fail when trying to retrieve host names. 

This pull request fixes the rpc call. 

the function "bat_hosts.get_bat_hosts_from_shared_state()" was created for testing purposes. 

```bash
 ubus call bat-hosts get_bathost '{"mac":"AE:40:41:1C:85:16"}'
{
        "status": "ok",
        "bathost": {
                "iface": "wlan1-mesh_17",
                "hostname": "LiMe-8037b1"
        }
}
```

### DIFF
--- a/packages/shared-state-bat_hosts/files/usr/lib/lua/bat-hosts.lua
+++ b/packages/shared-state-bat_hosts/files/usr/lib/lua/bat-hosts.lua
@@ -1,4 +1,3 @@
-local shared_state = require("shared-state")
 local JSON = require("luci.jsonc")
 local utils = require("lime.utils")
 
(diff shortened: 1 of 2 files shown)


G10h4ck:
shouldn't lime-app ported to shared-state-async ubus interface instead? so we can drop this plugin altogether?

javierbrk:
this is just a quick fix to #1093 It will bring consistency to the limeapp version linked in the repository.
I think that the other alternative requires a new release of the lime app.

G10h4ck (approved):

PR #1097, opened 2024-03-19
A few cleanups to packages

G10h4ck:
Remove broken un-fixable packages
Fix pirania circular dependency


selankon (approved):

ilario:
Thanks, there was some need for cleaning.

I suspect there are people using LuCI (maybe @digitigrafo ?) but I suppose that the LuCI modules we have here do not work anymore with recent LuCI?

@G10h4ck can you remove the delete packages from the [libremesh.sdk.config](https://github.com/libremesh/lime-packages/blob/f7ff091a6e5ec908b34ae3f2bdfa157830ca7afe/libremesh.sdk.config#L10) file? Or delete the file entirely if you consider it useless... (I cannot remember if it is useful or not)

G10h4ck:
> @G10h4ck can you remove the delete packages from the [libremesh.sdk.config](https://github.com/libremesh/lime-packages/blob/f7ff091a6e5ec908b34ae3f2bdfa157830ca7afe/libremesh.sdk.config#L10) file? Or delete the file entirely if you consider it useless... (I cannot remember if it is useful or not)

I think that file is not useful, let's see if someone complain about it being broken in a while and then remove it altogether if not

PR #1098, opened 2024-03-20
safe-upgrade: update calibration data partition name

G10h4ck:
DISCLAIMER: Breaks safe-upgrade retrocompatibility!

Newer OpenWrt rely on calibration data MTD partitions being named 'art' older OpenWrt on which librerouterOS 1.5 is based, looked for calibration calibration data for on-SoC radio on 'ART' partition. In particular librerouter v1 have an on SoC radio which have calibration data in art partition, safe-upgrade can either support older OpenWrt version or newer, but not both at least without change the code drastically. Here in development branch let's support new OpenWrt.

### DIFF
--- a/packages/safe-upgrade/files/usr/sbin/safe-upgrade
+++ b/packages/safe-upgrade/files/usr/sbin/safe-upgrade
@@ -346,7 +346,7 @@ local function upgrade(args)
     else
         fw_mtd_str = '7936k(firmware),7936k(fw2)'
     end
-    local boot_script_tpl = 'set bootargs console=ttyS0,115200 board=LIBREROUTERV1 mtdparts=spi0.0:256k(u-boot),64k(u-boot-env),%s,128k(res),64k(ART); echo booting part %d; bootm ${fw_addr};'
+    local boot_script_tpl = 'set bootargs console=ttyS0,115200 board=LIBREROUTERV1 mtdparts=spi0.0:256k(u-boot),64k(u-boot-env),%s,128k(res),64k(art); echo booting part %d; bootm ${fw_addr};'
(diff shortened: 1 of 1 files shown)


selankon (approved):

spiccinini:
When I hardcoded the partiions name years ago I tought that if the partitions name changed then they should come with the upgrade file so that it is not hardcoded here. So maybe sesrching for a specific env variable or file somewhere and use that or default to the hardcoded values is valuable?

G10h4ck:
One of the big problems is that such environments variables or config files where we could get such information, differ between the running system (the one where `safe-upgrade` is executed) and the image one is going to flash, so the running `safe-upgrade` is the old one in the old image and can access the information in the old image, while we would need the information stored in the new image, unless we mount the new image and look into it do discover such information we cannot de-hardcode the info.

spiccinini:
eupgrade supports an `installer` type that we used in the past to provide this kind of upgrades. The installer is just a .sh script that can also contain the image binary. 
I created this tool https://github.com/spiccinini/shellbinscript  that embeds files inside the installer.sh script. In that way we provided a new safe-upgrade to not rely in the one already installed. 
In this repo there are some install scripts we used in the past: https://gitlab.com/librerouter/lros-installersh-scripts/-/tree/v1.5?ref_type=heads

So the idea can to just use the installer type, providing the bootargs alongside the image. Other idea is to put the bootargs as metadata of the image (the image currently has the board name as metadata)

ilario:
Should we pick this to the 2024.1 branch? Is this good or bad for the next release candidate?

G10h4ck:
@ilario this is needed if you plan to use safe-upgrade with librerouter v1 devices based on OpenWrt current release

PR #1100, opened 2024-03-25
ci: multi-arch-build: use a different output path

a-gave:
I apologize for bothering again on this topic, but compiling via imagebuilder once again turns out to be broken.

I did not calculate that a subsequent compilation from the build.yml workflow would delete the subdirectory previously created by the multi-arch-build.yml workflow (e.g. now missing https://github.com/libremesh/lime-feed/tree/gh-pages/master/packages).

This fixes it, telling the multi-arch-build.yml workflow to use a different path for output packages to avoid overwriting.
This updates also `packages/lime-system/files/etc/uci-defaults/92_add-lime-repos` and `README.md` accordingly.

This also removes the feeds added by default by buildroot for each feed declared in feeds.conf and resulting in a known opkg error such as the one below:
```
root@ninux-598775:~# opkg update
[...]
Collected errors:
 * opkg_download: Failed to download http://downloads.openwrt.org/releases/19.07.10/packages/mips_24kc/libremesh/Packages.gz, wget returned 8.
 * opkg_download: Failed to download http://downloads.openwrt.org/releases/19.07.10/packages/mips_24kc/profiles/Packages.gz, wget returned 8.
 ```

### DIFF
--- a/.github/workflows/multi-arch-build.yml
+++ b/.github/workflows/multi-arch-build.yml
@@ -83,5 +83,5 @@ jobs:
           deploy_key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
           external_repository: libremesh/lime-feed
           publish_dir: bin/packages/${{ matrix.arch }}/libremesh/
-          destination_dir: ${{ env.DEST_DIR }}/packages/${{ matrix.arch }}
+          destination_dir: arch_packages/${{ env.DEST_DIR }}/${{ matrix.arch }}
 
--- a/README.md
+++ b/README.md
@@ -65,12 +65,12 @@ Within the container, add the `lime-packages` feeds:
 
 ```shell
 echo "src/gz libremesh https://feed.libremesh.org/master" >> repositories.conf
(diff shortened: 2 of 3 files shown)


ilario:
Thanks!
It looks great :)

And it also allows us to remove one step from the compilation instructions,
right?
I refer to this one:
"Deselect erroneous feeds locations:

Image configuration → Separate feed repositories → Enable feed libremesh

Image configuration → Separate feed repositories → Enable feed profiles"

from this page:
https://libremesh.org/development.html#compiling_libremesh_from_source_code

Thanks!




-- 
Inviato dal mio smartstone

a-gave:
> Thanks! It looks great :) And it also allows us to remove one step from the compilation instructions, right? I refer to this one: "Deselect erroneous feeds locations: Image configuration → Separate feed repositories → Enable feed libremesh Image configuration → Separate feed repositories → Enable feed profiles" from this page: https://libremesh.org/development.html#compiling_libremesh_from_source_code Thanks!

On second thought, it seems more correct to me that it is deselected at build time.

ilario:
Ok! Both options (to instruct the user to deselect that or to do it automatically) seem good to me.

ilario (approved):

PR #1103, opened 2024-04-08
Adds shared-state rpcd data,error output format and shared-state-async rpcd reimplementation 

javierbrk:
# Ubus interface for shared state 

current implementations of rpcd for shared-state are inconsistent this new implementation adds a consistent format across the three implementations (shared-state, shared-state-async, shared-state-multiwriter ) 
In every case calls to shared state get or sync will return a json object wit this format 

```JSON 
{"data":{},"error":0}
```

Also this new implementation for shared-state-async uses a shell script to simplify the implementation 

## Ubus interface for shared state async

ubus -S call shared-state-async get '{"data_type": "bat-hosts"}'
```JSON 
{
  "data": {
    "02:29:0f:a5:1e:d1": "LiMe_a51ed1_eth1_2_29",
    "02:58:47:a5:1e:d1": "LiMe_a51ed1_wlan0_mesh_29",
    "02:95:39:a5:1e:d1": "LiMe_a51ed1_eth0_29",
    "02:ab:46:a5:1e:d1": "LiMe_a51ed1_wlan1_mesh_29",
    "02:bb:ed:a5:1e:d1": "LiMe_a51ed1_eth1_29",
    "02:cc:4e:a5:1e:d1": "LiMe_a51ed1_wlan2_mesh_29",
    "02:db:d6:a5:1e:d1": "LiMe_a51ed1_eth0_1_29",
    "06:ef:f4:3d:4d:75": "LiMe_a51ed1_bat0",
    "a8:40:41:1c:85:16": "LiMe_a51ed1_wlan1_ap",
    "a8:40:41:1c:85:c3": "LiMe_a51ed1_wlan2_ap",
    "a8:40:41:1f:73:a8": "LiMe_a51ed1_wlan0_ap",
    "a8:40:41:1f:73:aa": "LiMe_a51ed1_eth0_1_17",
    "a8:40:41:1f:73:ab": "LiMe_a51ed1_eth1_17",
    "aa:40:41:1c:85:16": "LiMe_a51ed1_wlan1_apname",
    "aa:40:41:1c:85:c3": "LiMe_a51ed1_wlan2_apname",
    "aa:40:41:1f:73:a8": "LiMe_a51ed1_wlan0_apname",
    "ae:40:41:1c:85:16": "LiMe_a51ed1_wlan1_mesh",
    "ae:40:41:1c:85:c3": "LiMe_a51ed1_wlan2_mesh",
    "ae:40:41:1f:73:a8": "LiMe_a51ed1_wlan0_mesh"
  },
  "error": 0
}
```

ubus -S call shared-state-async get '{"data_type": "bat-hosss"}'
```JSON 
{"data":{},"error":208}
```
ubus -S call shared-state-async sync '{"data_type": "bat-hosts"}'
```JSON 
{"data":{},"error":146}
```
ubus -S call shared-state-async sync '{"data_type": "bat-hosss"}'
```JSON 
{"data":{},"error":208}
 ```
ubus -S call shared-state-async sync '{"data_type": "bat-hostss", "peers_ip":"10.13.30.20"}'
```JSON 
{"data":{},"error":208}
```
ubus -S call shared-state-async sync '{"data_type": "bat-hosts", "peers_ip":"10.13.30.20"}'
```JSON 
{"data":{},"error":148}
```
ubus -S call shared-state-async sync '{"data_type": "bat-hosts", "peers_ip":"127.0.0.1"}'
```JSON 
{"data":{},"error":0}

```
## Ubus interface for shared state and shared state Multiwriter

Get the list of available methods using:

ubus -v list shared-state

Get wifi links information: 

ubus -S call shared-state getFromSharedState '{"data_type": "wifi_links_info" }'

Write information that is common to every all the network:

ubus -S call shared-state insertIntoSharedStateMultiWriter '{"data_type": "ref_state_wifi_links", "json": {"primero":{"bleachTTL":27}}}'

Get the information in the same node:

ubus -S call shared-state getFromSharedStateMultiWriter '{"data_type": "ref_state_wifi_links"}'
```JSON 
{"data":{"primero":{"bleachTTL":27}},"error":0}
```

Get the info in another node, before getting the information the others need to sync on the same topic.
This has to be done before reading and data types usually do it in a cron job. 

shared-state-multiwriter sync ref_state_wifi_links

ubus -S call shared-state getFromSharedStateMultiWriter '{"data_type": "ref_state_wifi_links" }'

### DIFF
--- a/packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
+++ b/packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
@@ -1,13 +1,11 @@
 #!/bin/sh
-# --[[
-# Shared State Async
 
+# Shared State Async
 # Copyright (c) 2024 Javier Jorge <jjorge@inti.gob.ar>
 # Copyright (c) 2024 Instituto Nacional de Tecnologia Industrial
 # Copyright (C) 2024 Asociacion Civil Altermundi <info@altermundi.net>
-
 # SPDX-License-Identifier: AGPL-3.0-only
-# ]]
+
 
 sinc_args=""
  
@@ -18,8 +16,8 @@ case "$1" in
     call)
         # source jshn shell library
         . /usr/share/libubox/jshn.sh
-        read nmsg
-        json_load $nmsg
+        read msg
+        json_load $msg
         json_get_vars data_type 
         json_get_vars peers_ip
         case "$2" in 
--- a/packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
+++ b/packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
@@ -1,63 +1,43 @@
-#!/usr/bin/env lua
+#!/bin/sh
+# --[[
+# Shared State Async
 
---[[
-Shared State Async
+# Copyright (c) 2024 Javier Jorge <jjorge@inti.gob.ar>
+# Copyright (c) 2024 Instituto Nacional de Tecnologia Industrial
+# Copyright (C) 2024 Asociacion Civil Altermundi <info@altermundi.net>
 
-Copyright (c) 2024  Javier Jorge <jjorge@inti.gob.ar>
-Copyright (c) 2024  Instituto Nacional de Tecnolog..a Industrial
-Copyright (C) 2024  Asociacion Civil Altermundi <info@altermundi.net>
+# SPDX-License-Identifier: AGPL-3.0-only
+# ]]
 
-This is free software, licensed under the GNU AFFERO GENERAL PUBLIC LICENSE Version 3
-]]
---
-local utils = require('lime.utils')
-local json = require 'luci.jsonc'
-
-local function get(msg)
-    local error = os.execute("shared-state-async get " ..
-        msg.data_type .. " 2>/dev/null ")
-        -- if os.execute dont fail will print the output and rpcd wont execute
-        -- the following lines. If there is an error the above code wont print
-        -- anything and this code will return the error code.
-    utils.printJson({
-        error = error
-    })
-end
-
-local function sync(msg)
-    local error = os.execute("shared-state-async sync " ..
-            msg.data_type .. " " .. table.concat(msg.peers_ip or {}, " ") .. " 2>/dev/null ")
-    utils.printJson({
-      error = error
-    })
-end
-
---{"data_type":"data","peers_ip":["10.0.0.1","10.0.0.2"]}
---{"data_type":"data","peers_ip":["10.0.0.1"]}
-local methods = {
-    get = {
-        data_type = 'value'
-    },
-    sync = {
-        data_type = 'value',
-        peers_ip = 'value'
-    }
-}
-
-if arg[1] == 'list' then
-    utils.printJson(methods)
-end
-
-if arg[1] == 'call' then
-    local msg = utils.rpcd_readline()
-    msg = json.parse(msg)
-    if arg[2] == 'get' then
-        get(msg)
-    elseif arg[2] == 'sync' then
-        sync(msg)
-    else
-        utils.printJson({
-            error = "Method not found"
-        })
-    end
-end
+sinc_args=""
+ 
+case "$1" in 
+    list) 
+        echo '{ "sync": { "data_type": "str", "peers_ip": "str" }, "get": { "data_type": "str" } }'
+    ;; 
+    call)
+        # source jshn shell library
+        . /usr/share/libubox/jshn.sh
+        read nmsg
+        json_load $nmsg
+        json_get_vars data_type 
+        json_get_vars peers_ip
+        case "$2" in 
+            get) 
+                if output=$(shared-state-async get $data_type 2>/dev/null) 
+                then 
+                    echo {\"data\": $output , \"error\" :$?}
+                else 
+                    echo {\"data\": {} , \"error\": $? } 
+                fi 
+            ;; 
+            sync) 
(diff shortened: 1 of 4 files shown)


ilario:
This looks good :D

Do you think we should include it in the next release candidate once it is merged?

Do you plan to move all this to OpenWrt repositories at some point?

selankon:
> Do you think we should include it in the next release candidate once it is merged?

imo should be included! This makes the shared state calls more consistent... 

> Do you plan to move all this to OpenWrt repositories at some point?

I think shared state not exists on the OWRT repos no?

selankon (approved):
lgtm!

ilario:
> > Do you think we should include it in the next release candidate once it is merged?
> 
> imo should be included! This makes the shared state calls more consistent...
> 
> But, it will break the compatibility with older versions. Are this packages already included on the actual version?

Compatibility?
Currently we have the latest stable release 2020.4 based on OpenWrt 19.07 and including LibreMesh code from more than one year ago. Most importantly, it has the legacy non-async shared-state. So the compatibility with this one I feel is not going to be preserved anyway, no?

> > Do you plan to move all this to OpenWrt repositories at some point?
> 
> I think shared state not exists on the OWRT repos no?

Shared state does not exist on OpenWrt repository, correct. But it should be there. It should not be here. In case there is any shared-state developer willing to manage the migration, maybe some people from OpenWrt could help...? @aparcar

G10h4ck:
@ilario shared-state developers and it is pretty tied to libre-mesh stuff altought it might be used indipendently, at this stage of development, moving it to openwrt repository could be premature IMHO, but once shared-state-async is more stable and completely substitute shared-state Lua, it might be good to move it in openwrt-packages

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
```diff
@@ -1,63 +1,43 @@
...
-    elseif arg[2] == 'sync' then
-        sync(msg)
-    else
-        utils.printJson({
-            error = "Method not found"
-        })
-    end
-end
+#!/bin/sh
+# --[[
```
G10h4ck, line 2: this line should now be blank

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
```diff
@@ -1,63 +1,43 @@
...
+# Shared State Async
+
+# Copyright (c) 2024 Javier Jorge <jjorge@inti.gob.ar>
+# Copyright (c) 2024 Instituto Nacional de Tecnologia Industrial
+# Copyright (C) 2024 Asociacion Civil Altermundi <info@altermundi.net>
+
+# SPDX-License-Identifier: AGPL-3.0-only
+# ]]
```
G10h4ck, line 10: this like show now go away

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
```diff
@@ -1,63 +1,43 @@
...
+sinc_args=""
+ 
+case "$1" in 
+    list) 
+        echo '{ "sync": { "data_type": "str", "peers_ip": "str" }, "get": { "data_type": "str" } }'
+    ;; 
+    call)
+        # source jshn shell library
+        . /usr/share/libubox/jshn.sh
+        read nmsg
```
G10h4ck, line 21: What this variable name means? Either I miss something or a more readable name should be used (always remember what will I understand when I'll reopen this file a couple years after?)

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
G10h4ck, line 1: Using shell instead of Lua resulted in a much nicer code as anticipated :-)

packages/shared-state/files/usr/libexec/rpcd/shared-state
```diff
@@ -1,45 +1,51 @@
...
 Shared State
 
 Copyright (c) 2023  Javier Jorge <jjorge@inti.gob.ar>
 Copyright (c) 2023  Instituto Nacional de Tecnología Industrial
 Copyright (C) 2023  Asociación Civil Altermundi <info@altermundi.net>
 
 This is free software, licensed under the GNU AFFERO GENERAL PUBLIC LICENSE Version 3
-]] --                                                                                 
+]]
+   --
```
G10h4ck, line 11: Shouldn't this line be blank?

G10h4ck (changes requested):
Just a few improvements and it is ready to merge

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
javierbrk: I will change it to match the name of the same variable in all the other rpcd scripts "msg"

packages/shared-state-async/files/usr/libexec/rpcd/shared-state-async
javierbrk: much better, I was worried about json parsing but some else was also worried before me.

packages/shared-state/files/usr/libexec/rpcd/shared-state
javierbrk: ouch garbage from lua code

javierbrk:
Shared State rpcd/ubus interface is used only by lime app so they go together. 
The last release of limeapp is not using this interface. @selankon can confirm. So this is not required in the candidate release.

G10h4ck (approved):

PR #1107, opened 2024-05-09
Move mesh wide status packages to shared-state-async

javierbrk:
This pull request moves two packages from normal shared state to new async implementation.


ilario:
Thanks!
Do you think we should include this in the release candidate?
Do you think we should do this to all packages so that we can drop the legacy shared-state removing it from the release candidate?

javierbrk:
no, not yet, these packages will be used by another version of lime-app

javierbrk:
closed and replaced with #1112

PR #1108, opened 2024-05-09
shared-state-async: add publisher call interface

javierbrk:
This pull request enables rpcd publisher call, so a publish followed by a sync deliver accurate and updated data. 

Node A should rpcd call to Node B publish and then 
Node A should call its own rpcd sync using its datatype and the IP of node B
Bat hosts publisher name has to be updated before closing this pull request


selankon:
Add publish_all rpcd. It depends on https://github.com/libremesh/lime-packages/pull/1104

selankon (approved):
Everything works fine!

PR #1131, opened 2024-09-20
Ensure that wan interface gets configured on DSA devices, where wan could be the name of the interface

ilario:
Initially proposed in this comment: https://github.com/libremesh/lime-packages/pull/1080#issuecomment-2351774330

On DSA devices, the "wan" interface could be named just "wan", without having any lower interface.
This ensures that it gets configured by the protocols even if it has no lower interface.

@pony1k @G10h4ck do you think that the same will be needed for LAN?

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -259,12 +259,14 @@ function network.scandevices()
 
 		--! With DSA, the WAN is named wan. Copying the code from the lan case.
 		if dev:match("^wan$") then
+			devices[dev] = {}
 			local lower_if = network._get_lower(dev)
 			if lower_if then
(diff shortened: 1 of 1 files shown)


pony1k:
I think the same should be done for interfaces named lan, lan1, lan2 etc.

ilario:
> I think the same should be done for interfaces named lan, lan1, lan2 etc.

I agree.
I opened a specific issue, please comment there your opinion. I am not sure of the right solution (and I am neither very familiar with the problem): https://github.com/libremesh/lime-packages/issues/1132

@G10h4ck please merge this unless you have a general idea on how to solve the issue.

G10h4ck (approved):

ilario:
Cherry-picked to 2024.1 branch for the next release https://github.com/libremesh/lime-packages/commits/2024.1/

Issue #1077, opened 2024-02-05
Change pirania's iptables rules to nftables


hiurequeiroz:
[Pirania's iptables rules](https://github.com/libremesh/lime-packages/blob/master/packages/pirania/files/usr/bin/captive-portal)  need to be converted into nftables to be compatible with the new libremesh release.

henmohr:
Opened a PR here: https://github.com/libremesh/lime-packages/pull/1133

PR #1133, opened 2024-09-26
Pirania New Release - Issue/1077

henmohr:
Hi! I was working during GSoC 2024 in the project [Project Libremesh Pirania, new release. ](https://summerofcode.withgoogle.com/programs/2024/projects/J7QlUBlx)

As discussed here https://github.com/libremesh/lime-packages/issues/1077

Status:

- its working as expected except that after successfull login (with option with_vouchers=0 or 1) the web page displays a bad gateway message.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,8 @@ build
 luacov.stats.out
 tools/ansible/files/generic-rootfs.tar.gz
 tools/ansible/files/ramfs.bzImage
-
+deploy-pirania.sh
+updatepkg.sh
 # Vim text editor swap files
 **/*~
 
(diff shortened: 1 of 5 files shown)


ilario (approved):

ilario:
I coarsely tested during the GSoC.
I think it should be merged as it improves the current code, even if the development is not 100% complete.
Do you agree @hiurequeiroz @henmohr @germanferrero @spiccinini ?
Then we can go on discussing on #1077

spiccinini:
+1 from me. It would be good if @henmohr can describe in more detail what is missing, what already tried to debugz so the next one working can start from better understanding.

Issue #430, opened 2018-11-20
In lime-defaults `ap_ssid` required even when `list_modes 'ap'` is removed


Juul:
In `config lime 'wifi'` even if we remove `list modes 'ap'` it still seems to still require `ap_ssid`.

Error:

```
WARNING: Attempt to access undeclared default for: wifi.ap_ssid
stack traceback:
/usr/lib/lua/lime/config.lua:24: in function 'get'
/usr/lib/lua/lime/utils.lua:61: in function 'network_id'
/usr/lib/lua/lime/utils.lua:81: in function 'applyNetTemplate10'
/usr/lib/lua/lime/network.lua:55: in function 'primary_address'
/usr/lib/lua/lime/proto/lan.lua:16: in function 'configure'
/usr/lib/lua/lime/network.lua:230: in function </usr/lib/lua/lime/network.lua:230>
[C]: in function 'xpcall'
/usr/lib/lua/lime/network.lua:230: in function </usr/lib/lua/lime/network.lua:210>
[C]: in function 'xpcall'
/usr/bin/lime-config:45: in function 'main'
/usr/bin/lime-config:65: in main chunk
[C]: ?
/usr/lib/lua/lime/utils.lua:63: attempt to concatenate local 'network_essid' (a nil value)
stack traceback:
/usr/lib/lua/lime/network.lua:231: in function </usr/lib/lua/lime/network.lua:231>
/usr/lib/lua/lime/utils.lua:63: in function 'network_id'
/usr/lib/lua/lime/utils.lua:81: in function 'applyNetTemplate10'
/usr/lib/lua/lime/network.lua:55: in function 'primary_address'
/usr/lib/lua/lime/proto/lan.lua:16: in function 'configure'
/usr/lib/lua/lime/network.lua:230: in function </usr/lib/lua/lime/network.lua:230>
[C]: in function 'xpcall'
/usr/lib/lua/lime/network.lua:230: in function </usr/lib/lua/lime/network.lua:210>
[C]: in function 'xpcall'
/usr/bin/lime-config:45: in function 'main'
/usr/bin/lime-config:65: in main chunk
[C]: ?
```

[Example of non-working config file](https://github.com/libremesh/network-profiles/blob/master/peoplesopen.net/common/etc/config/lime-defaults#L25)

Juul:
Tagging @nicopace

nicopace:
I would propose to change the name of ap_ssid to network_name... 
[...]

PR #1134, opened 2024-10-20
APuP support for LibreMesh

G10h4ck:
Initial support for APuP in libremesh
    
    APuP WiFi is now usable and testeable in LiMe
    Some part of the code have workaround to OpenWrt ubus and netifd current
    bugs/limitations and will be modified once upstream limitations are solved
    
    To use APuP just set 'apup' as LiMe wifi mode
    as an example using the followings commands
    
    ```
    uci del lime-community.wifi.modes
    uci add_list lime-community.wifi.modes=apup
    uci commit
    lime-config
    reboot
    ```

### DIFF
--- a/packages/lime-system/files/etc/config/lime-defaults
+++ b/packages/lime-system/files/etc/config/lime-defaults
@@ -58,6 +58,7 @@ config lime wifi
 	option apname_ssid 'LibreMesh.org/%H'
 	option adhoc_ssid 'LiMe'
 	option adhoc_bssid 'ca:fe:00:c0:ff:ee'
+	option apup_ssid 'LibreMesh.org'
 	option ieee80211s_mesh_fwding '0'
 	option ieee80211s_mesh_id 'LiMe'
 	option unstuck_interval '10'
--- a/packages/lime-system/files/usr/bin/limed
+++ b/packages/lime-system/files/usr/bin/limed
@@ -1,0 +1,87 @@
+#!/usr/bin/lua
+
+--! LibreMesh community mesh networks meta-firmware
+--!
+--! Copyright (C) 2024  Gioacchino Mazzurco <gio@polymathes.cc>
+--! Copyright (C) 2024  Asociación Civil Altermundi <info@altermundi.net>
+--!
+--! SPDX-License-Identifier: AGPL-3.0-only
+
+local config = require("lime.config")
+
+local limed = {};
+
+limed.__PRIVATE_APUP_ENABLED = nil
+
+function limed.APUP_ENABLED()
+	if limed.__PRIVATE_APUP_ENABLED ~= nil then
+		return limed.__PRIVATE_APUP_ENABLED
+	end
+
+	function check_apup_mode_cb(section)
+		if(section['modes'] == null) then return end
+
+		for _,mode in pairs(section['modes']) do
+			if(mode == 'apup') then
+				limed.__PRIVATE_APUP_ENABLED = true;
+				return limed.__PRIVATE_APUP_ENABLED;
+			end
+		end
+	end
+
+	config.foreach(nil, check_apup_mode_cb)
+	return limed.__PRIVATE_APUP_ENABLED
+end
+
+-- No need to run if APuP is not enabled in configuration
+if not limed.APUP_ENABLED() then os.exit(0) end;
+
+local uloop = require("uloop");
+local network = require("lime.network")
+
+local function dumptable(table, nesting)
+  local nesting = nesting or 1
+  if type(table) ~= "table" then
+    print("dumptable: first argument is expected to be a table but you passed a", type(table), table)
+  else
+    if next(table) == nil then
+      print(table, "empty")
+    else
+      for k,v in pairs(table) do
+        print(string.rep('\t', nesting), k, ' = ', v)
+        if type(v) == 'table' then dumptable(v, nesting+1) end
+      end
+    end
+  end
+end
+
+local libubus = require("ubus")
+local ubus = libubus.connect()
+
+local peerSubscriber = {
+	notify = function(nData, nType)
+		if nType ~= "apup-newpeer" then return end
+
+		print("peerSubscriber:", nType, nData.ifname)
+		network.createStatic(nData.ifname)
+		network.runProtocols(nData.ifname)
+	end
+}
+
+local apupSubscriber = {
+	notify = function(nData, nType)
+		if nType ~= "bss.add" then return end
+
+		local apupDev = string.match(nData["name"], "wlan%d+%-apup");
+		if not apupDev then return end
+
+		local evPath = "hostapd." .. apupDev
+
+		print("Subscribing:", evPath)
+		ubus:subscribe(evPath, peerSubscriber)
+	end
+}
+
+uloop.init();
+ubus:subscribe("hostapd", apupSubscriber)
+uloop.run();
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -2,8 +2,8 @@
 
 --! LibreMesh community mesh networks meta-firmware
 --!
---! Copyright (C) 2013-2023  Gioacchino Mazzurco <gio@eigenlab.org>
---! Copyright (C) 2023  Asociación Civil Altermundi <info@altermundi.net>
+--! Copyright (C) 2013-2024  Gioacchino Mazzurco <gio@polymathes.cc>
+--! Copyright (C) 2023-2024  Asociación Civil Altermundi <info@altermundi.net>
 --!
 --! SPDX-License-Identifier: AGPL-3.0-only
 
@@ -15,12 +15,20 @@ local fs = require("nixio.fs")
 local config = require("lime.config")
 local utils = require("lime.utils")
 
-network.limeIfNamePrefix="lm_net_"
-network.protoParamsSeparator=":"
-network.protoVlanSeparator="_"
+
+function network.PROTO_PARAM_SEPARATOR() return ":" end
+function network.PROTO_VLAN_SEPARATOR() return "_" end
+function network.LIME_UCI_IFNAME_PREFIX() return "lm_net_" end
+
 
 network.MTU_ETH = 1500
 network.MTU_ETH_WITH_VLAN = network.MTU_ETH - 4
+
+-- Deprecated use corresponding functions instead
+network.protoParamsSeparator=":"
+network.protoVlanSeparator="_"
+network.limeIfNamePrefix="lm_net_"
+
 
 function network.get_mac(ifname)
 	local _, macaddr = next(network.get_own_macs(ifname))
@@ -268,7 +276,7 @@ function network.scandevices()
 			end
 		end
 
-		if dev:match("^wlan%d+"..wireless.wifiModeSeparator.."%w+$") then
+		if dev:match("^wlan%d+"..wireless.WIFI_MODE_SEPARATOR().."%w+$") then
 			devices[dev] = {}
 			utils.log( "network.scandevices.dev_parser found WiFi device %s",
 			           dev )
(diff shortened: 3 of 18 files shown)


packages/lime-system/files/usr/lib/lua/lime/wireless.lua
```diff
@@ -9,7 +9,10 @@ local iwinfo = require("iwinfo")
 wireless = {}
 
 wireless.limeIfNamePrefix="lm_"
-wireless.wifiModeSeparator="-"
```
ilario, line 12: Maybe we could leave this line for a while, as the similar ones in network.lua?

ilario: Or just update this line to use the function: https://github.com/libremesh/lime-packages/blob/ca6328347b97728b796fa28d243e02f95f7c6616/packages/lime-system/files/usr/lib/lua/lime/network.lua#L263

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -260,7 +268,41 @@ function network.scandevices(specificIfaces)
 			--! pluggable ethernet dongles.
 			utils.log( "network.scandevices.dev_parser found plain Ethernet " ..
 			           "device %s", dev )
-		elseif dev:match("^wlan%d+"..wireless.wifiModeSeparator.."%w+$") then
+		end
+
+		if dev:match("^eth%d+%.%d+$") then
```
ilario, line 273: This code block was removed in #1154. Maybe it should not be here?

packages/lime-system/files/usr/bin/limed
```diff
@@ -0,0 +1,106 @@
...
+	return limed.__PRIVATE_APUP_ENABLED
+end
+
+-- No need to run if APuP is not enabled in configuration
+if not limed.APUP_ENABLED() then os.exit(0) end;
+
+local uloop = require("uloop");
+local network = require("lime.network")
+
+local function dumptable(table, nesting)
```
ilario, line 43: This is a duplicate of the function you added in utils.dumptable, right?

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -538,4 +580,94 @@ function network.createMacvlanIface(baseIfname, linuxName, argsDev, argsIf)
...
+-- TODO: probably some code between here and configure might be deduplicaded
+function network.runProtocols(linuxBaseIfname)
+	utils.log("lime.network.runProtocols(%s, ...)", linuxBaseIfname)
+	local protoConfs = config.get("network", "protocols")
+	for _,protoConf in pairs(protoConfs) do
+		local args = utils.split(protoConf, network.PROTO_PARAM_SEPARATOR())
+		local protoModule = "lime.proto."..args[1]
+		if utils.isModuleAvailable(protoModule) then
+			local proto = require(protoModule)
+			xpcall(function() proto.runOnDevice(linuxBaseIfname, args) end,
```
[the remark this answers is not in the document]
ilario: Ok, maybe I understood: `configure` writes the configuration for the interface to get up at the next reboot or restart of the service while `runOnDevice` makes the thing happen right now in a volatile way. I think this should be explained more extensively in the comments for the future readers.

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -538,4 +580,94 @@ function network.createMacvlanIface(baseIfname, linuxName, argsDev, argsIf)
 	return owrtInterfaceName, linuxName, owrtDeviceName
 end
 
+--! Create a static interface at runtime via ubus
+function network.createStatic(linuxBaseIfname)
```
ilario, line 584: Could you add as a comment in the code the difference between `network.createStatic` and `network.createStaticIface`?

From what I can understand, the difference is that createStaticIface creates the configuration file (the actual interface will appear upon reboot or restart of the networking service), with a custom netmask, gateway and IPv4 or IPv6. While this creates the interface in the running system (not in the configuration file) with fixed netmask, no gateway and the primary IPv4. Right? If this is right, I think that it should have a more informative name than just `createStatic`.

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -538,4 +580,94 @@ function network.createMacvlanIface(baseIfname, linuxName, argsDev, argsIf)
...
+	local ifaceConf = {
+		name    = ubusIfaceName,
+		proto   = "static",
+		auto    = "1",
+		ifname  = linuxBaseIfname,
+		ipaddr  = ipv4:host():string(),
+		netmask = "255.255.255.255"
+	}
+
+	print('network', 'add_dynamic')
```
ilario, line 597: Are these two lines needed? They seem useful only during testing.

packages/lime-system/files/usr/lib/lua/lime/network.lua
```diff
@@ -538,4 +580,94 @@ function network.createMacvlanIface(baseIfname, linuxName, argsDev, argsIf)
...
+--! ATM work around the problem configuring IP addresses via ip command
+
+	utils.unsafe_shell("ip link set up dev "..ifaceConf.ifname)
+	utils.unsafe_shell("ip address add "..ifaceConf.ipaddr.."/32 dev "..ifaceConf.ifname)
+
+	return ifaceConf.name
+end
+
+--! Create a vlan at runtime via ubus
+function network.createVlan(linuxBaseIfname, vid, vlanProtocol)
```
ilario, line 635: Same comment as above: it would be convenient to rename this function so that the difference with `createVlanIface` is more clear.

ilario:
A few comments:

lime-curtigghio only has the `test.sh` and the readme files, right?
In this case, I would not place it in the `packages` folder, as it does not have a Makefile.
The `test.sh` file could be renamed and go in the `tests` folder.
The readme could be renamed (`lime-curtigghio` is not clear to me, does it mean anything?) and go either to the website libremesh.org as a new page or in the root of this repository.

In the readme there is no clear explanation about how to compile an image including your modifications of hostapd, for taking advantage of all the new code you propose here.

When a peer gets disconnected and connects back after a while (e.g. for bad signal or the remote device rebooted), the same APuP interface gets used or a new one gets created? In the second case, we would accumulate many interfaces. I could not spot anything that removes created interfaces, but this could be ok as the user can wipe them rebooting the device.

ilario:
> In the readme there is no clear explanation about how to compile an image including your modifications of hostapd, for taking advantage of all the new code you propose here.

Just realized that your edits have already been upstreamed :D amazing!
https://github.com/openwrt/openwrt/commit/e80520197c9ca7bced50d3605d6baba6dead6e35

packages/lime-system/files/etc/config/lime-defaults
```diff
@@ -59,6 +59,7 @@ config lime wifi
 	option apname_ssid 'LibreMesh.org/%H'
 	option adhoc_ssid 'LiMe'
 	option adhoc_bssid 'ca:fe:00:c0:ff:ee'
+	option apup_ssid 'LibreMesh.org'
```
ilario, line 62: The resulting AP interface(s) will simply be named `LibreMesh.org`? Will there be one new of such interfaces for each peer? Will the normal clients see a long list of identical ESSIDs? If they try to connect to it, will they succeed? In order for APuP to work, the `apup_ssid` and the `ap_ssid` have to be identical? Or should this string be identical just between the APuP peers?

Also, can you add this line to `packages/lime-docs/files/www/lime-example.txt`?

packages/lime-system/files/usr/bin/limed
G10h4ck: ups, debugging copy paste leftover

packages/lime-system/files/etc/config/lime-defaults
G10h4ck: No need for ap mode anymore, APuP create just one SSID both for clients and peers, and everyone will see just one SSID not many

packages/lime-system/files/usr/lib/lua/lime/network.lua
G10h4ck: yep

packages/lime-system/files/usr/lib/lua/lime/network.lua
G10h4ck: yeah this got pulled back in by rebase, this PR should not touch this stuff at all

G10h4ck:
I did respond to comments inline, but now cannot see my replies anymore, anyway the recent push should address them

packages/lime-docs/files/www/docs/lime-example.txt
```diff
@@ -82,7 +83,8 @@ config lime wifi					# Settings in this section applies to all radios.
 #	option ap_encryption 'psk2'
 	option apname_ssid 'LibreMesh.org/%H'		# SSID specific to each AP. A user can connect to the named AP to avoid roaming
 #	option apname_key 'SomeWPA2PskKey'
-#	option apname_encryption 'psk2'
+#	option apname_encryption 'psk2'             # set here your APuP based network name
+	option apup_ssid 'LibreMesh.org'
```
ilario, line 87: the comment is on the `apname_encryption` line XD

packages/lime-system/files/etc/config/lime-defaults
ilario: Wow ok, this is big. Nice!
In this case, we have to redefine `ap_ssid`.
I would propose this:
* remove both `apup_ssid` and `ap_ssid` (anyway ap_ssid was not a good description of what this variable was for, see https://github.com/libremesh/lime-packages/issues/430#issuecomment-442640124)
* the AP interface SSID, both for normal AP and for APuP, gets taken from a new `ssid` variable.
* the %N1 things (e.g. Batman-adv VLAN ID) gets set in a new variable, maybe `broadcast_id`, instead than being calculated from another setting.

So we would move from:

```
config lime network
[...]
	option main_ipv4_address '10.%N1.0.0/16'
[...]
	option main_ipv6_address 'fd%N1:%N2%N3:%N4%N5::/64'
[...]
	list protocols batadv:%N1
[...]
	option anygw_mac "aa:aa:aa:%N1:%N2:aa"
[...]

config lime wifi
[...]
	option ap_ssid 'LibreMesh.org'
[...]
	option apup_ssid 'LibreMesh.org'
```

To:

```
config lime network
	option broadcast_id 13
[...]
	option main_ipv4_address '10.%D1.0.0/16'
[...]
	option main_ipv6_address 'fd%D1:%D2%D3:%D4%D5::/64'
[...]
	list protocols batadv:%D1
[...]
	option anygw_mac "aa:aa:aa:%D1:%D2:aa"
[...]

config lime wifi
[...]
	option ssid 'LibreMesh.org'
[...]
```

Where `%D1` is `broadcast_id % 256`, `%D2` is the next 8 bits etc etc (so by default they would be all zeros except %D1 which is 13).

G10h4ck: I am not against changing this, as I agree ap_ssid is used for too many things that should not depends on it. But that is a major change, so let's keep it for another PR and merge this one AS-IS.

ilario: Ok, makes sense. 
Could you comment on the first and last paragraph of this comment? https://github.com/libremesh/lime-packages/pull/1134#issuecomment-2674143078

Then in the next weeks I hope I will manage to test it on some hardware... Should this work with OpenWrt 24.10?

ilario:
> lime-curtigghio only has the `test.sh` and the readme files, right? In this case, I would not place it in the `packages` folder, as it does not have a Makefile. The `test.sh` file could be renamed and go in the `tests` folder. The readme could be renamed (`lime-curtigghio` is not clear to me, does it mean anything?) and go either to the website libremesh.org as a new page or in the root of this repository.
> 
> When a peer gets disconnected and connects back after a while (e.g. for bad signal or the remote device rebooted), the same APuP interface gets used or a new one gets created? In the second case, we would accumulate many interfaces. I could not spot anything that removes created interfaces, but this could be ok as the user can wipe them rebooting the device.

Also, could you check what happened that caused the unit testing to start failing at this PR? It seems that something got broken, and not for external reasons (I tried to run the tests on another branch and they still work: https://github.com/ilario/lime-packages/actions)

G10h4ck:
As discussed in LibreMesh community meeting I have double checked and APuP is included in OpenWrt 24.10 stable release as you can see here:
https://github.com/openwrt/openwrt/blob/openwrt-24.10/package/network/services/hostapd/patches/780-Implement-APuP-Access-Point-Micro-Peering.patch

packages/lime-system/files/etc/config/lime-defaults
javierbrk: yes this works with OpenWrt 24.10

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
```diff
@@ -82,6 +85,20 @@ function wireless.mesh_ifaces()
 				table.insert(ifaces, entry.ifname)
 			end
 		end)
+	--add apup interfaces 
+	--this are not listed in uci 
+	local shell_output = utils.unsafe_shell("ls /sys/class/net/ -R")
```
javierbrk, line 90: this line made the hole testing framework to explode, but it seems that the problem was in lime location tests. I improved the stub handling in this file and the test framework started working again.

javierbrk (approved):
Tests are now working, and AP-UP is fully integrated into OpenWrt 24.10. I approve these changes so others can start using AP-UP

PR #1141, opened 2024-11-19
fixed typo ubus-lime-groundrouting in lime-app's Makefile

ilario:
That typo was introduced in January 2019 with this commit https://github.com/libremesh/lime-packages/commit/8fe8d05b3ad126d78f5615d3c8abf74c3e446949 part of https://github.com/libremesh/lime-packages/pull/453

It causes this warning and error messages when building LibreMesh using the [Buildroot method](https://libremesh.org/development.html#compiling_libremesh_from_source_code):

```
WARNING: Makefile 'package/feeds/libremesh/lime-app/Makefile' has a dependency on 'ubus-lime-grondrouting', which does not exist
 * pkg_hash_check_unresolved: cannot find dependency ubus-lime-grondrouting for lime-app
```

### DIFF
--- a/packages/lime-app/Makefile
+++ b/packages/lime-app/Makefile
@@ -22,7 +22,7 @@ define Package/$(PKG_NAME)
 	URL:=http://github.com/libremesh/lime-app
 	DEPENDS:=+rpcd +uhttpd +uhttpd-mod-ubus +uhttpd-mod-lua \
 		+ubus-lime-location +ubus-lime-metrics +ubus-lime-utils \
-		+rpcd-mod-iwinfo +ubus-lime-grondrouting
+		+rpcd-mod-iwinfo +ubus-lime-groundrouting
 	PKGARCH:=all
 endef
 


ilario:
This error got revealed when testing #1139.

Using this fix, I also got rid of these other errors:

```
 * pkg_hash_fetch_best_installation_candidate: Packages for lime-app found, but incompatible with the architectures configured
 * opkg_install_cmd: Cannot install package lime-app.
```

Before the adoption of libremesh.mk in #1139 (see also #766) this is not failing as the same typo is present also in the Makefile (since https://github.com/libremesh/lime-packages/commit/1b2d501f5f632d63fa6bf672cbc740a00cf20789). After #1139 the package name is taken from the directory name, where the typo is not present.

a-gave:
Thanks! I seem to have encountered this testing 1139 but skipped because building without lime-app.

a-gave (approved):

Issue #492, opened 2019-03-25
shared-state package requires documentation


natzlob:
Hi

I would like to use the shared-state package for sending channel (and other parameter) change information around a bmx6 mesh but I am not clear how to install or use it. Do I need to install all the packages that contain the words shared-state, or just shared-state-bat_host, and then how do I create my own json to send around? 

Please create a README to explain how to use the package.

Thanks

nicopace:
it seems there is a commandline tool called shared-state
this one: https://github.com/libremesh/lime-packages/blob/master/packages/shared-state/files/usr/bin/shared-state
ou call it with the data you want to share with others:
https://github.com/libremesh/lime-packages/blob/master/packages/shared-state-bat_hosts/files/usr/bin/shared-state-publish_bat_hosts#L38
`io.popen("shared-state insert bat-hosts", "w"):write(JSON.stringify(batHostTable))`
and you register for this changes by adding a symbolic link to the command that needs to be called with the JSON that was received:
https://github.com/libremesh/lime-packages/tree/master/packages/shared-state-bat_hosts/files/etc/shared-state/hooks/bat-hosts
and you do what you want with that JSON, like in here:
https://github.com/libremesh/lime-packages/blob/master/packages/shared-state-bat_hosts/files/usr/bin/shared-state-generate_bat_hosts

not documentation per se, but hope it can be used as the beginning of one for you, @G10h4ck, @ilario or anyone interested

ilario:
@nicopace I never tried shared-state, the only documentation line I found was [this one](https://github.com/libremesh/lime-packages/blob/master/packages/shared-state/Makefile#L27), thanks.

G10h4ck:
shared-state distribute data in same broadcast domain. it is not suitable for pure bmx6 mesh, I would go for bmx6-sms in that case

nicopace:
@G10h4ck i thought it was using link local, not broadcast, but now I see in the code here https://github.com/libremesh/lime-packages/blob/master/packages/shared-state/files/usr/bin/shared-state-get_candidates_neigh#L42 that you are filtering the neighbours of only the same broadcast domain.
It could be possible that if it detects that it is not in a batman setup, for it to check neighbours over bmx and not linklocal, so it could be compatible with this setup.
Of course, PR welcomed :)

ilario:
5 months have passed, has some documentation been written?
I think I managed to use it in a basic way, but it was not easy: documentation is really needed!!!

[...]

PR #1142, opened 2024-11-22
New full documentation for shared-state

javierbrk:
This pull request is to address issue #492 and also to help development around shared state. 
Please feel free to make suggestions and even suggest full rewrite of the doc. 
It is a really interesting toppic.


ilario (approved):
Amazing!!

PR #1143, opened 2024-11-23
Fixes on ubus-lime-utils

### LINKED ISSUE #956: ubus-lime-utils uses watchping but does not depend on it and should check for client-wwan before running
As commented here https://github.com/libremesh/lime-packages/issues/927#issuecomment-1331084524 the ubus-lime-utils package uses watchping here:
https://github.com/libremesh/lime-packages/blob/400ac95448a1f1d5a83496488e40c302cde766f0/packages/ubus-lime-utils/files/etc/udhcpc.user#L11

but it does not include it in its dependencies.
This results in the following error:
```
/usr/share/udhcpc/default.script: /etc/udhcpc.user.d/50-client-wwan-watchping: line 11: /etc/init.d/watchping: not found
```

Additionally, if `udhcpc` is run manually, it also complains about the nonexistence of `client-wwan` interface:
```
ip: can't find device 'client-wwan'
```
so that a check should be included for the existence of the interface before proceeding.

ilario:
The first and the third commits fix #956 

I considered that the usage of watchping is not a real dependency of this package (this package does things even if watchping is not doing its things). So, rather than adding the dependency, I added a check for the existence of watchping (first commit).

The second commit is fixing a bug undetected since its introduction in https://github.com/libremesh/lime-packages/commit/400ac95448a1f1d5a83496488e40c302cde766f0 within #890

The third commit is checking if the `client-wwan` interface exists before running ip on it.

This could also be merged in the 2024.1 branch for the upcoming release.

### DIFF
--- a/packages/ubus-lime-utils/files/etc/udhcpc.user.d/50-client-wwan-watchping
+++ b/packages/ubus-lime-utils/files/etc/udhcpc.user.d/50-client-wwan-watchping
@@ -8,7 +8,7 @@ setup_hotspot_watchping() {
         uci set system.hotspot_watchping.pinghosts=$gw
         uci set system.hotspot_watchping.pinginterval=20s
         uci commit system
-        /etc/init.d/watchping restart
+        [ -e /etc/init.d/watchping ] && /etc/init.d/watchping restart
     fi
 }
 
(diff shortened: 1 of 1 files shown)


G10h4ck (approved):

ilario:
Picked in branch 2024.1

Issue #1023, opened 2023-05-02
Include auto-defer mechanism in deferable-reboot


ilario:
There is an auto-defer mechanism in LibreRouterOS here:

https://gitlab.com/librerouter/librerouteros/-/blob/librerouter-1.5/files/etc/uci-defaults/90-cron-defer-reboot

This should be included in the deferable-reboot package, and the pinged IP should be moved to the lime-defaults config file.

ilario:
Without this, and keeping the default configuration, deferable-reboot reboots the router every 27 hours.
https://github.com/libremesh/lime-packages/blob/fee2efe53387715d5dee70818678f0af50ac7753/packages/deferable-reboot/files/usr/lib/lua/deferable_reboot.lua#L6

ilario:
Fixed in #1147

PR #1147, opened 2024-11-25
Automatically postpone automated reboots by deferable-reboot

ilario:
As explained in #1023, deferable-reboot was developed in two pieces, the main one on this repository and a very useful piece in the LibreRouterOS repository.

Without the second piece, the routers would just reboot every 27 hours, annoying.
With the second piece, the routers are checking the internet connectivity every 3 minutes (by means of a ping to a known IP on the internet) and the reboot is deferred by 13 minutes if the ping is successful.

With this PR, the second piece is also included in the deferable-reboot package in this repository.

The IP to be pinged has been de-hardcoded and moved to the LibreMesh configuration files in the system section.

The chosen IP is one of the ones being used as DNS:
https://github.com/libremesh/lime-packages/blob/6642f373908d67c9cfc9d8f360843e9c528a8b94/packages/lime-docs/files/www/docs/lime-example.txt#L48-L49

I am testing in these days, please do not merge yet.


ilario:
Tested and it works.
@germanferrero once this gets merged, you should consider removing the file in the LibreRouterOS repository https://gitlab.com/librerouter/librerouteros/-/blob/librerouter-1.5/files/etc/uci-defaults/90-cron-defer-reboot

ilario:
Just increased a bit the deferring time (from 13 min to 30 min) in order to be sure that the network had time to re-establish the connection after a failure.
Also, added a random delay in the execution of the ping for not bombarding too hard the 4.2.2.2 with pings (does it ban IPs pinging too much?).

packages/deferable-reboot/files/etc/uci-defaults/90-cron-defer-reboot
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+unique_append()
+{
+	grep -qF "$1" "$2" || echo "$1" >> "$2"
+}
+
+unique_append \
+	"*/3 * * * * ((sleep $((RANDOM % 120)); if ping -c 10 -W 10 \$(uci -q get lime-autogen.system.deferable_reboot_ping_target) &> /dev/null; then awk '{print \$1 + 1800}' /proc/uptime > /tmp/deferable-reboot.defer ; fi )&) "\
```
G10h4ck, line 9: Seems a bit ugly and breakage prone to add a command so complex in the crontab, do you mind to create a script that does that, and call it from the crontab ? Ultimately I could even merge it like this but I do prefer if a more elegant solution is proposed.

G10h4ck (changes requested):
see inline comments

packages/deferable-reboot/files/etc/uci-defaults/90-cron-defer-reboot
ilario: Ok, thanks for the suggestion. Check it out now

ilario:
I took advantage for renaming the package fixing a typo in the name of the package itself.
I think that we should fix this before including this package in a LibreMesh release.

Also, I think that this package should not be in LibreMesh repositories. Instead it should be in OpenWrt ones, and I think that the typo should be fixed before pushing it to OpenWrt (unless it was intentional?).

G10h4ck (approved):

ilario:
Picked in branch 2024.1

PR #1149, opened 2024-11-26
ubus-lime-utils add dependency from ip package

### LINKED ISSUE #1137: ubus call lime-utils get_node_status problem in some dsa routers
a call to 
`ubus call lime-utils get_node_status`

wont always work

```
echo "" | /usr/libexec/rpcd/lime-utils call get_node_statusBusyBox v1.36.1 (2024-06-04 19:09:48 UTC) multi-call binary.

Usage: ip [OPTIONS] address|route|link|neigh|rule [ARGS]

OPTIONS := -f[amily] inet|inet6|link | -o[neline]

ip addr add|del IFADDR dev IFACE | show|flush [dev IFACE] [to PREFIX]
ip route list|flush|add|del|change|append|replace|test ROUTE
ip link set IFACE [up|down] [arp on|off] [multicast on|off]
 [promisc on|off] [mtu NUM] [name NAME] [qlen NUM] [address MAC]
 [master IFACE | nomaster] [netns PID]
ip neigh show|flush [to PREFIX] [dev DEV] [nud STATE]
ip rule [list] | add|del SELECTOR ACTION
**lua: /usr/lib/lua/lime/node_status.lua:141: attempt to index local 'dsa_json' (a nil value)**
```
**lua: /usr/lib/lua/lime/node_status.lua:141: attempt to index local 'dsa_json' (a nil value)**

the problem seems to be related to the use of unsupported options in the ip command on some specific ip compilations. The command that the scrip is trying to run is 

```bash

ip -j -p link show 
``` 
and some routers that use busybox don't have the -j option enabled.

there must be a way to pragmatically include a "full" ip package or change the code to use the simple ip and some tools to parse the output.

ilario:
Fix #1137 adding the dependency from the ip package. An alternative solution would be to parse the output of the ip command shipped with Busybox.

### DIFF
--- a/packages/ubus-lime-utils/Makefile
+++ b/packages/ubus-lime-utils/Makefile
@@ -12,9 +12,9 @@ define Package/$(PKG_NAME)
   CATEGORY:=Ubus
   MAINTAINER:=Santiago Piccinini <spiccinini@altermundi.net>
   SUBMENU:=3. Applications
-  TITLE:=LIbremesh ubus utils module
+  TITLE:=LibreMesh ubus utils module
   DEPENDS:= +lua +libubox-lua +libubus-lua +libuci +lime-system +libiwinfo-lua +cgi-io +rpcd-mod-file \
-	    +luci-lib-jsonc
+	    +luci-lib-jsonc +ip
 
   PKGARCH:=all
 endef


javierbrk:
Hi Ilario, even this pull fixes the problem, as you suggest it is not optimal. I've just created new pull to parse the output of the ip command shipped with Busybox.  https://github.com/libremesh/lime-packages/pull/1159 
Please let me know what do you think of it.

G10h4ck:
superseeded by https://github.com/libremesh/lime-packages/pull/1159

PR #1140, opened 2024-11-19
fix: ci build packages failed

a-gave:
this is intended to postpone #1139 while restoring the automatic build of packages

PR #1153, opened 2024-12-07
fix: typo of ubus-lime-groundrouting; fix multi-arch-build

a-gave:
fix typo from #1141 also in package ubus-lime-groundrouting 
copy fix from #1140 also in job multi-arch-build.yml


ilario (approved):
Thanks!! This should also be picked in the 2024.1 branch. I can do the cherry pick in these days.

Issue #1151, opened 2024-12-06
Observations when testing on PlasmaCloud PA1200


ilario:
We tested LibreMesh compiled on OpenWrt 22.03 during BattleMesh v15 in 2023 observing various unexpected behaviours, see https://github.com/libremesh/lime-packages/issues/1059
The network profiles that were finally used with these devices are the "outdoor" ones described here:
https://github.com/libremesh/network-profiles/tree/master/calafou

Now, I asked a PlasmaCloud PA1200 to @pedro-nonfree for testing again.

A few interesting observations, so far:

- a non-existing eth0 interface gets configured, while the existing ethernet1 interface stays down (in the lime-report below, you can see how ethernet1 is down even if it was physically connected to my laptop, so I had to connect via wifi) which is likely related to https://github.com/libremesh/lime-packages/issues/1132
- a lot of shared-state-async error messages are visible, see https://github.com/libremesh/lime-packages/issues/1150 @G10h4ck @javierbrk 
- mesh is working, even if I forgot to replace ath10k-ct with ath10k (CandelaTech drivers are recommended for AP but not for mesh).
- For some reason, iwinfo indicates that the mesh is WPA3 encrypted while it is not. But this happens on other devices when wpad-mesh-openssl is present (also on the YouHua WR1200JS).

Here you can see the output of lime-report when LibreMesh is installed (I included also the commit from these pull requests: #1139 #1143 #1145 #1147 #1148 #1149

[lime-report-20241206-PA1200-libremesh.txt](https://github.com/user-attachments/files/18042641/lime-report-20241206-PA1200-libremesh.txt)

And here the output of lime-report when installing a clean OpenWrt 23.05.5:

[lime-report-20241206-PA1200-clean_openwrt.txt](https://github.com/user-attachments/files/18042650/lime-report-20241206-PA1200-clean_openwrt.txt)

PR #1154, opened 2024-12-10
Scrape device names from board.json

### LINKED ISSUE #1121: Problems with ethernet connections between routers in default configuration
Continuing #1118 here because I'm not sure if trying to detect switches between routers is the right way to go forward.

Let's discuss the issue described by @ilario in this mail:
https://lists.autistici.org/message/20240714.140352.58fe57b2.en.html
 
In default configuration, ethernet interfaces are added to the br-lan bridge, while also being configured as batadv hard interface. In some setups, this leads to error messages appearing in the kernel log in a high rate and network instability.

It was suspected that the error appears iff there is a switch between the two routers. I tried to reproduce the issue with a dumb switch, but without success (everything working fine, no errors in kernel log), as described in this mail:
https://lists.autistici.org/message/20240726.150840.dcc0e028.en.html

I then tried to reprduce it by replacing the switch with an OpenWrt-router (without DSA), basicly acting as a managable switch, with no sucess either.

Then when I connected the two LibreMesh routers directly, suprisingly I could observe the issue. I could observe the error messages in the kernel logs and batadv didn't mesh over ethernet. On `mr70x-v1`, `batctl n` did not list the `fritz4040` as neighbour on the lan interfaces, also `batctl bbt` showed no routers in the backbone table on both routers. `batctl tcpdump lan1_29` could see batman OGMs appearing on the `lan1_29`, not sure why the interface was not showing up in the neighbour table. After a while, the wifi connection between my laptop and the routers became quite unusable. When I ran tcpdump on the mesh interfaces I found that there was a lot of broadcast and some frames were duplicated many times (I saw ICMPv6 messages with  same id and seq-no many times over long time periods. Plus, on my laptop, I saw the same echo request being received over and over at a high rate. So there was a loop and that clogged the wifi interface.

It is not the kind of loop I described in #1032 .

Later I booted the routers again, to further investigate the issue. Annoyingly, everything is working fine now. No kernel logs, meshing over ethernet works, no frames looping around. I'm not able to replicate the issue again. I also tried with resetting the configuration to firstboot state, but to no avail. So, unfortunatly, it is currently not possible for me to find out when excactly this happens and why.

I find it strange that batman is also configured on `eth0` on dsa enabled devices. I don't think we are supposed to use that directly. Next time someone observes this issue, maybe they could add
```
config net
	option linux_name 'eth0'
	list protocols 'manual'
```
to the `lime-node` file and see if it helps.

### LINKED ISSUE #1132: Consider new interface names
In the swconfig era (OpenWrt 19.07 and previous), interfaces had usually a name like eth0.2.
With DSA, interfaces can have funny names like [ethblue](https://github.com/openwrt/openwrt/blob/01996b785dcef61e6dadb50f61a13f905f3e497d/target/linux/ramips/dts/mt7621_gnubee_gb-pc2.dts#L148-L155), [game](
https://github.com/openwrt/openwrt/blob/eb13076e77c412cfe4dd6de48138507ea6f3c8b3/target/linux/mediatek/dts/mt7986a-acer-predator-w6.dts#L348-L351), lan1, ethernet1...

So we either find a general way to deal with arbitrary interface names, or we extract a list of the interface names from the targets we care about (for example, these: https://repo.libremesh.org/releases/2020.4-ow19/targets/).

The code that needs adaptation is the dev_parser in network.lua in lime-system:

https://github.com/libremesh/lime-packages/blob/d5befee61083bea0af7a29b6f0119f35dfdd7473/packages/lime-system/files/usr/lib/lua/lime/network.lua#L234-L271

pony1k:
- Scrape configurable network devices from /etc/board.json.
- Avoid configuration on DSA conduit devices.
- Allow device specific configuration for network devices with
  arbitrary names. This is more intuitive from a users perpective. Before, devices whose names don't match certain patterns were ignored. This could be helpful for network devices with unforseen names.
- Remove 'nobridge' tag, as this seems no longer be needed.
- Add optional parameter 'port' to utils.is_dsa().
- Add 'dsa' tag for DSA user ports. This could be used to fi_x #1121 like suggested by @ilario in [this comment](https://github.com/libremesh/lime-packages/issues/1121#issuecomment-2525078108).

fixes #1132

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -220,10 +220,16 @@ function network._get_lower(dev)
     return lower_if and lower_if:gsub("\n", "")
 end
 
-function network.scandevices()
+function network._is_dsa_conduit(dev)
+	return "reg" == fs.stat("/sys/class/net/" .. dev .. "/dsa/tagging", "type")
+end
+
+
+function network.scandevices(specificIfaces)
 	local devices = {}
-	local switch_vlan = {}
 	local wireless = require("lime.wireless")
+	local cpu_ports = {}
+	local board = utils.getBoardAsTable()
 
 	function dev_parser(dev)
 		if dev == nil then
@@ -231,50 +237,43 @@ function network.scandevices()
 			return
 		end
 
+		--! Avoid configuration on DSA conduit interfaces.
+		--! See also:
+		--! https://www.kernel.org/doc/html/latest/networking/dsa/dsa.html#common-pitfalls-using-dsa-setups
+		if network._is_dsa_conduit(dev) then
+			utils.log( "network.scandevices.dev_parser ignored DSA conduit " ..
+			           "device %s", dev )
+			return
+		end
+
+		--! Filter out ethernet ports connected to switch in a swconfig device.
+		for cpu_port,_ in pairs(cpu_ports) do
+			if cpu_port == dev then
+				utils.log( "network.scandevices.dev_parser ignored ethernet " ..
+				           "device %s connected to internal switch", dev )
+				return
+			end
+		end
+
 		if dev:match("^eth%d+$") then
-			devices[dev] = devices[dev] or {}
+			--! We only get here with devices not listed in board.json, e.g
+			--! pluggable ethernet dongles.
 			utils.log( "network.scandevices.dev_parser found plain Ethernet " ..
 			           "device %s", dev )
-		end
-
-		if dev:match("^eth%d+%.%d+$") then
-			local rawif = dev:match("^eth%d+")
-			devices[rawif] = { nobridge = true }
-			devices[dev] = {}
-			utils.log( "network.scandevices.dev_parser found vlan device %s " ..
-			           "and marking %s as nobridge", dev, rawif )
-		end
-
-		--! With DSA, the LAN ports are not anymore eth0.1 but lan1, lan2...
-		--! or just lan.
-		if dev:match("^lan%d*$") then
-			local lower_if = network._get_lower(dev)
-			if lower_if then
-				devices[lower_if] = { nobridge = true }
-				devices[dev] = {}
-				utils.log( "network.scandevices.dev_parser found LAN port %s " ..
-				           "and marking %s as nobridge", dev, lower_if )
-			end
-		end
-
-		--! With DSA, the WAN is named wan. Copying the code from the lan case.
-		if dev:match("^wan$") then
-			devices[dev] = {}
-			local lower_if = network._get_lower(dev)
-			if lower_if then
-				devices[lower_if] = { nobridge = true }
-				utils.log( "network.scandevices.dev_parser found WAN port %s " ..
-				           "and marking %s as nobridge", dev, lower_if )
-			else
-				utils.log( "network.scandevices.dev_parser found WAN port %s", dev)
-			end
-		end
-
-		if dev:match("^wlan%d+"..wireless.wifiModeSeparator.."%w+$") then
-			devices[dev] = {}
+		elseif dev:match("^wlan%d+"..wireless.wifiModeSeparator.."%w+$") then
 			utils.log( "network.scandevices.dev_parser found WiFi device %s",
 			           dev )
+		elseif specificIfaces[dev] then
+			utils.log( "network.scandevices.dev_parser found device %s that " ..
+			           "matches the config net section %s", dev,
+			           specificIfaces[dev][".name"])
+		else
+			return
 		end
+
+		local is_dsa = utils.is_dsa(dev)
+		devices[dev] = devices[dev] or {}
+		devices[dev]["dsa"] = is_dsa
 	end
 
 	function owrt_ifname_parser(section)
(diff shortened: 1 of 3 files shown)


pony1k:
I have tested the following:
- DSA ports with specific configuration.
- DSA ports without specific configuration.
- DSA conduit does not get configured.
- dummy ethernet port 'eth9' gets automatically configured.
- dummy network device called 'foobar' gets configured iff there is a specific `config net` `option linux_name 'foobar'` section for it.
- 'eth0' is ignored on a swconfig device.
- 'eth0.1' and 'eth0.2' are configured with and without specific configuration.

To create the dummy devices, I installed 'kmod-dummy' and then ran `ip link add foobar type dummy`.
It would be good to test this a little bit more. Maybe also on a PlasmaCloud PA1200, if someone has one available.

ilario:
This looks amazing :D

I flashed 3 routers with this and here you are the outputs of lime-report (beware that I included in the flashed images all the pull requests currently open):

A PlasmaCloud PA1200:
[lime-report-20241215-PA1200-libremesh.txt](https://github.com/user-attachments/files/18155076/lime-report-20241215-PA1200-libremesh.txt)
There is a 192.168.1.2 IPv4, but I vaguely recall I may have set it manually myself.
You can take as a reference the file here: https://github.com/libremesh/lime-packages/issues/1151#issue-2723748316

A TP-Link WDR3600:
[lime-report-20241215-WDR3600-libremesh.txt](https://github.com/user-attachments/files/18155078/lime-report-20241215-WDR3600-libremesh.txt)

A YouHua WR1200JS:
[lime-report-20241215-YouHua-libremesh.txt](https://github.com/user-attachments/files/18155080/lime-report-20241215-YouHua-libremesh.txt)

If you consider useful to have some references good for comparisons, I can flash images with and without this pull request, just let me know!

With this code, the ethernet1 and ethernet2 interfaces from the PlasmaCloud device are correctly configured! :D

One doubt:
In the TP-Link WDR3600, the `batctl if` output includes the ethernet interfaces, while in the YouHua WR1200JS it includes just the WAN port (additionally to the wifi interfaces that are always there) and on the PlasmaCloud PA1200 it does not include any cabled interface. I suppose this is because the YouHua WAN port was connected to the ISP router while the LAN ports were unused? And also on the PA1200, while I tested there was nothing connected via cable. I will have to check connecting cables.

ilario:
Adding the report from a Ubiquiti NanoStation LoCo M2 XM (with just 32 MB of RAM, so I had to create a very minimal image).
The device has only one ethernet interface, identified in OpenWrt as LAN.

[lime-report-20241216-NanoStation_Loco_M2_XM-libremesh_bis.txt](https://github.com/user-attachments/files/18157160/lime-report-20241216-NanoStation_Loco_M2_XM-libremesh_bis.txt)

pony1k:
> One doubt:
> In the TP-Link WDR3600, the `batctl if` output includes the ethernet interfaces, while in the YouHua WR1200JS it includes just the WAN port (additionally to the wifi interfaces that are always there) and on the PlasmaCloud PA1200 it does not include any cabled interface. I suppose this is because the YouHua WAN port was connected to the ISP router while the LAN ports were unused? And also on the PA1200, while I tested there was nothing connected via cable. I will have to check connecting cables.

Interfaces are added to batadv when they are up and removed when they are down. On WDR3600, eth0 is always up, because it's connected to the switch. So this is expected.

One thing that at first striked me as odd is that with PlasmaCloud PA1200, without patch it has
```
config interface 'wan'
	option device 'ethernet2'
	option proto 'none'
```
and with patch it has
```
config interface 'wan'
	option device 'ethernet2'
	option proto 'dhcp'
```
. But this is expected too, because
https://github.com/libremesh/lime-packages/blob/50a4a097ebcfaeb13f04d1a1a5df401ca7892919/packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua#L16-L24
is only run when there is an actual interface configured with lime-proto-wan (which does not happen without the patch), while
https://github.com/libremesh/lime-packages/blob/50a4a097ebcfaeb13f04d1a1a5df401ca7892919/packages/lime-system/files/usr/lib/lua/lime/network.lua#L174-L181
is always run.

All reports look good to me.

ilario:
Great!
I tried small network with some cabled connections and I can confirm that it works as it should :D

The network was like this:

ISP router --cable-- TP-Link WDR3600 --wifi-- (Ubiquiti NanoStation LoCo M2 XM --cable-- LAN3 YouHua WR1200JS LAN1 --cable-- LAN PlasmaCloud PA1200)

I did not configure the ethernet interfaces manually, so I would also expect some issues due to #1121.

The internet connection was not working due to a new issue #1157. No idea what caused this new issue.

The report from the Ubiquiti NanoStation LoCo M2 XM is incomplete as, in this network, it was frequently rebooting.

[lime-report-20241218-YouHua-libremesh.txt](https://github.com/user-attachments/files/18188985/lime-report-20241218-YouHua-libremesh.txt)

[lime-report-20241218-WDR3600-libremesh.txt](https://github.com/user-attachments/files/18188986/lime-report-20241218-WDR3600-libremesh.txt)

[lime-report-20241218-PA1200-libremesh.txt](https://github.com/user-attachments/files/18188987/lime-report-20241218-PA1200-libremesh.txt)

[lime-report-20241218-NanoStation_Loco_M2_XM-libremesh.txt](https://github.com/user-attachments/files/18189012/lime-report-20241218-NanoStation_Loco_M2_XM-libremesh.txt)

pony1k:
> The report from the Ubiquiti NanoStation LoCo M2 XM is incomplete as, in this network, it was frequently rebooting.

Why would it do that? Does it have something to do with this PR?

ilario:
> > The report from the Ubiquiti NanoStation LoCo M2 XM is incomplete as, in this network, it was frequently rebooting.
> 
> Why would it do that? Does it have something to do with this PR?

No, it is completely unrelated. I think it is going out of memory and killing something actually useful and then rebooting. Totally unrelated to this PR.

pony1k:
@G10h4ck Are you okay with this solution?

If this is going to be merged, I would like to add some test cases beforehand.

ilario (approved):
Tested and it works perfectly!
Specifically, the `ethernet1` and `ethernet2` interfaces of PlasmaCloud PA1200 have been configured correctly!
I documented a bit the testing here: https://github.com/libremesh/network-profiles/tree/master/ilario-testing

ilario:
> If this is going to be merged, I would like to add some test cases beforehand.

To have tests is always nice :D 
But I am merging this already, so that we can use this for the 2024.1 release :D
So, if you don't mind, I would leave the tests for another pull request.

PR #1159, opened 2024-12-23
Fix get node status to work with busybox ip command

javierbrk:
Fix https://github.com/libremesh/lime-packages/issues/1137 parsing the output of the "ip" command shipped with Busybox (default OpenWrt option)
With this fix, it is no longer necessary to include the ip-tiny package, which has a size of approximately 126 KB depending on the platform. On small hardware platforms where storage is limited, every byte counts.


G10h4ck:
Do I understand well, that with this PR it works both with busybox ip command and with ip-tiny ?

G10h4ck (approved):

ilario:
Picked in branch 2024.1

PR #1161, opened 2025-01-05
network: Do not configure protocols on DSA switch port network devices by default.

pony1k:
work around issue #1121

This adds an optional parameter 'port' to util.is_dsa(), to allow checking if a specific network device is a DSA switch port and prevents such network devices to be configured with protocols other than 'lime-proto-lan', unless there is a 'config net' section for that network device.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -397,10 +397,18 @@ function network.configure()
 
 		for _,protoParams in pairs(deviceProtos) do
 			local args = utils.split(protoParams, network.protoParamsSeparator)
-			if args[1] == "manual" then break end -- If manual is specified do not configure interface
-			local protoModule = "lime.proto."..args[1]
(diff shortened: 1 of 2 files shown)


pony1k:
@ilario Can this workaround go into the 2024.1release?

ilario:
Thanks Pony!!!
I will test it.
No idea if @a-gave already started compiling stuff...?

ilario:
@G10h4ck @digitigrafo can you also have a look/test? Do you think it is ok to delay the 2024.1 release for having the time to test this?

a-gave:
> Thanks Pony!!! I will test it. No idea if @a-gave already started compiling stuff...?

No, but I am running the preliminary steps/checks.
It looks good and I would have no problem including it if it works as expected.
I was waiting for the next 2024.1 or 2024.1-rc2 tag to do the final steps and produce the firmwares.

ilario (approved):
Tested and it works :D
I documented a bit the testing here: https://github.com/libremesh/network-profiles/tree/master/ilario-testing

PR #1164, opened 2025-04-02
WIP: Pirania rewrite

luandro:
This is a re-write of Pirania using nftables, still WIP.

Must haves:
- [x] Open captive portal with Internet
- [x] Block access to Internet
- [x] Permit access to local network
- [ ] Working whitelist of macs
- [ ] Working whitelist of ips
- [ ] Working blocklist of ips
- [ ] Working with Ubus
- [ ] Working with LimeApp

### DIFF
--- a/packages/pirania/Makefile
+++ b/packages/pirania/Makefile
@@ -13,12 +13,11 @@ define Package/$(PKG_NAME)
   CATEGORY:=Network
   MAINTAINER:=Asociación Civil AlterMundi <info@altermundi.net>
   TITLE:=Captive portal with vouchers.
-  DEPENDS:=+ip6tables-mod-nat +ipset +shared-state +shared-state-pirania \
+  DEPENDS:=+nftables +liblucihttp0 +uhttpd +uhttpd-mod-ubus +shared-state +shared-state-pirania \
 	+uhttpd-mod-lua +lime-system +luci-lib-jsonc \
 	+liblucihttp-lua +luci-lib-nixio +libubus-lua +libuci-lua
   PKGARCH:=all
 endef
-
 define Package/$(PKG_NAME)/description
 	Captive Portal for routers that want to share their Internet connection via vouchers.
 endef
--- a/packages/pirania/LEIAME.md
+++ b/packages/pirania/LEIAME.md
@@ -1,0 +1,210 @@
+![PIRANHA](https://i.imgur.com/kHWUNOu.png)
+
+## Solução de Voucher e Portal Cativo para redes comunitárias
+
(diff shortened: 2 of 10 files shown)


ilario:
Looks amazing :D
Compiling an image with this code and OpenWrt 24.10, I observed this warning message:
`WARNING: Makefile 'package/feeds/libremesh/pirania/Makefile' has a dependency on 'liblucihttp0', which does not exist`
Seems that from the buildroot point of view, that package is named `liblucihttp` instead of `liblucihttp0`.

ilario:
> Compiling an image with this code and OpenWrt 24.10, I observed this warning message: `WARNING: Makefile 'package/feeds/libremesh/pirania/Makefile' has a dependency on 'liblucihttp0', which does not exist` Seems that from the buildroot point of view, that package is named `liblucihttp` instead of `liblucihttp0`.

I looked into this: the package is named liblucihttp and is pulled as a dependency of liblucihttp-lua, but there is an `ABI_VERSION:=0` thing in the Makefile that causes the `0` to be appended to the package name.

So, that warning message seems to be erroneous and everything is actually ok. Is this correct @jow- ?

ilario:
> The source level dependency should be just `+liblucihttp` - buildroot will take care of turning it into `Depends: liblucihttp0` automatically in binary package metadata.

Thanks!!!! I did not see that there was also an explicit dependency from liblucihttp0 in the Makefile!

packages/pirania/Makefile
```diff
@@ -13,12 +13,11 @@ define Package/$(PKG_NAME)
   CATEGORY:=Network
   MAINTAINER:=Asociación Civil AlterMundi <info@altermundi.net>
   TITLE:=Captive portal with vouchers.
-  DEPENDS:=+ip6tables-mod-nat +ipset +shared-state +shared-state-pirania \
+  DEPENDS:=+nftables +liblucihttp0 +uhttpd +uhttpd-mod-ubus +shared-state +shared-state-pirania \
```
ilario, line 16: Can you rename the dependency `liblucihttp0` to `liblucihttp`? See the comment by jow- in the conversation here https://github.com/libremesh/lime-packages/pull/1164#issuecomment-2961499968 
You could even remove it, as it is pulled by liblucihttp-lua anyways.

luandro: Thanks, done!

Jadhav-Dhanashri:
Hello maintainers ,
 I'm new to freifunk.net open source contribution & this would be my first issue. I'd like to work on this if it's available & learn    from the process. Could you please assign it to me? 
Thank you !

PR #1170, opened 2025-04-29
Ethernet configuration using lime-app

javierbrk:
This pull request is the back-end of this [issue](https://github.com/libremesh/lime-app/issues/390), it enables ubus methods to get and set user defined Ethernet configurations.

It enables three modes, wan, lan (to connect clients), mesh (to connect other nodes) or default with removes the user defined configuration for this interface.  
```
ssh root@thisnode.info '/usr/libexec/rpcd/lime-eth-config call get_eth_config'

ssh root@thisnode.info 'echo "{\"device\":\"eth1.2\",\"role\":\"wan\"}" | /usr/libexec/rpcd/lime-eth-config call set_eth_config'

ssh root@thisnode.info 'echo "{\"device\":\"eth1.2\",\"role\":\"default\"}" | /usr/libexec/rpcd/lime-eth-config call set_eth_config'

ssh root@thisnode.info 'echo "{\"device\":\"eth0.1\",\"role\":\"lan\"}" | /usr/libexec/rpcd/lime-eth-config call set_eth_config'

ssh root@thisnode.info 'echo "{\"device\":\"eth1.2\",\"role\":\"mesh\"}" | /usr/libexec/rpcd/lime-eth-config call set_eth_config'

``

### CODE REVIEW

**Comment by ilario** [comment]
Nice!!
I still have to test this.
One comment: could you migrate the Makefile to use the "new" compact style? I mean, like this: https://github.com/libremesh/lime-packages/pull/1139/commits/86b80548c451ca878391094a4127c4a5a0f65462

**Comment by a-gave** [comment]
Thanks! 

I wonder if would be worth to use `lime-autogen` instead of `lime-node` when getting info
to respect the hierarchy of lime-[files]: 0- lime-autogen; 1- lime-node; 2- lime-MAC; 3- lime-community; 4- lime-defaults

`uci:foreach("lime-node", "net", function(entry)` [R20](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R20)
`local eth_role = uci:get("lime-node", tag_value, "eth_role")` [R62](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R62)

But I don't know if it is the case that lime-community will ever contain reasonable defaults to configure specific interfaces.
Or alternatively we could write somewhere in the doc that interface specific configurations should only be written in lime-node (and not lime-defaults nor lime-community or lime-MAC)

I wont' be sure how to handle these deletions (but probably using `config.delete` instead of `config:uci:delete("lime-node"` should work)
[R54](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R54) [R62](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R62)

**Comment by ilario** [comment]
Good comment @a-gave !
I think that some communities could insert the ethernet configurations in the network-profiles which could be specific for each router model, like here: 
https://github.com/libremesh/network-profiles/blob/master/calafou/indoor/root/etc/config/lime-community#L29-L34

Another doubt: @javierbrk is the frontend for this already in lime-app? It is perfectly ok if it is not, I ask just for knowing.

**Comment by javierbrk** [comment]
> Nice!! I still have to test this. One comment: could you migrate the Makefile to use the "new" compact style? I mean, like this: [86b8054](https://github.com/libremesh/lime-packages/commit/86b80548c451ca878391094a4127c4a5a0f65462)

Yess I completely forgot about it!

**Comment by javierbrk** [comment]
> Good comment @a-gave ! I think that some communities could insert the ethernet configurations in the network-profiles which could be specific for each router model, like here: https://github.com/libremesh/network-profiles/blob/master/calafou/indoor/root/etc/config/lime-community#L29-L34

community configs are expected to be identical in the whole community. For that use case, a user changing community settings we will have the mesh config feature. so you will be able to change the config of the whole network from one node. this needs to replicate changes in all the active nodes.

> 
> Another doubt: @javierbrk is the frontend for this already in lime-app? It is perfectly ok if it is not, I ask just for knowing.

yes there is a partly implemented front but it is not yet linked to this.

**Comment by javierbrk** [comment]
> Thanks!
> 
> I wonder if would be worth to use `lime-autogen` instead of `lime-node` when getting info to respect the hierarchy of lime-[files]: 0- lime-autogen; 1- lime-node; 2- lime-MAC; 3- lime-community; 4- lime-defaults
> 
> `uci:foreach("lime-node", "net", function(entry)` [R20](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R20) `local eth_role = uci:get("lime-node", tag_value, "eth_role")` [R62](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R62)
> 
> But I don't know if it is the case that lime-community will ever contain reasonable defaults to configure specific interfaces. Or alternatively we could write somewhere in the doc that interface specific configurations should only be written in lime-node (and not lime-defaults nor lime-community or lime-MAC)
> 
> I wont' be sure how to handle these deletions (but probably using `config.delete` instead of `config:uci:delete("lime-node"` should work) [R54](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R54) [R62](https://github.com/libremesh/lime-packages/pull/1170/files#diff-6ceba13e8a09608f55c92728b3dae06cc25a396bf69fd90bb65f77788eed63d0R62)

This feature implements a change that is of interest for this node alone. the typical use case is "the node is in the roof and the cable is in the wrong port." :) 

I'm not completely sure where should this be...probably since this is the NEW intended role of the port it should be top priority.

**Comment by ilario** [comment]
> community configs are expected to be identical in the whole community.

Why?
This would be the case for people using FirstBootWizard, but it does not work on single-band routers (https://github.com/libremesh/lime-packages/issues/661), so I suspect that many communities are not using it.

In my experience at BattleMesh v15 in Calafou, it was quite convenient to have a profile for each combination of hardware and role.
And seems that also other few communities are doing something like that [1](https://github.com/libremesh/network-profiles/tree/master/openNET.io), [2](https://github.com/libremesh/network-profiles/tree/master/mobilquattrosud.com), [3](https://github.com/libremesh/network-profiles/tree/master/jardibotanic).

> For that use case, a user changing community settings we will have the mesh config feature. so you will be able to change the config of the whole network from one node. this needs to replicate changes in all the active nodes.

What is that? Is that part of FBW?

> This feature implements a change that is of interest for this node alone. the typical use case is "the node is in the roof and the cable is in the wrong port." :)

In the Calafou setup, I used the network port configuration for setting one router's port as LAN and another router's as mesh, very useful.

> I'm not completely sure where should this be...probably since this is the NEW intended role of the port it should be top priority.

I am not sure if I get this right, but lime-autogen is the highest priority. One option is to write and delete the config lines to and from lime-node, but to read them from lime-autogen. The advantage of reading from lime-autogen is that it will work regardless if the setting was contained in lime-node or lime-community or lime-MACaddress.

**Review by G10h4ck — APPROVED** [review]

File: packages/lime-eth-config/files/usr/lib/lua/lime-eth-config.lua [hunk 1]
```diff
@@ -0,0 +1,117 @@
...
+        eht_config.delete_eth_config(tag_value)
+    else
+        uci:set("lime-node", tag_value, "net")
+        uci:set("lime-node", tag_value, "eth_role", role)
+        uci:set("lime-node", tag_value, "linux_name", device)
+
+        if role == "wan" then
+            uci:set("lime-node", tag_value, "protocols", {"wan", "dynamic"})
+        elseif role == "lan" then
+            uci:set("lime-node", tag_value, "protocols", {"lan"})
```
ilario, line 104: Sorry for not commenting before. I still have to find time for testing... Does the anygw IP work even without specifying anygw proto here?

pony1k: This should not affect anygw, because it configures a specific network interface and the function that is called when anygw is listed for a specific network interface doesn't do anything:
https://github.com/javierbrk/lime-packages/blob/65e46dd3d858a3aca5e491284a186a68f87585c4/packages/lime-proto-anygw/files/usr/lib/lua/lime/proto/anygw.lua#L131

packages/lime-eth-config/files/usr/lib/lua/lime-eth-config.lua
```diff
@@ -0,0 +1,117 @@
...
+        elseif role == "mesh" then
+            uci:set("lime-node", tag_value, "protocols", {"batadv:%N1", "babeld:17"})
```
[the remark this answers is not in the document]
pony1k: Yes, but mainly because no one yet has dared to change it to babeld (the network spreads across large distances with multiple rooftops). I don't think anyone should go to great lengths to keep compatibility to bmx7.

packages/lime-eth-config/files/usr/lib/lua/lime-eth-config.lua
ilario: Well observed, thanks! :)

PR #1171, opened 2025-05-18
Update testing.md

VGDSpehar:
Update TESTING.md in order to provide more up-to-date firmware links for testing in QEMU. 

Also clarify the usage of qemu-dev-stop script, and a possible issue encountered (netcat with -N parameter support).

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -4,7 +4,7 @@ LibreMesh has unit tests that help us add new features while keeping maintenance
 
 We encourage contributors to write tests when adding new functionality and also while fixing regressions.
 
-LibreMesh unit testing is based in the powerful [busted](https://olivinelabs.com/busted/) library which has a very good documentation.
+LibreMesh unit testing is based in the powerful [busted](https://lunarmodules.github.io/busted/) library which has a very good documentation.
 
 Tests are run inside a x86_64 Docker image with some lua and openwrt libraries avaible.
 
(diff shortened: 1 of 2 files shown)


ilario:
Thanks!!
That `-N` comes from this commit 
https://github.com/libremesh/lime-packages/pull/702/commits/64635d7565fe2d130878ad060e4b418d5da15e76#diff-35f1bd2c18213b381415b3a8a58eb2d5ef992ecf0ac819ce9a1af10dbd0756b2 inside this pull request https://github.com/libremesh/lime-packages/pull/702

Could you check if it is actually needed?

@spiccinini do you think that the -N is still needed?

VGDSpehar:
From my testing it doesn't appear to be necessary per se, but removing it will cause some idling between the moment the command is sent and the moment the VM finally shutdown. 

What I think could be done is : 
- use  -w _timeout_ which is a portable option for netcat, so the user don't experience the idling
- simply put the netcat process in background

ilario:
Ah wow, ok!
I like the `-w` option more than the `-N` then!
If you also think it's better, could you add this change to this pull request?
The timeout can also be large, to kick in only if something goes wrong (e.g. if the closing takes 2 seconds on your laptop, the timeout could be 20 seconds).
Thanks!!!

VGDSpehar:
Just rebased on top of new master, and added the 'nc -w 1'  so netcat will timeout after 1 second instead of waiting for the command system_powerdown to complete (instead of nc -N), as well as removed the documentation regarding the possible error.

PR #1194, opened 2025-06-27
ci: tests: remove on pull_request

a-gave:
Following the discussion in [#1139](https://github.com/libremesh/lime-packages/pull/1139#issuecomment-3012596719)

Let's remove the on pull_request from tests
As a tool unittests can be runned locally via `./run_tests`
Unittests executed by the ci can eventually be manually triggered during a review.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,8 +3,6 @@ name: tests
 on:
   push:
     branches: [ '**' ]
-  pull_request:
-    branches: [ '**' ]
 
   # Allows you to run this workflow manually from the Actions tab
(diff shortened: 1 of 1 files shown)


a-gave:
> Unit testing is important. A good test coverage amount is something ideal. In our case, unit tests are not exhaustive, but they help us avoid regression bugs and reduce the manual testing. They’re like the seat belt of a car.
> 
[...]
> So my vote is to keep them before we merge it into master. Are there any other better options ?

Thanks I agree, let's solve the problem in other way, which could be by moving the CD repositories that build packages from those for CI

PR #1196, opened 2025-06-28
tests: allows tests to be run outside the user's HOME

### LINKED ISSUE #1195: Testing guide: specify where lime-packages folder should be
I tried to run `./run_tests` as mentioned on #1194 and in the TESTING.md file in this repository, but I got these errors:

```
# ./run_tests
0 successes / 0 failures / 3 errors / 0 pending : 0.000536 seconds
Error → Cannot find file or directory: packages/*/tests/test*.lua
Error → Cannot find file or directory: tests/test*.lua
Error → No test files found matching Lua pattern: _spec
```

This happened because I was running it as root and the lime-packages repository folder was not inside the home folder of the root user.
So we need to document how to run these tests without being root (maybe it is obvious how to do that, but I am not super familiar with docker sorry). And to specify that the lime-packages folder should be inside the user's home folder or to add PWD to the folders accessible to docker.

a-gave:
And document how to run tests without being root, fix #1195

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -17,8 +17,21 @@ Just execute `./run_tests`:
 ![run_tests](https://i.imgur.com/TBIE7Gp.png)
 
 
-This will build the testing Docker image automaticaly in the first run and then execute the tests and create the coverage report.
-Note: you must have Docker installed and running.
+This will build the testing Docker image automatically in the first run and then execute the tests and create the coverage report.    
+Note: you must have Docker installed and running.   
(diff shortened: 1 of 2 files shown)


ilario:
Amazing! Thanks! Could you specify that the user has to login/logout for applying the new group membership?
Otherwise, also `su - $USER` works for having a shell with the new group membership already in place.

a-gave:
> Amazing! Thanks! Could you specify that the user has to login/logout for applying the new group membership? Otherwise, also `su - $USER` works for having a shell with the new group membership already in place.

Yes, thanks. Done

PR #1215, opened 2025-08-31
ci: build packages for dev branches

a-gave:
feat: add automatic build of packages for branches starting with dev* to encourage and simplify testing.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on:
   release:
     types: [published]
   push:
-    branches: [master]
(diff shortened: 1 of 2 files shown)


ilario:
Some actions failed with `mips_4kec `and `riscv64`, maybe these are unused in OpenWrt 24.10 and the respective output directories are therefore not created by the action?

https://github.com/libremesh/lime-packages/actions/runs/17369620355/job/49302577678#step:4:7

https://github.com/libremesh/lime-packages/actions/runs/17369620355/job/49302577931#step:4:7

PR #1216, opened 2025-08-31
lime-owut: init the package

a-gave:
Add a script to setup a working owut.
Warning about packages missing can be just ignored for now.

### DIFF
--- a/packages/lime-owut/Makefile
+++ b/packages/lime-owut/Makefile
@@ -1,0 +1,41 @@
+include ../../libremesh.mk
+
+# owut is available since openwrt-24.10
+PKG_AVAILABLE:=1
+OW_BRANCH:=$(shell echo $$(grep -m1 -o "releases/.*" \
+	$(TOPDIR)/include/version.mk || echo "main") | \
+	cut -d"/" -f2 | cut -d"." -f1,2 )
+
+ifneq ($(OW_BRANCH),main)
+  ifeq ($(shell test $$(echo $(OW_BRANCH) | cut -d"." -f1) -lt 24; echo $$?),0)
+    PKG_AVAILABLE:=0
+  endif
+endif
+
+ifeq ($(PKG_AVAILABLE),1)
+
+	define Package/$(PKG_NAME)
+		TITLE:=$(PKG_NAME) allows to use owut OpenWrt Upgrade Tool in LibreMesh
+		CATEGORY:=LibreMesh
+		URL:=http://libremesh.org
+		DEPENDS:=+owut
+		PKGARCH:=all
+	endef
+
+	define Package/$(PKG_NAME)/postinst
+#!/bin/sh
+# Preserve the complete opkg status if CONFIG_CLEAN_IPKG=y
+grep -q CONFIG_CLEAN_IPKG=y $${TOPDIR}/.config || exit 0
+[ -f $${IPKG_INSTROOT}/usr/lib/opkg/status ] || exit 0
+
+cp $${IPKG_INSTROOT}/usr/lib/opkg/status $${IPKG_INSTROOT}/etc/opkg/
+cat << EOF > $${IPKG_INSTROOT}/etc/uci-defaults/99_owut-packagelist
+#!/bin/sh
+rm /usr/lib/opkg/status
+ln -s /etc/opkg/status /usr/lib/opkg/status
+EOF
+	endef
+
+	$(eval $(call BuildPackage,$(PKG_NAME)))
+
+endif
--- a/packages/lime-owut/README.md
+++ b/packages/lime-owut/README.md
@@ -1,0 +1,60 @@
+# lime-owut
+
+A script to add LibreMesh package repositories to owut build requests (available since openwrt-24.10).
+
+## Description
+To request a firmware build, owut asks it to an asu server (online imagebuilder) [0].
+This package by default instructs owut to use an asu instance [1] which allows adding feeds from libremesh.
+A source code with these changes to allow libremesh repositories within asu can be found at note [2].
+
+**owut**, OpenWrt Upgrade Tool, [3] _"is command line tool that upgrades your router's firmware. It creates custom images of OpenWrt using the sysupgrade server and installs them, retaining all of your currently installed packages and configuration."_
+
+In LibreMesh, it can be used to update the libremesh and openwrt packages:
+- while remaining with the same version of openwrt (like in the example below)
+- when updating to a newer version of openwrt, whether it is a minor update (e.g., 24.10.2 -> 24.10.3) or a major update (24.10.2 -> 25.xx.xx).
+
+**Note**: updating between major releases of OpenWrt is generally only “officially supported” between one major release and the next (i.e., 24.10.2 -> 26.xx.xx may work but is not officially supported). 
+
+## Example output running `owut download`
+
+```
+root@LiMe-870f23:~# owut download
+ASU-Server     https://sysupgrade.antennine.org
+Upstream       https://downloads.openwrt.org
+Target         mediatek/filogic
+Profile        cudy_wr3000s-v1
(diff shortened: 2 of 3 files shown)


ilario:
Wowow, this tool is nice!
owut is not in OpenWrt 23.05 repos, so I suppose it will generate a warning in the 23.05 buildroot, that can be safely ignored.

A couple of comments:

you could mention in the readme that the https://sysupgrade.antennine.org ASU server will be used.

can you explain either here of as comments in the code what does the `updates.missing = false;` string do?

does this tool upgrade between major versions (e.g. from LibreMesh 2020.4 to LibreMesh 2024.1) or only small things?

do you think this could be used inside the tools [eupgrade ](https://github.com/libremesh/lime-packages/tree/master/packages/eupgrade)(semi-automated upgrades, I never tried it) and [safe-upgrade](https://github.com/libremesh/lime-packages/tree/master/packages/safe-upgrade) (uses 2 partitions, works only with LibreRouter, then I never had the opportunity to try it) or are these completely independent tools?

a-gave:
> you could mention in the readme that the https://sysupgrade.antennine.org ASU server will be used.

Done

> can you explain either here of as comments in the code what does the `updates.missing = false;` string do?

Removed warnings, now it checks also for newer lime-packages

> does this tool upgrade between major versions (e.g. from LibreMesh 2020.4 to LibreMesh 2024.1) or only small things?

Yes it does the job between mayor versions - not really from 2020.4 because owut weren't present - but for subsequent ones it does

> do you think this could be used inside the tools [eupgrade ](https://github.com/libremesh/lime-packages/tree/master/packages/eupgrade)(semi-automated upgrades, I never tried it) and [safe-upgrade](https://github.com/libremesh/lime-packages/tree/master/packages/safe-upgrade) (uses 2 partitions, works only with LibreRouter, then I never had the opportunity to try it) or are these completely independent tools?

I don't know much of these tools, but it seems possible eventually

ilario:
Compiling with OpenWrt 25.12, I got these error messages referring to the Makefile:

```
bash: -c: line 1: syntax error near unexpected token `)'
bash: -c: line 1: `test $(echo 25.12-SNAPSHOT) | cut -d"." -f1) -lt 24; echo $?'
Makefile:17: *** missing 'endef', unterminated 'define'.  Stop.
```

ilario:
> Compiling with OpenWrt 25.12, I got these error messages referring to the Makefile:
> 
> ```
> bash: -c: line 1: syntax error near unexpected token `)'
> bash: -c: line 1: `test $(echo 25.12-SNAPSHOT) | cut -d"." -f1) -lt 24; echo $?'
> Makefile:17: *** missing 'endef', unterminated 'define'.  Stop.
> ```

@a-gave did you also observe this message?

packages/lime-owut/Makefile
```diff
@@ -0,0 +1,73 @@
...
+PKG_NAME:=lime-owut
+PKG_SOURCE_DATE:=2026-08-24
+PKG_RELEASE:=1
+
+PKG_SOURCE_PROTO:=git
+PKG_SOURCE_VERSION:=06139f3ad65c7971c6cc97c390dacde78f1dbe11
+PKG_SOURCE_URL:=https://github.com/a-gave/owut.git
+PKG_MIRROR_HASH:=035f319a43da7a3b3148ca9302a20763871ae91b82df514356886ba5c9b3602d
+
+PKG_MAINTAINER:=Eric Fahlgren <ericfahlgren@gmail.com>
```
ilario: @a-gave wouldn't make more sense to have your contact here, as maintainer of the lime-owut package?

a-gave: Right done, I'm trying to upstream the changes here https://github.com/openwrt/packages/pull/30374 and here https://github.com/efahl/owut/pull/82.
If these were integrated upstream, we would still maintain a lime-owut package that adds the uci-defaults for libremesh.

packages/lime-owut/Makefile
```diff
@@ -0,0 +1,73 @@
...
+PKG_MIRROR_HASH:=035f319a43da7a3b3148ca9302a20763871ae91b82df514356886ba5c9b3602d
+
+PKG_MAINTAINER:=Samuele Longhi <agave@dracaena.it>
+PKG_LICENSE:=GPL-2.0-only
+
+include $(INCLUDE_DIR)/package.mk
+
+define Package/$(PKG_NAME)
+  SECTION:=utils
+  CATEGORY:=Utilities
```
ilario: To have it in "Utilities" in menuconfig makes sense. Yet, it is also a LibreMesh-specific package, as it sets LibreMesh's ASU server URL.
Having this in the LibreMesh section of menuconfig would be convenient when one compiles LibreMesh with the BuildRoot, please consider moving it there (or leaving it here, up to you).

ilario:
> > Compiling with OpenWrt 25.12, I got these error messages referring to the Makefile:
> > ```
> > bash: -c: line 1: syntax error near unexpected token `)'
> > bash: -c: line 1: `test $(echo 25.12-SNAPSHOT) | cut -d"." -f1) -lt 24; echo $?'
> > Makefile:17: *** missing 'endef', unterminated 'define'.  Stop.
> > ```

Seems fixed now :)

packages/lime-owut/Makefile
a-gave: Changed the Category to 'LibreMesh'

The handling of external repositories has been included in owut here: https://github.com/efahl/owut/commit/2fab581877f3c17542b1836e2df8933554063ace

And the owut package in the openwrt's main branch already contains the latest code from owut
For the 25.12 branch there is an open pull request https://github.com/openwrt/packages/pull/30438
Maybe let's wait for this last one to be merged 

Then this pull request will be ready to be merged

a-gave:
This could be ready to go. https://github.com/openwrt/packages/pull/30438 has been merged so the handling of external repositories has been included in owut both in openwrt-main and in openwrt-25.12

To upgrade between fixed libremesh releases it could be used a script like `sed -i 's|master|2026.1|g' /etc/apk/repositories.d/limefeeds.list` (if coming from a libremesh built with a reference on the master branch)

I took another look at `lime-sysupgrade` which is a wrapper around `sysupgrade` that backup by default list of files defined in files listed in `lime-autogen.system.keep_on_upgrade='libremesh dropbear minimum-essential /etc/sysupgrade.conf` which seems to be definitively usable combined with owut. It should probably be documented better in the website.

ilario:
> This could be ready to go. [openwrt/packages#30438](https://github.com/openwrt/packages/pull/30438) has been merged so the handling of external repositories has been included in owut both in openwrt-main and in openwrt-25.12

I tried to test quickly and it seems working :D

> To upgrade between fixed libremesh releases it could be used a script like `sed -i 's|master|2026.1|g' /etc/apk/repositories.d/limefeeds.list` (if coming from a libremesh built with a reference on the master branch)

Ok, it is not super straightforward to change from full upgrade to upgrade within the same release, but seems ok to have the full upgrade as a default. So it is perfect as it is now.

> I took another look at `lime-sysupgrade` which is a wrapper around `sysupgrade` that backup by default list of files defined in files listed in `lime-autogen.system.keep_on_upgrade='libremesh dropbear minimum-essential /etc/sysupgrade.conf` which seems to be definitively usable combined with owut. It should probably be documented better in the website.

Mh, would it be something like using `owut download` to download the new image and then using `lime-sysupgrade` for flashing it to the device's memory?

ilario (approved):
I tested a bit, it seems good to me!

a-gave:
> Mh, would it be something like using `owut download` to download the new image and then using `lime-sysupgrade` for flashing it to the device's memory?

A minor problem is that the default keep_on_upgrades keeps both lime-community and lime-node which doesn't play well if you wish to keep only the lime-node, i.e. if the lime-community file is provided by a network profile or should be updated

Issue #1191, opened 2025-06-14
libstdcpp dependency of shared-state-async takes 2 MB


ilario:
Installing LibreMesh packages using OPKG, I observed this message:

```
Collected errors:
 * verify_pkg_installable: Only have 480kb available on filesystem /overlay, pkg libstdcpp6 needs 2070
 * opkg_install_cmd: Cannot install package shared-state-bat_hosts.
```

Unpacking the compiled libstdcpp6, I can confirm that its `/usr/lib/libstdc++.so.6.0.32` file is 2.1 MB, which makes it a problem for installation on routers with 8 MB of flash memory.

This dependency is defined here:
https://github.com/libremesh/lime-packages/blob/de9ff02e5b4fced2529d091f66fe32271c6d0a3a/packages/shared-state-async/Makefile#L45-L48

Is this dependency unavoidable?

ilario:
Already comparing the compressed size of the packages, it is clear that libstdcpp6 is the main problem for routers with not-huge flash. The list below is taken from a typical LibreMesh installation.

du size | package file
-- | --
656 | targets/ramips/mt76x8/packages/libstdcpp6_13.3.0-r4_mipsel_24kc.ipk
80 | targets/ramips/mt76x8/packages/kmod-batman-adv_6.6.86.2024.3-r5_mipsel_24kc.ipk
64 | packages/mipsel_24kc/base/liblua5.1.5_5.1.5-r11_mipsel_24kc.ipk
56 | packages/mipsel_24kc/routing/babeld_1.13.1-r2_mipsel_24kc.ipk
40 | packages/mipsel_24kc/routing/batctl-default_2024.3-r1_mipsel_24kc.ipk
28 | packages/mipsel_24kc/libremesh/lime-system_2025.06.08~1749408087_all.ipk
28 | packages/mipsel_24kc/luci/luci-lib-nixio_25.119.71430~be769af_mipsel_24kc.ipk
28 | packages/mipsel_24kc/packages/iputils-ping_20240905-r1_mipsel_24kc.ipk
12 | packages/mipsel_24kc/luci/luci-lib-ip_25.119.71430~be769af_mipsel_24kc.ipk
12 | targets/ramips/mt76x8/packages/kmod-macvlan_6.6.86-r1_mipsel_24kc.ipk
8 | packages/mipsel_24kc/base/libiwinfo-lua_2024.10.20~b94f066e-r1_mipsel_24kc.ipk
[...]

PR #1218, opened 2025-09-10
add suport for static linked library

javierbrk:
#1191 this makefile change enables the use of a statically linked shared state version

https://github.com/libremesh/shared-state-async/pull/2

It can be tested by compiling an image using this repos in the feeds 

src-git libremesh https://github.com/javierbrk/lime-packages.git;shared-state-async-static

This is just a draft and still requires further testing.

### DIFF
--- a/packages/shared-state-async/Makefile
+++ b/packages/shared-state-async/Makefile
@@ -39,7 +39,6 @@ define Package/shared-state-async
   TITLE+= (dynamic linking)
   DEPENDS:=+libstdcpp
   VARIANT:=dynamic
-  DEFAULT_VARIANT:=1
   CONFLICTS:=shared-state-async-static
 endef
 
--- a/packages/shared-state-async/Makefile
+++ b/packages/shared-state-async/Makefile
@@ -10,19 +10,12 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=shared-state-async
 
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/libremesh/shared-state-async.git
-PKG_SOURCE_DATE:=2024-05-13
-PKG_SOURCE_VERSION:=db58e3d16e8658417956f9bb42e7e87f6aadcd4d
+PKG_SOURCE_URL:=https://github.com/javierbrk/shared-state-async.git
+PKG_SOURCE_VERSION:=use-statically-linked-stdlib
 PKG_MAINTAINER:=Asociación Civil Altermundi <info@altermundi.net>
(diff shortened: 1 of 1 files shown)


javierbrk:
Another package or a package option, it must be possible to use both dynamic or static.

ilario:
Let us know when you consider this ready to merge, so that we test it :)

javierbrk:
Thanks for the feedback! I've been testing it myself first before sharing.
I think it will be better to add a different package that can be used
indistinctly just like with the original one.


Saludos
Ing. Javier Alejandro Jorge

ilario:
When running `make menuconfig`, I got this warning:

```
tmp/.config-package.in:65077:error: recursive dependency detected!
tmp/.config-package.in:65077:	symbol PACKAGE_shared-state-async depends on PACKAGE_shared-state-async-static
tmp/.config-package.in:65087:	symbol PACKAGE_shared-state-async-static depends on PACKAGE_shared-state-async
For a resolution refer to Documentation/kbuild/kconfig-language.rst
subsection "Kconfig recursive dependency limitations"
```

packages/shared-state-async/Makefile
```diff
@@ -10,19 +10,12 @@ include $(TOPDIR)/rules.mk
 PKG_NAME:=shared-state-async
 
 PKG_SOURCE_PROTO:=git
-PKG_SOURCE_URL:=https://github.com/libremesh/shared-state-async.git
-PKG_SOURCE_DATE:=2024-05-13
-PKG_SOURCE_VERSION:=db58e3d16e8658417956f9bb42e7e87f6aadcd4d
+PKG_SOURCE_URL:=https://github.com/javierbrk/shared-state-async.git
+PKG_SOURCE_VERSION:=use-statically-linked-stdlib
```
ilario, line 14: I got an error compiling on OpenWrt 25.12.
I think that is because OpenWrt 25.12's package manager apk (Alpine Package Keeper) does not like having dashes in the version numbering.
This is the error:
```
ERROR: info field 'version' has invalid value: package version is invalid
ERROR: failed to create package: package version is invalid
```

ilario: Ah hem, I did remove the dashes (that you re-introduced in 8e2c453cb9b33670f2106327465c56476e4fd11c) as they are not accepted when compiling with APK (used since OpenWrt 25.12) https://wiki.alpinelinux.org/wiki/APKBUILD_Reference#pkgver so it is ok if you use dashes during the testing, but please use a valid version number when the pull request will be ready for merge :)

PR #1227, opened 2025-10-31
Remove unused luci.model.network from groundrouting

Pablomonte:
I am unfamiliar with the functionality of the groundrouting package, but when analyzing possibilities for slimming down the bundle, we found that some dependencies could be removed.
The unit tests pass. And we test it on real hardware.
How could I verify that this change does not introduce any issues?

files/usr/libexec/rpcd/lime-groundrouting: remove unused ntm variable
Makefile: remove luci-compat dependency

The luci.model.network module was imported but never used (ntm variable defined but not referenced). This dependency pulled in luci-base and luci-lua-runtime unnecessarily.

Standalone luci-lib-jsonc (already in lime-system deps) is sufficient.


ilario:
Thanks!!

PR #1231, opened 2025-12-26
lime-system: wireless: add option distance_use_auto_if_available

### LINKED ISSUE #932: Distance setting problem and "auto" setting
Talking to community members seems that is common the experience of trying LibreMesh over long links and discard it as it "does not work". This is due to the default "distance" setting on the 5 GHz link (1 km):
https://github.com/libremesh/lime-packages/blob/78162b4d773c9e8d01f63801ce3464ea84ac89b6/packages/lime-system/files/etc/config/lime-defaults#L73

There is no "right" value (in my opinion, to have it too large is better than having it too small), and surely is of foremost importance to recommend in the documentation to change it.

Asking around at BattleMeshV14 we just realized that there is an "auto" setting for distance in ath9k driver.
The Aredn people set it and unset it a few times in their firmware https://github.com/aredn/aredn_ar71xx/search?q=distance&type=issues so seems like it had some problems, at least in the past.

I think we could recommend the users to select the
`CONFIG_PACKAGE_ATH_DYNACK=y` (in Menuconfig it is in Kernel modules > Wireless drivers > kmod-ath > Enable Dynack support) and set "auto" or we could even make it default with a fallback value.

Something like: if the distance value is negative (e.g. `-1000` but I would push for `-5000`) and have lime-config checking if Dynack is available (how??????), otherwise just apply the distance number. If the distance is positive, just use that value without checking.

a-gave:
lime-system: wireless: add option distance_use_auto_if_available

Closes #932

Introduce the wifi option `distance_use_auto_if_available` by default true
to enforce the value "auto" as distance for radio drivers that supports it
like kmod-ath9k when compiled with CONFIG_PACKAGE_ATH_DYNACK=y

Implement a check if driver supports auto setting.
At firstboot run 'iw phyX set distance auto' relying on the iw error
  i.e. 'command failed: Not supported (-95)' if unsupported
Save the result in lime-node in a format like `lime-node.radio0.dynack=0`
Re run the iw command on the same phy when the macaddress has changed
in case of powering off and plugging a different usb wireless adapter

Implement a check if distance 'auto' is requested for a driver that does not support it.
Use the lime-defaults value as safer option than allowing to set "auto", which could lead
to the minimum being applied at next reboot, potentially compromising long distance wireless links.

If auto is requested for a driver that doesn't support it
iw will fail with 'command failed: Not supported (-95)',
the value auto is not applied and the previous valid value i.e. '1000' is kept
However at next reboot if the value auto was saved in /etc/config/wireless
the coverage will be set to 0 on tested hardware.

To verify the applied coverage on a given phyX:
  iw phy phyX info | grep Coverage
In ath9k check the ack timeout in uS, followed by A for auto-determined or S set:
  cat /sys/kernel/debug/ieee80211/phy0/ath9k/ack_to

Further resources and links
- "[RFC] API for setting ACK timeout" thread on ml linux wireless
  https://www.spinics.net/lists/linux-wireless/msg43572.html
- Introduction in linux of setting coverage class for long distance wireless links:
  https://linux-wireless.vger.kernel.narkive.com/qo3X9KdF/patch-0-4-setting-coverage-class-and-ack-timeout-and-slot-time
- Introduction in iw of dynack:
  https://git.kernel.org/pub/scm/linux/kernel/git/jberg/iw.git/commit/?id=e642142d68850b7eb161489a984ec6817b10c51b
- Support of dynack in openwrt:
  https://git.openwrt.org/?p=openwrt%2Fopenwrt.git&a=search&s=dynack
- Comments from communities using dynack:
  https://www.arednmesh.org/content/auto-distance-setting-0

Due to different driver implementations, producing a list of wireless drivers
that support different distance/coverage optimizations for outdoor usage could be complex.
Looking at the source code a starting point could be to check drivers that handle NL80211_ATTR_WIPHY_COVERAGE_CLASS
  i.e. grep -rl 'distance\|coverage' <linux>/drivers/net/wireless/
Among those, currently dynack is supported just by ath9k

Tested on these devices running lime master and openwrt 24.10.x or snapshot:
- ubnt_litebeam-m5-xw
- gemtek_wvrtm-127acn
- gemtek_wvrtm-130acn
- cudy_wr3000s-v1
- openwrt_one (usb adapter rtl8192cu)

### DIFF
--- a/packages/lime-docs/files/www/docs/lime-example.txt
+++ b/packages/lime-docs/files/www/docs/lime-example.txt
@@ -154,6 +154,7 @@ config lime wifi
                                                         # that rescan all available frequencies in active radios.
 
     option unstuck_timeout '300'                        # Timeout in seconds that defines how long the mentioned above workaround should go.
+    option distance_use_auto_if_available true
 
 #########################################################
 ### WiFi specific band options
--- a/packages/lime-system/files/etc/config/lime-defaults
+++ b/packages/lime-system/files/etc/config/lime-defaults
@@ -64,6 +64,7 @@ config lime wifi
 	option ieee80211s_mesh_id 'LiMe'
 	option unstuck_interval '10'
 	option unstuck_timeout '300'
+	option distance_use_auto_if_available true
 
 config lime-wifi-band '2ghz'
 	option channel '11'
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -212,7 +212,7 @@ function wireless.configure()
 
 			local uci = config.get_uci_cursor()
 			uci:set("wireless", radioName, "disabled", 0)
-			uci:set("wireless", radioName, "distance", options["distance"])
+			uci:set("wireless", radioName, "distance", wireless.get_distance(radioName, options))
 			uci:set("wireless", radioName, "noscan", 1)
 			uci:set("wireless", radioName, "channel", channel)
 			if options["country"] then uci:set("wireless", radioName, "country", options["country"]) end
@@ -238,6 +238,8 @@ function wireless.configure()
 					                and (not key:match("distance"))
 					                and (not key:match("unstuck_interval"))
 					                and (not key:match("unstuck_timeout"))
+					                and (not key:match("dynack"))
+					                and (not key:match("phy_macaddr"))
 					                and (not (wireless.isMode(keyPrefix) and keyPrefix ~= modeName)))
 					if isGoodOption then
 						local nk = key:gsub("^"..modeName.."_", "")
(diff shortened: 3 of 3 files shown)


ilario:
Thanks!
I consider this as very useful!
Do you know if the auto distance is available on many routers? Only on ath9k?

A doubt on the code:
The function wireless.is_distance_auto_available works trying to set the distance to auto. If a set distance is specified in the options, will it be then applied later or the auto thing will remain? For example, after running lime-config is the active distance the one set in the config files, or the distance remains to auto and the user should reboot for applying the correct distance?

a-gave:
> Thanks! I consider this as very useful! Do you know if the auto distance is available on many routers? Only on ath9k?
> 
> A doubt on the code: The function wireless.is_distance_auto_available works trying to set the distance to auto. If a set distance is specified in the options, will it be then applied later or the auto thing will remain? For example, after running lime-config is the active distance the one set in the config files, or the distance remains to auto and the user should reboot for applying the correct distance?

I updated the description. If the distance option contains a valid value it will be applied at next reboot or after running `wifi`, if the wrong value is wrote in /etc/config/wireless running `wifi` won't apply it but at next reboot the value will be set to 0.

ilario:
Ah, now I see :)
So the command `iw " .. phy .. " set distance auto` gets executed only the first time a radio is installed, not at every execution of `lime-config`. Nice!!

ilario:
> Seems the same. In this case i was wondering if writing it in lime-autogen would be better. Are you suggesting it should be so?

I don't know what is better. But maybe using the same mechanism for this and for lime-hwd-openwrt-wan would make things easier to understand. From a user point of view, I think that it would be much more convenient if both used lime-node. But I suspect that from the technical point of view it makes more sense to keep lime-node for user-written stuff and lime-autogen for autigenerated things...

> > Question: on the devices where this is possible, does this work with the default OpenWrt kernel selected options? I mean, does it work when creating images with OpenWrt's ImageBuilder or one should enable this feature and create an image with it using the BuildRoot?
> 
> Nope. It should be enabled during the kernel compilation, so:
> 
> * yes BuildRoot
> * no OpenWrt's ImageBuilder
> * yes ImageBuilders compiled with the reduced kernel like https://hub.docker.com/layers/agave0/openwrt-imagebuilder/ath79-generic-24.10.4-alpine-small-flash

Argh. We could check how much does the image grow due to this option (e.g. compiling with and without, then checking the size of the stuff in the sysupgrade.bin files using binwalk). Then, if the chame is negligible, we could propose OpenWrt to activate it by default in the future releases (very likely it's too late for including this change in 25.12).

ilario (approved):
Tested on TP-Link WDR3600, it works :D
I did not try what happens if radio0 changes to be another radio.
Thanks!

a-gave:
Thanks! 
Submitted a pr to try to enable it by default in https://github.com/openwrt/openwrt/pull/23881

PR #1233, opened 2026-01-02
config: Call hotplug hooks in more places

rah2501:
Also add an ACTION environment variable containing the same value as passed as a command-line argument.

Some scripts presume they will only be called in the one existing place so they need to be updated also.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/config.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/config.lua
@@ -182,7 +182,27 @@ function config.uci_commit_all()
     end
 end
 
+function config.execute_hooks(action)
+	for hook in fs.dir(config.hooksDir) do
+		local hookCmd = config.hooksDir.."/"..hook
+		local shCmd = "sh -c 'ACTION="..action.." "..hookCmd.."'"
+		print("Executed hook:", shCmd, os.execute(shCmd))
+	end
+end
+
+--! FIXME: This should be somewhere more appropriate
+function config.file_exists(name)
+	return fs.stat(name, "type") == "reg"
+end
+
 function config.main()
+	--! Check whether this is the first ever run
+	if not config.file_exists(config.uci:get_confdir() .. "/" .. config.UCI_AUTOGEN_NAME) then
+		config.execute_hooks("init")
+	end
+
+	config.execute_hooks("pre")
+
 	--! Get mac address and set mac-based configuration file name
 	local network = require("lime.network")
 	local utils = require("lime.utils")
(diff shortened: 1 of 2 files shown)


packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -182,7 +182,33 @@ function config.uci_commit_all()
...
+
+--! FIXME: This should be somewhere more appropriate
+function config.file_exists(name)
+	local cf = io.open(name)
+	if not cf then
+		return false
+	else
+		cf:close()
+		return true
+	end
```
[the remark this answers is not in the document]
rah2501: Mnyaaaa.. good question.  I looked at this myself.  Initially the function was just moving around existing code.  When I looked at the general consensus on how to do this in Lua, I found the following:

"Using plain Lua, the best you can do is see if a file can be opened for read, as per LHF."
-- https://stackoverflow.com/questions/4990990/check-if-a-file-exists-with-lua

I believe "LHF" here refers to Luiz Henrique de Figueiredo, the creator of Lua.

So this is both in line with the Lua way of doing things, and it's what the existing code already does so I left it as is.  That said, I'm certainly open to changing it to be more POSIXey but I was trying to do things in a Lua way because my Lua experience is very limited.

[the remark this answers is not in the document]
pony1k: Good point. Both seems fine to me.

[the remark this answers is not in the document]
G10h4ck: I suggest to switch to @pony1k suggested version as it is more compact and we already depend on those libraries

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -182,7 +182,33 @@ function config.uci_commit_all()
     end
 end
 
+function config.execute_hooks(action)
```
G10h4ck: This function or something very similar feels like to be used/needed in more place in LiMe code, let's try to avoid duplication

G10h4ck (changes requested):
See inline comments

packages/lime-system/files/usr/lib/lua/lime/config.lua
```diff
@@ -182,7 +182,33 @@ function config.uci_commit_all()
...
+	for hook in fs.dir(config.hooksDir) do
+		local hookCmd = config.hooksDir.."/"..hook.." "..action
+		local shCmd = "sh -c 'ACTION="..action.." "..hookCmd.."'"
+		print("Executed hook:", shCmd, os.execute(shCmd))
```
a-gave: The `.." "..action` in L187 causes the 'action' to be printed twice in the logs, and passed unnecessarily as an argument to the hotplug script.
Since it seems that the scripts in /etc/hotplug.d/\*/\* generally handle the ACTION env variable and none handle $1, would you mind to remove it?

packages/lime-system/files/usr/lib/lua/lime/config.lua
rah2501: > have you double checked?

I haven't single checked :-)  I'm not sure what your expectation is here.  What do you expect me to have checked?  Bearing in mind I don't know either the codebase or Lua well at all.  Can you tell me where you would expect me to check?

a-gave: This pr could be ok as is! Let's merge it as is to allow to go further.
I'm rewriting some packages related to watchcat and the changes in this pull request are needed.
I'll take care of checking other packages that could be rewritten to use these hooks!

PR #1234, opened 2026-01-03
Reign in the nuking of wireless interfaces

rah2501:
Don't nuke all the wireless interfaces, just the interfaces that LibreMesh itself creates.  Except for the OpenWrt-generated default interfaces, which are disabled.

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/wireless.lua
@@ -20,10 +20,31 @@ function wireless.get_phy_mac(phy)
 	return utils.split(mac, ":")
 end
 
+function wireless.pre()
+	local uci = config.get_uci_cursor()
+
+	local function disable_default(s)
+		if s[".name"] == "default_" .. s["device"] then
+			print("Disabling default wireless interface '" .. s[".name"] .. "'")
+			uci:set("wireless", s[".name"], "disabled", "1")
+		end
+	end
+	uci:foreach("wireless", "wifi-iface", disable_default)
+
+	uci:save("wireless")
+end
+
 function wireless.clean()
 	utils.log("Clearing wireless config...")
 	local uci = config.get_uci_cursor()
-	uci:foreach("wireless", "wifi-iface", function(s) uci:delete("wireless", s[".name"]) end)
+
+	local function delete_lime_section(s)
+		if utils.stringStarts(s[".name"], wireless.limeIfNamePrefix) then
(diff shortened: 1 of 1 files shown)


G10h4ck:
@rah2501 are you sure keeping the default interface disabled but around doesn't create any problem? The nuking was most probably due to historical quirks, with disabled things remaining around.

rah2501:
> @rah2501 are you sure keeping the default interface disabled but around doesn't create any problem?

It's not possible for any person to be 100% sure of such a thing.  Certainly it hasn't created any problems for me and I can't envision how it would create a problem for others.

G10h4ck:
Testing it on the hardware you have at hands seems good enough to me

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
```diff
@@ -20,10 +20,31 @@ function wireless.get_phy_mac(phy)
 	return utils.split(mac, ":")
 end
 
+function wireless.pre()
```
G10h4ck, line 23: can the name of the function be a little more descriptive/longer? think of someone that will read this code 3 years later, also please add some inline documentation

G10h4ck (changes requested):
See inline comments

packages/lime-system/files/usr/lib/lua/lime/wireless.lua
rah2501: > can the name of the function be a little more descriptive/longer?

No, this function name is part of the module API:

https://github.com/libremesh/lime-packages/blob/b88562f6cb58d8163a8c3557c647da92a140fb55/packages/lime-system/files/usr/lib/lua/lime/config.lua#L206

> think of someone that will read this code 3 years later, also please add some inline documentation

No other module API functions note that they're module API functions and it's a pretty simple function which even prints a comment about what it's doing so it doesn't really seem necessary to me.

G10h4ck: didn't know that was the case, I am not sure now but I think the API function name could have been a bit more descriptive in the first place. Anyway good enough for this PR.

G10h4ck (approved):

G10h4ck (approved):

ilario:
Sounds good to me. I tested it in qemu and it works :)
I understand that the idea behind this is that one could add interfaces and avoid that lime-config deletes them, right?

ilario:
Thanks @rah2501 !

PR #1235, opened 2026-01-04
lime-system: activate mesh_nolearn

### LINKED ISSUE #1232: lime-system: activate mesh_nolearn if batman-adv is used
I stumbled upon this while trying to find/produce a current list of wireless drivers that support ‘auto’ as a distance value (https://github.com/libremesh/lime-packages/pull/1231)

Related links:
1. from Linux https://github.com/torvalds/linux/blob/7d0a66e4bb9081d75c82ec4957c50034cb0ea449/include/uapi/linux/nl80211.h#L5017-L5021

```
 * @NL80211_MESHCONF_NOLEARN: Try to avoid multi-hop path discovery (e.g.
 *      PREQ/PREP for HWMP) if the destination is a direct neighbor. Note that
 *      this might not be the optimal decision as a multi-hop route might be
 *      better. So if using this setting you will likely also want to disable
 *      dot11MeshForwarding and use another mesh routing protocol on top.
```
2. introduced in iw https://git.kernel.org/pub/scm/linux/kernel/git/jberg/iw.git/commit/?id=94ded888ba696a1ec03571d596e96cf944ccc0c8
3. from the commit message in the addition in Gluon https://github.com/freifunk-gluon/gluon/pull/3512/commits/768172c7cd088da3c8c492af3c1ea4b4206140b2 (see also https://github.com/freifunk-gluon/gluon/pull/3512#issuecomment-2932230284)
> We do not use the multi-hop mesh routing capabilities of 802.11s and
> set mesh_fwding=0 in gluon unconditionally.
> In turn there is no use in discovering 802.11s mesh paths,
> as there are no multi-hop mesh paths due to the lack of forwarding.
> 
> While the router identifies this on its own after a few ms,
> it is simply not necessary to waste resources on this discovery.

a-gave:
Closes #1232

Set the wifi option ieee80211s_mesh_nolearn='1' in lime-defaults in a similar way to how ieee80211s_mesh_fwding is set to 0.

### DIFF
--- a/packages/lime-docs/files/www/docs/lime-example.txt
+++ b/packages/lime-docs/files/www/docs/lime-example.txt
@@ -145,6 +145,7 @@ config lime wifi
     option adhoc_ssid 'LiMe'                            # SSID of the APs (nodes) when meshing in ad-hoc mode, i.e., the nodes form an IBSS. Not used when meshing in 802.11s (the default)
     option adhoc_bssid 'ca:fe:00:c0:ff:ee'
     option ieee80211s_mesh_fwding '0'                   # Settings needed only for 802.11s
+    option ieee80211s_mesh_nolearn '1'                  # Disable multi-hop mesh routing capabilities of 802.11s
     option ieee80211s_mesh_id 'LiMe'                    # Mesh cloud identifier (close to SSID in concept). Used by the nodes to join and participate in the mesh network.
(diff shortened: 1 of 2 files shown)


ilario:
Just for me to remember how these options work:
They get parsed here and passed to the mode:
https://github.com/libremesh/lime-packages/blob/c95c7bc20592eb1a24cab4eca3fa7d9fa051fb61/packages/lime-system/files/usr/lib/lua/lime/wireless.lua#L228-L251

The edit is justified in #1232 and the code edit seems safe. I didn't manage to test but I think this is safe to merge.

PR #1246, opened 2026-03-13
Add USB tethering support via ubus methods

usg-ishimura:
This PR adds WAN connection in LiMe through USB tethering or 4G USB dongle (not tested), in this way any node can get WAN connection even if there is no wired network where the routers are deployed or there is only one wireless chip per router `radio0` so the hotspot feature cannot be used. The implemented ubus APIs can be called like so:
```
ubus call wan-service get_wan_status '{}'
ubus call wan-service-admin set_wan_interface '{"ifname":"eth1"}'
ubus call wan-service-admin set_wan_interface '{"ifname":"usb0"}'
```
This feature works with the following kernel modules included at build time: `kmod-usb-net` and `kmod-usb-net-rndis`


packages/ubus-lime-utils/files/usr/lib/lua/lime/wan_service.lua
```diff
@@ -0,0 +1,156 @@
...
+    if link_type == "1" then
+        return "ethernet"
+    end
+
+    return nil
+end
+
+--! Returns true if the interface is currently present (visible in sysfs).
+--! Uses io.open, no unsafe_shell.
+local function iface_exists(ifname)
```
ilario: Maybe `network.interface_exists` can be used here? Please have a look, does it do the same?
https://github.com/libremesh/lime-packages/blob/1e0e15ecaeedec454ff4e3a6134b371699b6d862/packages/lime-system/files/usr/lib/lua/lime/network.lua#L61-L63

ilario:
Sorry for taking so long to review! Looking at the code it seems perfect. I still have to test it on real hardware with a phone connected doing USB tethering.

packages/ubus-lime-utils/files/usr/lib/lua/lime/wan_service.lua
usg-ishimura: Yes, the functions do the same thing, but the version in `network.lua` is more efficient because it doesn't open and read the file; instead, it simply reads the symbolic link's metadata via a system call. I'll update the function used by `wan_service.lua` and test as soon as I can.

Issue #1253, opened 2026-05-16
lime-config fails if the device doesn't have a network device named 'eth0'


a-gave:
Found trying to run libremesh on a tplink-eap650-outdoor-v1 which have a physical interface named 'lan'

The device is currently unsupported (see https://github.com/openwrt/openwrt/pull/23056) but the issue should be valid for other routers without a network device named 'eth0' like the tplink-eap610-outdoor-v1 which is currently supported.

**Details:**
When configuring the network lime-config expects that `lime-autogen` contains a config `network.primary_interface` else it will use 'eth0'
If eth0 is not present the first lime-config fails to run. 
The strange fact is that it is not sufficient to modify the lime-defaults `network.primary_interface`  to 'auto' or 'lan'
The workaround I found was to copy also the file lime-defaults to lime-autogen.
It seems that the function 'network.primary_interface()' requires a lime-autogen with a config `network.primary_interface`  to be present before it's first run.
The device remains in a sort of bricked state, recoverable via serial interface. 
I'll add some more details and the error log once I reproduce the issue.

https://github.com/libremesh/lime-packages/blob/aebe482d278cfb9aac14a73ed4a5f8094250dab2/packages/lime-system/files/usr/lib/lua/lime/network.lua#L68
```
function network.primary_interface()
	local ifname = config.get("network", "primary_interface", "eth0")
	if ifname == "auto" then
		local board = utils.getBoardAsTable()
		ifname = board['network']['lan']['device']
	end
	network.assert_interface_exists(ifname)
	return ifname
end
```

ilario:
Ooh!
Do you think that we could solve setting the value to `auto` both in lime-system's lime-defaults file and in the `config.get` fallback value?

a-gave:
> Ooh! Do you think that we could solve setting the value to `auto` both in lime-system's lime-defaults file and in the `config.get` fallback value?

This works for the device in question! But will break for sure all other devices which have a more common 'eth0'. Tested changing the code and lime-defaults as you suggested and loading an initramfs on the device with the lan port renamed from `lan` to `eth0`.

I imagine that eth0 could be left as default value in lime-defaults. But I still need to figure out which could be a better fallback mechanism in case an eth0 port doesn't exist

This was the error messages
```
Executed hook:	sh -c 'ACTION=init /etc/hotplug.d/lime-config/babeld-auto-gw.lua'	0
[...]

PR #1254, opened 2026-06-05
fix: lime-config improve the fallback to network.primary_interface

a-gave:
Force a fallback to auto in case the specified primary_interface doesn't exist

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/network.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/network.lua
@@ -55,20 +55,72 @@ function network.get_own_macs(interface_filter)
 	return result
 end
 
+function network.interface_exists(ifname)
+	return fs.lstat("/sys/class/net/"..ifname) and true or false
+end
+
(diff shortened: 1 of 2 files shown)


ilario:
Reading the code it seems perfect, I like also the concept discussed in https://github.com/libremesh/lime-packages/issues/1253#issuecomment-4637985152
I hope I will manage to test this in a few days :)
Thanks!!!

ilario:
The old function `network.assert_interface_exists` could call the new `network.interface_exists`.

The tests are failing here: 

https://github.com/libremesh/lime-packages/blob/0bf768ec4315bf93a3e9acba6ad4765f83c36666/packages/lime-system/tests/test_lime_network.lua#L36-L38

In that test, `assert_interface_exists` gets redefined, so for keeping the same logic maybe we should redefine also `interface_exists` there.

a-gave:
Updated the commit description after a minor cleaning, fixed the tests and introduced an overwrite of the lime-autogen.network.primary_interface setting 'auto' to check only once the existence of 'eth0' during the execution of lime-config.

ilario (approved):
Great! Thanks!

PR #1255, opened 2026-06-13
Add rpcd dependency in lime-system

ilario:
As reported here https://github.com/libremesh/lime-packages/pull/1245#issuecomment-4119550080 and also reported by @rah2501 on the Matrix chat, when compiling using the BuildRoot method, `lime-system` does not appear in menuconfig unless Base system > `rpcd` gets selected first.
This happens because without selecting rpcd, rpcd-mod-rpcsys cannot be selected, and lime-system depends on the latter since #1245.


rah2501:
Can attest that adding rpcd fixes the build system.

a-gave (approved):
Merged thanks!

PR #1256, opened 2026-06-15
Consolidate module handling code into a new script

rah2501:
This makes the interface to module functions much more robust and clean and also as a side-effect, makes the config.lua script much cleaner.

While we're at it, an init() function for lime modules called once on first boot (essentially a one-liner with the consolidation).  Note that it is called from /etc/init.d/boot and not uci-defaults so the stock OpenWrt wireless config will be present.


ilario:
The code looks great to me :D
I will try to test

ilario (approved):
The images compiled including this commit seemed working just fine. This contribution makes the code cleaner and adds the option for a "pre" step to be included in the modules. Thanks!

PR #1257, opened 2026-06-18
feat(ci): add LibreRouter v1 support via dual-TFTP boot

francoriba:
## Summary

- Add `IMAGE_FORMAT=dual-tftp` to `build_image.sh` that emits `kernel.bin` + `rootfs.cpio` as separate TFTP artifacts
- Add `librerouter_v1` target entry in `targets.yml` (mips_24kc, ath79-generic, dual-tftp)
- Extend `lab_stage_firmware.sh` to detect and stage dual-file artifacts (`LG_IMAGE` + `LG_IMAGE_INITRD`)
- Forward `LG_IMAGE_INITRD` env var in the workflow test step

U-Boot TFTP-loads kernel and rootfs CPIO to distinct RAM addresses, then passes `rd_start`/`rd_size` via bootargs so the MIPS kernel mounts the external CPIO as rootfs. This avoids the need for a full source build (ImageBuilder is used for everything).

Depends on: fcefyn-testbed/libremesh-tests@staging (strategy + target YAML changes).

## Test plan

- [ ] CI build succeeds for librerouter_v1 (mips_24kc feed + dual-tftp image)
- [ ] Existing targets (linksys_e8450, openwrt_one, qemu_x86_64, bananapi_bpi-r4) unaffected
- [ ] Manual hardware test: verify two sequential TFTPs + rd_start/rd_size on LibreRouter U-Boot 1.1.x


francoriba:
Wrong target repo - creating PR in the fork instead

PR #1258, opened 2026-06-29
ci: add linting via luacheck and shellcheck

a-gave:


ilario:
Excellent!
Automatically checking the code from the pull requests will surely improve its quality without putting any extra weight on the maintainers :D
But, before merging, we should fix aaaaaalll the issues that these tools find with the current code, right? Otherwise the checks will always fail, right?

ilario:
I would merge this even if the test is going to fail always, as it allows the developer to check if it is failing also in the code that is being proposed.

ilario (approved):
This will be useful, even if the output is going to be polluted by plenty of pre-existing suboptimal code lines.

PR #1263, opened 2026-07-17
Fix linting

a-gave:
Related to #1258

Lint all files with luacheck and shellcheck.

Another missing point would be to restyle the code using stylua, but this will be done in a separate commit to avoid mixing two pull request that change so many lines of code.


a-gave:
I double checked the changes and I'm pushing these changes to the branch 'test' to compile and test on real hardware

francoriba (approved):
lgtm! 🤞🏼

a-gave:
Tested on: gemtek-wvrtm127acn with default packages and openwrt-25.12.5

I'm merging this to go on with the introduction of a check of the formatting via stylua and shfmt

PR #1265, opened 2026-07-18
fix: remove WAN link-local firewall exception

### LINKED ISSUE #1241: lime-proto-wan fails trying to allow link local traffic on WAN. Fix or erase?
Restarting the firewall using `/etc/init.d/firewall restart` on a LibreMesh with the latest LibreMesh code on top of OpenWrt 24.10, I observed this message:

`Section lime_allow_wan_all_link_local (lime_allow_wan_all_link_local) is restricted to IPv4 but referenced source IP is IPv6 only, skipping`

That is caused by this section present in `/etc/config/firewall`:

```
config rule 'lime_allow_wan_all_link_local'       
        option name 'lime_allow_wan_all_link_local'
        option src 'wan'                    
        option family 'ipv6'                    
        option src_ip 'fe80::/10'                  
        option dest_ip 'fe80::/10'                
        option target 'ACCEPT'                    
        option family ipv4
```

The `option family` is specified twice, which is wrong.
Weirdly, in `lime-proto-wan`, only `option family ipv6` is present:

https://github.com/libremesh/lime-packages/blob/5f955d5870f4055610905e620a611c375e7be5de/packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua#L31-L42

Running again `lime-config` made disappear that `option family ipv4` line, so now I am a bit confused, can anyone confirm this?

Anyway, before trying to fix, **I would like to make sure that we still want to have this section**. What this section means is that even if the firewall blocks all ports (e.g. blocks SSH connections on port 22 coming from the internet), it does not block any of them if IPv6 is used for the connection. Which means that the internet cannot access your port 22, but that your internet service provider can. Or whatever you have on the other side of the WAN cable, which can be a not-so-trusted network. I think that this is a bit risky AND nobody would expect this, also because it is not documented at all.

Negative aspect of having that section: lowers security without the user expecting it.
Positive aspect of having that section: an additional way to log in when some seriously bad configuration blocks you from using the LAN ports or the WiFi access point or the WiFi mesh.

This has been added by @altergui and approved by @p4u and @G10h4ck in https://github.com/libremesh/lime-packages/pull/239. Do we still want this?

w3lld1:
## Summary

I removed the `lime_allow_wan_all_link_local` firewall rule from `lime-proto-wan`, so configuring a WAN interface no longer opens every service to IPv6 link-local peers on that interface.

I also added a focused regression test that keeps the WAN device setup intact while checking that the firewall exception is absent.

## Validation

I ran:

- `luac5.1 -p packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua packages/lime-proto-wan/tests/test_wan.lua`
- focused Lua harness against `wan.setup_interface` (passes on this change and fails on `master` because `master` still modifies the firewall)
- `luacheck packages/lime-proto-wan/files/usr/lib/lua/lime/proto/wan.lua --no-unused --globals wan`
- `luacheck packages/lime-proto-wan/tests/test_wan.lua --globals describe it before_each after_each stub assert`
- `git diff --check`

The full `./run_tests` suite is deferred to CI because the local Docker socket is not accessible.

Fixes #1241


packages/lime-proto-wan/tests/test_wan.lua
```diff
@@ -0,0 +1,24 @@
...
+local wan = require 'lime.proto.wan'
+
+local uci
+
+describe('LiMe WAN protocol tests #protowan', function()
+    it('does not allow link-local traffic from WAN', function()
+        wan.setup_interface('eth0', {})
+
+        assert.is.equal('eth0', uci:get('network', 'wan', 'device'))
+        assert.is_nil(uci:get('firewall', 'lime_allow_wan_all_link_local'))
```
ilario: There is no need to test if this rule is effectively gone. This looks like an AI quirk.

ilario:
Thanks @w3lld1 for the contribution, I can confirm that this fixes #1241 :)
I still have to run the test you added. Please check my comment in the code of the test: at first sight it seems that AI added a line that makes very little sense.

packages/lime-proto-wan/tests/test_wan.lua
w3lld1: Removed the unnecessary regression test in 05460dee. The PR now only removes the WAN link-local firewall exception.

ilario (approved):

G10h4ck:
No please this feature was needed, and still is needed! The exception doesn't open the door to the whole IPv6 internet but just to link local addresses, it means only peers that are attached to the same switch as your router!
So the issue should be fixed, do not remove the needed feature! It also breaks babel routing on wan interface.

a-gave:
> So the issue should be fixed, do not remove the needed feature! It also breaks babel routing on wan interface.

Thanks a lot for the clarification! I was wondering if it would break babel on wan.

PR #1268, opened 2026-07-23
Revert "fix: remove WAN link-local firewall exception"

G10h4ck:
Reverts libremesh/lime-packages#1265

This feature was needed, and still is needed! The exception doesn't open the door to the whole IPv6 internet but just to link local addresses, it means only peers that are attached to the same switch as your router!
So the issue should be fixed, do not remove the needed feature! It also breaks babel routing on wan interface, and any other protocol which uses IPv6 link local for discovery or any other operation on wan interface.


a-gave (approved):

ilario:
This feature has been broken since several years. If you want to propose to re-enable it, at least fix the issue reported in #1241 and answer to my comment in https://github.com/libremesh/lime-packages/issues/1241#issuecomment-5058355701

PR #1271, opened 2026-08-14
shared-state-dnsmasq_leases: move broken dhcpscript to /etc/hotplug.d/dhcp/

a-gave:
Using dhcpscript the commands in /usr/bin/dnsmasq-lease-share.sh are sourced by /usr/lib/dnsmasq/dhcp-script.sh.
The execution of shared-state-publish_dnsmasq_leases was already broken, but it was not visible
until a recent change from '&>/dev/null' to '>/dev/null' that revelead the stderr.
```
Fri Aug 14 14:16:35 2026 daemon.debug dnsmasq-script[1]: /usr/lib/dnsmasq/dhcp-script.sh: /usr/bin/dnsmasq-lease-share.sh: line 2: /usr/bin/shared-state-publish_dnsmasq_leases: not found
```

The failure of the command is probably related to the usage of ujail, it could be tested removing
the default package procd-ujail. Cfr. [0].

Moved instead the command to /etc/hotplug.d/dhcp/00-lease-share that fires only on ACTION 'add' and 'del' and not on 'update'.

Since the previous was instead continuously failing and un-neededly running on updates:
as per the openwrt wiki the usage of dhcpscript do: `Run a custom script upon DHCP lease add / renew / remove actions` [1]

[0] https://forum.openwrt.org/t/dhcp-script-wont-execute/139587
[1] https://openwrt.org/docs/guide-user/base-system/dhcp?s[]=dhcpscript

### DIFF
--- a/packages/shared-state-dnsmasq_leases/files/etc/hotplug.d/dhcp/00-lease-share
+++ b/packages/shared-state-dnsmasq_leases/files/etc/hotplug.d/dhcp/00-lease-share
@@ -1,0 +1,4 @@
+#!/bin/sh
+if [ "$ACTION" = "add" ] || [ "$ACTION" = "del" ]; then
+	(shared-state-publish_dnsmasq_leases "$@" >/dev/null) &
+fi
--- a/packages/shared-state-dnsmasq_leases/files/etc/uci-defaults/90_dnsmasq-lease-share
+++ b/packages/shared-state-dnsmasq_leases/files/etc/uci-defaults/90_dnsmasq-lease-share
@@ -2,7 +2,6 @@
 
 dhcpHostsFile="/tmp/dhcp.hosts_remote"
 
-uci set dhcp.@dnsmasq[0].dhcpscript=/usr/bin/dnsmasq-lease-share.sh
 uci set dhcp.@dnsmasq[0].dhcphostsfile=$dhcpHostsFile
 uci commit dhcp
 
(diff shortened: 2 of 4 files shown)


javierbrk:
> The failure of the command is probably related to the usage of ujail, it could be tested removing
> the default package procd-ujail. Cfr. [0].


Yes!! I see! the service is running inside a procd jail and I confirm that it can't reach the binary. The file "/etc/init.d/dnsmasq" defines what the service "can see", even though the binary is in the path, the service won't be able to access or run it because  /usr/bin/shared-state-publish_dnsmasq_leases is never added to any of the mount lists. 

I think that the proposed solution is the most standard way, another way it wold be to include the binary as a mount for this service. 

I tried adding the script to my router and it is being called. 

As a side quest, It can be interesting to add a test to a virtual testbed where one node connects to another node as a  client and the node announces the assigned IP over shared state.

javierbrk (approved):

a-gave:
> > The failure of the command is probably related to the usage of ujail, it could be tested removing
> > the default package procd-ujail. Cfr. [0].
> 
> Yes!! I see! the service is running inside a procd jail and I confirm that it can't reach the binary. The file "/etc/init.d/dnsmasq" defines what the service "can see", even though the binary is in the path, the service won't be able to access or run it because /usr/bin/shared-state-publish_dnsmasq_leases is never added to any of the mount lists.
> 
> I think that the proposed solution is the most standard way, another way it wold be to include the binary as a mount for this service.
> 

I didn't found the documentation about adding custom mounts for the dhcpscript, and to try to keep the dhcpscript but avoiding acting on 'updates'
I would keep the proposed `/etc/hotplug.d/dhcp/` script since it seems more flexible and permits to have multiple scripts to be executed in the folder. 

> I tried adding the script to my router and it is being called.
> 
> As a side quest, It can be interesting to add a test to a virtual testbed where one node connects to another node as a client and the node announces the assigned IP over shared state.

That's very interesting to me too!

I'm merging this as is for now.

PR #1272, opened 2026-08-14
ci: build-firmware: SDK use EXTRA_DEPENDS for lime-packages to speed up the compilation

a-gave:
The rest of the lime-packages built via by the ci via the action gh-action-sdk are compiled with NO_DEFAULT_FEEDS.
No issue so far since there is no libremesh package that needs to have one of its dependency actually compiled during it's own compilation.
Instead it speed up the compilation of a single architecture from 40m-1h to 10-15min.

### DIFF
--- a/.github/workflows/build-firmware.yml
+++ b/.github/workflows/build-firmware.yml
@@ -110,7 +110,7 @@ jobs:
         env:
           ARCH: ${{ matrix.sdk_arch }}
           FEEDNAME: lime_packages
-          PACKAGES: ${{ needs.prepare-matrix.outputs.lime_packages_list }} ${{ matrix.extra_packages }}
+          # PACKAGES: ${{ needs.prepare-matrix.outputs.lime_packages_list }} ${{ matrix.extra_packages }}
           # Extra src-git feeds, space-separated `<type>|<name>|<url>[^<sha>]`
           # entries; gh-action-sdk appends each to feeds.conf. Pinning the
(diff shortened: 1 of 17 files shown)


ilario (approved):
It seems excellent to me :)

a-gave:
mmm.. looking instead at the output of the action now. It seems that the build_feed actually needs some openwrt packages to build the initramfs..
maybe it could be restructured with a list of these required tools and install only them

ilario:
Is it `vwifi`?
Likely we are using it for the tests on qemu...

a-gave:
> mmm.. looking instead at the output of the action now. It seems that the build_feed actually needs some openwrt packages to build the initramfs.. maybe it could be restructured with a list of these required tools and install only them

No sorry, the initramfs part is done using the imagebuilder.

> Is it `vwifi`? Likely we are using it for the tests on qemu...

Yes it seems that there is still an issue with vwifi for x86_64, didn't realize how to fix yet.

ilario:
Just checked the code. It seems eeeexcellent to me!!!!
Please merge when you consider to :)

a-gave:
> Just checked the code. It seems eeeexcellent to me!!!! Please merge when you consider to :)

Thanks added only a missing check that the sed regex to transform DEPENDS to EXTRA_DEPENDS is done only inside the portion of the makefiles that contains a definition of a package.

I'm merging this.

PR #1274, opened 2026-08-27
treewide: reword lime_release and extend the usage of libremesh.mk

### LINKED ISSUE #766: Migrate LibreMesh-specific packages' Makefile to libremesh.mk new format

a-gave:
- libremesh.mk: extend the usage to all packages without a pkg_source_url
- lime-debug: use libremesh.mk
- lime-docs: fix banner not displayed
- lime-system: use libremesh.mk to remove comments from the built package
- lime-system: drop compatibility script for v2020.4 migrate-wifi-bands 
- lime-system: reword /etc/lime_release 'development' to 'nightly' 
which is shorter and emphasizes the fact that the code is usable but it is not a fixed/stable release. While 'development' is more vague.
- lime-system: reword lime_description and add it to /etc/os-release
from: LiMe master development (master rev. 0b6b5c5 20260810_1130)
to:   LibreMesh 25.12.5 r26.222.40775~0b6b5c5 nightly
This text then displayed:
   - in the lime-app
   - via ssh after the banner via the profile.d/10-lime-banner.sh
   - via cli running `ubus call system board`
- shared-state-async: symlink script and add the missing dependency on shared-state

Fixes: #766


a-gave:
Edited the description to mention the removal of comments from the package lime-system using libremesh.mk

packages/lime-system/files/etc/uci-defaults/89_lime-release
```diff
@@ -4,11 +4,18 @@
...
-	-e s/"^DISTRIB_DESCRIPTION=.*"/"DISTRIB_DESCRIPTION='$LIME_DESCRIPTION'"/g \
-	/etc/openwrt_release
+LIME_VERSION_CODE="$LIME_RELEASE $LIME_REVISION $LIME_CODENAME"
+if [ -z "$LIME_RELEASE" ]; then
+	LIME_VERSION_CODE="$LIME_REVISION $LIME_CODENAME"
+fi
+
+LIME_DESCRIPTION="$LIME_ID $DISTRIB_RELEASE $LIME_VERSION_CODE"
+
+sed -i -e "s,^\(OPENWRT_RELEASE=\).*$/\1$LIME_DESCRIPTION/g" \
```
ilario: here it seems that the expression uses a "," as a field separator, but then changes to "/", no?

packages/lime-system/Makefile
```diff
@@ -32,28 +23,25 @@ define Package/$(PKG_NAME)
...
 define Package/$(PKG_NAME)/install
 	$(INSTALL_DIR) $(1)/
-	$(CP) ./files/* $(1)/
+	[ ! -d ./build/files ] || $(CP) ./build/files/* $(1)/
 	$(INSTALL_DIR) $(1)/etc/hotplug.d/lime-config/
-	$(LIME_VERSION_SED) \
-		$(1)/etc/lime_release \
-		$(1)/etc/profile.d/10-lime-banner.sh
+	$(INSTALL_DIR) $(1)/etc/lime-assets/community/
+	$(INSTALL_DIR) $(1)/etc/lime-assets/node/
```
ilario, line 42: Aren't the new `.placeholder` files you introduced in this pull request enough for ensuring that these folders are created?

packages/lime-system/files/etc/uci-defaults/89_lime-release
a-gave: Thanks fixed in https://github.com/libremesh/lime-packages/commit/e1c5e310b8483bbd9835e3df1d73df2d9b1b04dc https://github.com/libremesh/lime-packages/compare/783a7bf374899917371c7a8a98f4ed8e589bab5a..e1c5e310b8483bbd9835e3df1d73df2d9b1b04dc

packages/lime-system/Makefile
a-gave: The `/etc/lime-assets/{community,node}/.placeholder` is deleted as well as the whole `/etc/lime-assets/`.
It is created instead in the same way as the directory `/etc/hotplug.d/lime-config/` during the package install phase, so by apk. It is indeed just to uniform the way lime-system creates the three empty directories.

ilario: I don't understand, but it is ok :)

ilario:
The `development` here should also be replaced by `nightly` (but it will likely cause a conflict with #1216 as in there the line just above gets modified):
https://github.com/libremesh/lime-packages/blob/04b26f35e24d49cf0838cea82569ef04dc96749a/packages/lime-system/files/etc/uci-defaults/92_add-lime-repos#L12

packages/lime-system/files/etc/uci-defaults/89_lime-release
```diff
@@ -4,11 +4,18 @@
...
-	-e s/"^DISTRIB_DESCRIPTION=.*"/"DISTRIB_DESCRIPTION='$LIME_DESCRIPTION'"/g \
-	/etc/openwrt_release
+LIME_VERSION_CODE="$LIME_RELEASE $LIME_REVISION $LIME_CODENAME"
+if [ -z "$LIME_RELEASE" ]; then
+	LIME_VERSION_CODE="$LIME_REVISION $LIME_CODENAME"
+fi
+
+LIME_DESCRIPTION="$LIME_ID $DISTRIB_RELEASE $LIME_VERSION_CODE"
+
+sed -i -e "s,^\(OPENWRT_RELEASE=\).*$,\1$LIME_DESCRIPTION,g" \
```
ilario: This line ends up missing the quotes:

```
# cat /etc/os-release 
[...]
OPENWRT_DEVICE_REVISION="v0"
OPENWRT_RELEASE=LibreMesh 25.12-SNAPSHOT r26.266.44771~48d386e nightly
OPENWRT_BUILD_DATE="1790106349"
```

So that sourcing it generates this error:

```
# . /etc/os-release
-ash: /etc/os-release: line 19: 25.12-SNAPSHOT: not found
```

ilario:
>       * via cli running `ubus call system board`

I am using OpenWrt 25.12 compiled from its branch, and the ubus command does still use the old description.
Seems that that command takes the description from the `/usr/lib/os-release` file, so we should maybe modify that one also?

PR #1276, opened 2026-09-04
lime-utils: init the package with the script unique_append

### LINKED ISSUE #949: unique_append should check if /etc/crontabs/root file exist before appending

a-gave:
[WIP] still untested

**lime-system: utils.lua use nixio.fs for file operations**
This includes also a similar minor fix:
- fix: use the lua native #var to count tableLength(var)

**lime-utils: init the package with the script unique_append**
The following changes are made:
- initialize the package lime-utils with the script unique_append
- add a dependency on lime-utils to all packages
- always 'exit 0' on uci defaults that add new crontabs, to execute them
only at firstboot and not at every boot

Fixes: #949

### DIFF
--- a/packages/lime-system/files/usr/lib/lua/lime/utils.lua
+++ b/packages/lime-system/files/usr/lib/lua/lime/utils.lua
@@ -243,34 +243,15 @@ function utils.is_valid_hostname(hostname)
 end
 
 function utils.file_exists(name)
-	local f = io.open(name, "r")
-	if f ~= nil then
-		io.close(f)
-		return true
-	else
-		return false
-	end
+	return fs.stat(name, "type") == "reg"
 end
 
 function utils.read_file(name)
-	local f = io.open(name, "r")
-	local ret = nil
-	if f ~= nil then
-		ret = f:read("*all")
-		f:close()
-	end
-	return ret
+	return fs.readfile(name)
 end
 
 function utils.write_file(name, content)
-	local f = io.open(name, "w")
-	local ret = false
-	if f ~= nil then
-		f:write(content)
-		f:close()
-		ret = true
-	end
-	return ret
+	return fs.writefile(name, content)
 end
 
 function utils.is_installed(pkg)
@@ -304,11 +285,7 @@ function utils.tableMelt(t1, t2)
 end
 
 function utils.tableLength(t)
-	local count = 0
-	for _ in pairs(t) do
-		count = count + 1
-	end
-	return count
+	return #t
 end
 
 function utils.indexFromName(name)
--- a/packages/lime-utils/Makefile
+++ b/packages/lime-utils/Makefile
@@ -1,0 +1,19 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2006-2014 OpenWrt.org
+
+include ../../libremesh.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=LibreMesh utils core
+  CATEGORY:=LibreMesh
+  MAINTAINER:=Samuele Longhi <agave@dracaena.it>
+  URL:=https://libremesh.org
+  PKGARCH:=all
+endef
+
+define Package/$(PKG_NAME)/description
+	LibreMesh utils, core package with tools and helper scripts to use also outside the LibreMesh system.
+endef
+
+$(eval $(call BuildPackage,$(PKG_NAME)))
--- a/packages/check-date-http/Makefile
+++ b/packages/check-date-http/Makefile
@@ -9,7 +9,7 @@ define Package/$(PKG_NAME)
 	CATEGORY:=LibreMesh
 	MAINTAINER:=Gioacchino Mazzurco <gio@altermundi.net>
 	URL:=http://libremesh.org
-	DEPENDS:=+libuci-lua +lua
+	DEPENDS:=+libuci-lua +lua +lime-utils
 	PKGARCH:=all
 endef
 
--- a/packages/check-date-http/files/etc/uci-defaults/check-date-http-cron
+++ b/packages/check-date-http/files/etc/uci-defaults/check-date-http-cron
@@ -1,9 +1,6 @@
 #!/bin/sh
-
-unique_append() {
-	grep -qF "$1" "$2" || echo "$1" >>"$2"
-}
-
 unique_append \
 	"*/20 * * * * ((sleep \$((RANDOM % 600)); check-date-http &> /dev/null)&)" \
 	/etc/crontabs/root
+
+exit 0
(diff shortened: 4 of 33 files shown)


packages/lime-system/files/usr/lib/lua/lime/utils.lua
```diff
@@ -243,34 +243,15 @@ function utils.is_valid_hostname(hostname)
...
 
 function utils.file_exists(name)
-	local f = io.open(name, "r")
-	if f ~= nil then
-		io.close(f)
-		return true
-	else
-		return false
-	end
+	return fs.stat(name, "type") == "reg"
```
ilario, line 246: I like the new implementation!
But this function passed to be true also for directories to be true only for regular files, and this requires updating this test:

https://github.com/libremesh/lime-packages/blob/1e0e15ecaeedec454ff4e3a6134b371699b6d862/tests/test_test_utils.lua#L55

and some (mostly unused) code here (just checked the ones where the variable name looked like a directory's one):

https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua#L84

https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/eupgrade/files/usr/lib/lua/eupgrade.lua#L16

https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/pirania/files/usr/lib/lua/read_for_access/read_for_access.lua#L12

And also here as it seems that keeping directories over updates is supported (according to lime-example.txt, I did not check):

https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/lime-system/files/usr/lib/lua/lime/utils.lua#L437

ilario:
I love the new lime-utils package!!!
Do you think that eventually, slowly slowly, it could contain the `utils.lua`, `config.lua` and `modules.lua` files from lime-system, as proposed in https://github.com/libremesh/lime-packages/issues/771 ?

packages/lime-utils/Makefile
```diff
@@ -0,0 +1,19 @@
+# SPDX-License-Identifier: GPL-3.0-only
+#
+# Copyright (C) 2006-2014 OpenWrt.org
+
+include ../../libremesh.mk
+
+define Package/$(PKG_NAME)
+  TITLE:=LibreMesh utils core
+  CATEGORY:=LibreMesh
```
ilario, line 9: Do you think this package could be used out of LibreMesh and be pushed to OpenWrt's repositories, eventually?
In this case, you could already place the package in the Utilities category.

a-gave: Yes is seems sensed, ok for the category good catch will do!

a-gave:
> I love the new lime-utils package!!! Do you think that eventually, slowly slowly, it could contain the `utils.lua`, `config.lua` and `modules.lua` files from lime-system, as proposed in #771 ?

Yes the intention is that!

I started looking at utils.lua and I'm planning to move the functions like 'set_root_secret' to the module /usr/lib/lua/lime/system.lua and to port to the new lime-utils 'utils.lua' and 'config.lua'

I initially tought that modules could remain in lime-system, maybe also they should have their own directory which could be /usr/lib/lua/lime/modules and the list of modules is a list of files contained in that folder, to ease the possibility to add new modules

what do you think?

ilario:
> > I love the new lime-utils package!!! Do you think that eventually, slowly slowly, it could contain the `utils.lua`, `config.lua` and `modules.lua` files from lime-system, as proposed in #771 ?
> 
> Yes the intention is that!

Great!!!

> I started looking at utils.lua and I'm planning to move the functions like 'set_root_secret' to the module /usr/lib/lua/lime/system.lua and to port to the new lime-utils 'utils.lua' and 'config.lua'

Wow, so that utils.lua would be free of LibreMesh-specific functions? It would be amazing, but also a lot of work for moving many functions (I did not check them one by one, but I suspect they are many)!

> I initially tought that modules could remain in lime-system,

You are right, it is better.
To keep `modules.lua` in `lime-system` without having `lime-utils` depending on it, we could eliminate `utils.isModuleAvailable` (a function in utils that calls one in modules) and use `modules.is_available` instead.

Now that I check this again, also `config.lua` should stay in `lime-system`, as it is LibreMesh-specific. For removing the dependency of `utils.lua` from `config.lua`, we could move out of `utils.lua` the functions `utils.network_id` and `utils.set_shared_root_password`.

> maybe also they should have their own directory which could be /usr/lib/lua/lime/modules and the list of modules is a list of files contained in that folder, to ease the possibility to add new modules

Nice! It would be more tidy :)

packages/lime-system/files/usr/lib/lua/lime/utils.lua
ilario: > https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/lime-system/files/usr/lib/lua/lime/generic_config.lua#L84

Just checked this one. Actually the code does not support directories (it uses `cp -dpf`, without `-r`), so in this case it is ok the new `utils.file_exists` behaviour.

> https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/eupgrade/files/usr/lib/lua/eupgrade.lua#L16
> https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/pirania/files/usr/lib/lua/read_for_access/read_for_access.lua#L12

These two are designed to be true only for directories, not even for files. Do you think that it would make sense to have a `utils.dir_exists` function?
 
> And also here as it seems that keeping directories over updates is supported (according to lime-example.txt, I did not check):
> 
> https://github.com/libremesh/lime-packages/blob/e2876af2d4cb9f04be68dfcaf7340a9369b813d7/packages/lime-system/files/usr/lib/lua/lime/utils.lua#L437

I did check a bit more: it actually seems that both files and directories can be preserved, as the list gets passed to `tar`:

https://github.com/libremesh/lime-packages/blob/597853c37d143fd5a38d0a44c6023be90a018e2d/packages/safe-upgrade/files/usr/sbin/safe-upgrade#L239

So here there should be something like `utils.file_exist OR utils.dir_exists`.

PR #1277, opened 2026-09-08
fix(ci): point LG_ENV at -dual-tftp target when staging kernel + ramdisk

francoriba:
## Context

Follow-up to [libremesh/libremesh-tests#5](https://github.com/libremesh/libremesh-tests/pull/5) which splits the LibreRouter v1 target YAML into two variants:

- `targets/librerouter_librerouter-v1.yaml` - single-image (default)
- `targets/librerouter_librerouter-v1-dual-tftp.yaml` - kernel + rootfs uImage

**Why the split**: the single-image variant is what mesh tests, LibreMesh releases and `openwrt-tests` healthcheck actually use (initramfs-kernel.bin, only `LG_IMAGE` set). The previous librerouter YAML required both `LG_IMAGE` and `LG_IMAGE_INITRD` unconditionally, breaking every non-dual flow with:

```
InvalidConfigError: configuration file '.../librerouter_librerouter-v1.yaml'
refers to unknown variable 'LG_IMAGE_INITRD'
```

## Problem in this CI

`.github/workflows/build-firmware.yml` pre-sets:

```
LG_ENV=targets/${{ matrix.device }}.yaml
```

After the libremesh-tests split, that path points at the **single-image** variant.

When `tools/ci/lab_stage_firmware.sh` then detects the dual-tftp payload (`bin` + `uimage` pair produced by `build_image.sh` for devices like the LibreRouter v1 on ath79), it exports `LG_IMAGE_INITRD` — but the single-image YAML does not reference that variable, so labgrid loads only the kernel and boot fails without a rootfs.

## Fix

When `lab_stage_firmware.sh` enters the dual-TFTP branch, override `LG_ENV` with the `-dual-tftp` sibling target file so labgrid resolves `LG_IMAGE_INITRD` and issues both TFTP loads:

```bash
DUAL_ENV="targets/${DEVICE}-dual-tftp.yaml"
echo "LG_ENV=$DUAL_ENV" >> "$GITHUB_ENV"
```

No change for single-image devices (`openwrt_one`, `linksys_e8450`, `bananapi_bpi-r4`, `qemu_*`): the else branch keeps the pre-set `LG_ENV`.

## Depends on

- [libremesh/libremesh-tests#5](https://github.com/libremesh/libremesh-tests/pull/5) - merge first so the `-dual-tftp` target file exists in the checked-out libremesh-tests.

### DIFF
--- a/.github/workflows/build-firmware.yml
+++ b/.github/workflows/build-firmware.yml
@@ -562,6 +562,12 @@ jobs:
 
       - name: Lock labgrid place
         working-directory: libremesh-tests
+        env:
+          # LG_PLACE must be available before labgrid-client parses LG_ENV so
+          # that !template "$LG_PLACE" in target YAMLs resolves correctly.
+          # (lab_stage_firmware.sh may override LG_ENV to a -dual-tftp variant
+          # that still references $LG_PLACE in RemotePlace.)
+          LG_PLACE: labgrid-fcefyn-${{ matrix.place }}
         run: |
           set -euo pipefail
           echo "LG_PROXY=labgrid-fcefyn" >> "$GITHUB_ENV"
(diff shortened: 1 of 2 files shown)


ilario (approved):
You are the master of the continuous deployment, so I approve any change you consider that make sense to do on those workflows :) Feel free to merge yourself when you think think that the code is ready.

Issue #1279, opened 2026-09-17
drop dependencies of lime-debug and shared-state on iputils-ping


a-gave:
I noticed that the package 'iputils-ping' provides a retrocompatibility that installs 'ping6' which seems unused.
Another related discussion was: https://github.com/libremesh/lime-packages/issues/794#issuecomment-1497395956 which seems in favor of the removal

### OFF TOPIC
I noticed it when installing the package 'profile-libremesh-suggested-packages' on a Gemtek WVRTM-127ACN from a plain openwrt 25.12.5.
Installation which now seems to work correctly. 
To completely install the profile I had to run 'apk add <profile>' twice since "most probably" the device lost the internet connection provided via wireless-wan after the first run of lime-config, after the installation of lime-system.

The package iputils-ping is indeed not really involved in this, it was just a coincidence that the installation process seemed stuck at installing 'iputils-ping', but it was that the device had instead simply lost the wireless-wan connection.

This possibility to install libremesh packages from openwrt via apk seems to be working properly now. 
Probably due to recent changes in openwrt related to the switch from opkg to apk.
It is now possible to ask to the firmware-selector a build for libremesh without specifying to remove '-dnsmasq' and it succeds without failures.
Didn't checked carefully but this seems related to https://github.com/openwrt/openwrt/pull/21358

The other commonly removed package '-odhcpd-ipv6only' is otherwise harmless and comes with a default setting that keeps it disabled

```
root@OpenWrt:~# apk add profile-libremesh-suggested-packages
( 1/69) Purging dnsmasq (2.93-r1)
  Executing dnsmasq-2.93-r1.pre-deinstall
( 2/69) Installing dnsmasq-dhcpv6 (2.93-r1)
  Installing file to etc/config/dhcp.apk-new
  Executing dnsmasq-dhcpv6-2.93-r1.post-install
  * udhcpc: started, v1.37.0
  * udhcpc: broadcasting discover
  * udhcpc: no lease, failing
( 3/69) Installing liblua5.1.5 (5.1.5-r11)
  Executing liblua5.1.5-5.1.5-r11.post-install
( 4/69) Installing libubus-lua (2026.06.28~24864e78-r2)
  Executing libubus-lua-2026.06.28~24864e78-r2.post-install
( 5/69) Installing babeld (1.13.1-r3)
  Executing babeld-1.13.1-r3.post-install
( 6/69) Installing jq (1.8.1-r2)
  Executing jq-1.8.1-r2.post-install
[...]

PR #1280, opened 2026-09-18
lime-system: lime-apply: add pre and post hooks around reload_config

a-gave:
This removes the execution of 'lime-apply' from the uci-defaults contained in the package lime-app.

As described here https://github.com/libremesh/lime-packages/issues/1279#issue-5493608710 it is now mostly doable to install libremesh from a plain openwrt, simply installing libremesh packages, like the metapackage 'profile-libremesh-suggested-packages'.

Tested with the changes included in this pull request on a Gemtek WVRTM-127ACN following these steps:
1. install the metapackage with all libremesh default packages
```
apk add profile-libremesh-suggested-packages
```
2. now generate and apply the configurations with:
```
lime-config; lime-apply
```


ilario:
It is much cleaner now, thanks! :)
A question: when installing a pre-compiled LibreMesh firmware, lime-apply does not get executed, and thus the title of lime-app does not get updated to include the hostname, right? (not a big issue indeed...)

a-gave:
> It is much cleaner now, thanks! :) A question: when installing a pre-compiled LibreMesh firmware, lime-apply does not get executed, and thus the title of lime-app does not get updated to include the hostname, right? (not a big issue indeed...)

I left this as draft because I'd like to test also this. My guess is that the lime-apply in that uci-default could be redundant or "eventually" it could be moved to /etc/hotplug.d/lime-config.

I'll test doing a sysupgrade with the changes included in this pull request
using feeds from my clone at https://raw.githubusercontent.com/a-gave/lime-feed/refs/heads/gh-pages/master/openwrt-25.12/x86_64/packages.adb and report back

a-gave:
Tested on a Gemtek WVRTM-127ACN doing the sysupgrade with the code that add a lime-config hook.
- the lime-config's hook produces a small situation where the hostname is changed in the browser web page title but not inside the page h1. Is not much important because it implies a mixed situation of CLI running lime-config (without calling lime-apply) and WEB browser access. However it confirms that the change in the hostname should happens in lime-apply
- Reworked so the lime-apply to have hooks around `/sbin/reload_config` similar to lime-config's hooks (I know this may seem like too big a change, but it actually makes it easier to organize for future needs)
- Maintained a first replacement of the hostname inside the lime-app's uci-default to overwrite the initial title 'LiMeApp' without calling the entire lime-apply
- Untested: but that first lime-apply from the uci-default of lime-app that was calling /sbin/reload_config was probably slowing down a bit the boot process

**Still untested with the new changes that introduce lime-apply's hooks**

a-gave:
Renamed `/etc/uci-defaults/90_lime-app` to `/etc/uci-defaults/93_lime-app` to move the script after the first execution of lime-config during `/etc/uci-defaults/91_lime-config`

Tested on a Gemtek WVRTM-127ACN and all seems ok. before the rename from '90' to '93' the device boots with the name 'OpenWrt', after it takes the name 'LiMe-cad579'.

This seems to me ready!

ilario (approved):
Tested a bit, seems working to me!

